PR #2, opened 2022-12-15
Improve UI

TheEvilSkeleton:
This MR does the following:
- Add equation window can be closed with Esc
- Create `save_file_chooser` function for save dialogs (to avoid copy pasting multiple lines of code)
- Set file chooser dialogs as modal (they stick with the parent window)
- Cancel button closes the window and does nothing for file chooser
- Set Preferences window as modal
- Remove arrow from hamburger menu

Criticism welcome. :)

### DIFF
--- a/src/add_equation_window.ui
+++ b/src/add_equation_window.ui
@@ -7,6 +7,16 @@
    <property name="default-width">256</property>
   <property name="default-height">125</property>
   <child>
+   <object class="GtkShortcutController">
+    <child>
+     <object class="GtkShortcut">
+      <property name="trigger">Escape</property>
+      <property name="action">action(window.close)</property>
+     </object>
+    </child>
+   </object>
+  </child>
+  <child>
    <object class="GtkBox">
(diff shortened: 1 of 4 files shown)


SjoerdB93:
Hey,

Thanks so much for the PR! Nice chances all over the board :)
It's indeed nice with the modal dialogs. At one point I'd probably want to set the add_equation window and the transform_window as modal as well, but couldn't figure it out right now so will look at that later. 

There was an issue with this PR, as you were setting model to true in the function that generates the file chooser dialog giving rise to a "AttributeError: 'function' object has no attribute 'set_modal". I've played around a bit, and fixed it as well as some other bugs:
- I no longer edit the file path itself, this allows the flatpak to save the files without granting permissions to the file system
- Saving multiple files now triggers a select_folder action instead of a hacked open action that does the same thing
- I've added a placeholder text in the equation window

Will  merge the above chances with this PR and then merge that with main. Thanks a lot!

PR #4, opened 2022-12-15
Fix GNOME typos

TheEvilSkeleton:


SjoerdB93 (approved):
You're right, these should be capitalized

PR #12, opened 2023-01-11
icon: new, more detailed app icon

### LINKED ISSUE #11: New icon?
The current icon doesn't follow the GNOME icon guidelines, would you like help with a new one?

Here's a first attempt:

image

bertob:
Closes #11 

image


SjoerdB93 (approved):
Thanks, I tested the build on my device and it all works great (as expected, but wanted to double check before merging). It looks really nice, and blends much more nicely with the other applications on my machine! I also like that you included a devel one by the way.

Really, thanks a lot!

PR #16, opened 2023-01-19
Make the Project run in Gnome-Builder

cmkohnen:
This pulls in files and changes from https://github.com/flathub/se.sjoerd.DatMan and let the program run out of the box in Gnome-Builder

### DIFF
--- a/python3-matplotlib.json
+++ b/python3-matplotlib.json
@@ -1,0 +1,90 @@
+{
+    "name": "python3-matplotlib_modules",
+    "buildsystem": "simple",
+    "build-commands": [],
+    "modules": [
+{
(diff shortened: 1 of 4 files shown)


SjoerdB93 (commented):
As you may see fron the project structure, I've been using GNOME Builder in the first place. However, I've kept the JSON files seperated as those are just the Flatpak Manifest which are hosted on the Flathub repo.

Honestly, I didn't really think about this breaking out-of-the-box compatibility with Gnome Builder. While end-users should generally use the Flatpak directly anyway, there's absolutely merits to direct compatibility with Gnome builder this way. I'll double check the files, but will then probably merge this!

PR #17, opened 2023-01-19
Fix XML indentation and closing errors

cmkohnen:
Fixes all indentation errors I've spotted. Also removes two closing brackets, who had no opening bracket.


SjoerdB93 (approved):
Thanks, I typically use the "Text Pieces" tool to fix my indentation, but these indentation errors still accumulate a bit over time. Unfortunately, auto-indentation and such doesn't work great on gnome builder yet..

 Hadn't even noticed the closing brackets, thanks for pointing those out.

Issue #14, opened 2023-01-14
UI: Avoid grid of buttons


bertob:
Currently the interface has a grid of relatively random buttons on the left, which is not a pattern that makes sense in software, since actions should always be contextual and ideally connected to the content they affect. I can't make recommendations for what to do with all of them since I don't really know the main use cases/workflows for the app, but some general guidance:
- Actions should only be visible if they actually work in the current state (e.g. most of the buttons make no sense until there's data)
- Global app actions (e.g. save data) can go in the headerbar or primary menu
- Actions specific to the left pane could maybe go in a toolbar on the bottom of the left pane
- Less important actions could go in a menu/popover
- For selection this should use the regular selection mode pattern (like e.g. in Contacts) https://developer.gnome.org/hig/patterns/containers/selection-mode.html

![image](https://user-images.githubusercontent.com/1908896/212445971-6b5d5252-55cf-4622-85ee-4d72641207d2.png)

SjoerdB93:
Hey,

Thanks for opening this issue. I agree that the grid needs an UI update. Another major issue for me, is that the layout quickly starts looking like a 2000's karaoke machine if I would add more buttons. It's quite pragmatic, everything is available and it works quite well, but it doesn't exactly scream modern design and there's a risk of getting overwhelmed with too many choices presented. Especially if adding more.

I wanted to mention that I'd actually asked for advice in the Gnome Design Matrix channel, but I see you that you were the one who responded. So thanks for that. Unfortunately I've been quite occupied with other stuff last two weeks, but next weeks should be better. 

To reply a bit more in-depth. I think the save data button indeed should go somewhere in the header bar (so either the bar itself, or more likely in a menu. Perhaps the primary menu). I like the idea of splitting the header bar as we see in many GNOME applications with a side-bar

The biggest challenge is probably the fact that almost all actions on the left-bar are all operations that are performed on the data sets (exceptions being "Select All/Select None", "Select span", and "Save Data"), in other words they kinda act the same.. 
All actions are also performed on the selected data (except for "shift vertically", which is performed on all data). So there's no obvious hierarchy here. They all basically behave the same, and they're all relevant at the same moment (which is, after you have opened a data set). The only exception maybe is "Cut Data", which is only really relevant when you have selected a span. See video:
[Screencast from 2023-01-14 16-24-25.webm](https://user-images.githubusercontent.com/68477016/212479563-0e1f9b44-866f-4faa-930a-40b1f2d7cbf0.webm)


[...]

PR #22, opened 2023-01-20
Proposal: UI changes

cmkohnen:
A few Changes:

- Move the Graph Box out of "selection_box"
- use AdwFlap for the sidebar, moving the graph into the content child, the vertical separator to the separator child and the sidebar into flap
- Move the save button to the headerbar
- Use AdwExpanderRows for housing the buttons
- Use AdwEntryRows for input fields
- Housing all AdwExpanderRows in a ScrollWindow
- Removing a few " characters

Preview: 
grafik
grafik

Feel free to request changes

### DIFF
--- a/src/window.ui
+++ b/src/window.ui
@@ -1,7 +1,7 @@
 <?xml version='1.0' encoding='utf-8'?>
 <interface>
   <requires lib="gtk" version="4.0" />
-  <requires lib="Adw" version="1.0" />
+  <requires lib="Adw" version="1.2" />
   <template class="DatManWindow" parent="AdwApplicationWindow">
     <property name="default-width">1024</property>
     <property name="default-height">768</property>
@@ -19,7 +19,7 @@
                     <property name="always-show-arrow">True</property>
                     <property name="icon-name">list-add-symbolic</property>
                     <property name="menu-model">add_data_menu</property>
-                    <property name="tooltip-text" translatable="yes">Add new data</property>"
+                    <property name="tooltip-text" translatable="yes">Add new data</property>
                   </object>
                 </child>
                 <child>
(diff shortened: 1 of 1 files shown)


SjoerdB93 (changes requested):
Wow, this is great and goes a long way in solving issue #14 .
In general I feel like this is a good improvement, as it does get rid of a cluttered grid with actions that may not be relevant to 90% of the use cases, while still allowing them to be visible when you need them.  

A typical workflow I've been thinking about is whenever you need to apply one obscure function relatively often. Hiding away such functions can make this a bit tedious. This design allows the user to have the actions open that are relevant to them which I really like.

Personally I feel like it makes more sense to give the text in the header itself a regular title tag.  See screenshot: 
image


Although I think it's quite fine with keeping the rows themselves as subtitles. I experimented a bit, and having all as regular titles makes it cluttered really quickly. Having both titles and subtitles may be nice, but that has the same problem of adding clutter. Just having the headers as titles still works nicely though. I should check maybe which option works best with the GNOME HIG. If you have a good reason to use subtitles for all text, I'm willing to consider it, but my initial thought is that titles for the headers work better.

Other than that. I was personally planning to put the save button indeed in the title bar, but also implement a split title bar there such as GNOME contacts has:
image

So then it would be a bit like the following mock-up:
image

But also here I maybe should check what fits best with the GNOME HIG guidelines.

SjoerdB93:
One final comment, "Transform and Multiply" should be "Translate and Multiply"

cmkohnen:
Thanks for the fairly quick response!
I do think, that having the headers use titles is the better solution. It helps them stand out if they are expanded and space isn't an issue here. 

In tackling the select buttons, maybe a list with checkboxes would be more intuitive and more easy to see which ones are selected.
As a header there could be a checkbox which automatically selects/deselects all objects.
However, I haven't seen this behavior in gnome but only in QT/Windows Applications, so I don't know if it would fit into GNOMEs HIG.

SjoerdB93 (approved):
Thanks a lot, this is looking great!

I'll most likely push a new build to Flathub with Graphs 1.4.0 this weekend, which will then include these changes. I'll just try to see how to split the header bar and implement that, and maybe try and squash some small bugs I've encountered. 

I'll also need to change the about section to properly credit contributors before pushing version 1.4.0.

PR #23, opened 2023-01-20
move icons to subfolder

cmkohnen:
Moves custom icons to a subfolder called "icons".
Makes the project structure look a little cleaner.


SjoerdB93 (commented):
Thanks, this was much needed. I actually tried to do this in the past, but it didn't work out due to a misconfiguration in the meson build file. So this being split like this is much appreciated. 

I just went to bed (typing this on my phone) so I didn't have the opportunity to test this yet. While it seems okay, it's probably best practice if I test it before merging. But I'll do so tomorrow morning :)

SjoerdB93 (approved):
Works great!

PR #24, opened 2023-01-21
bump libadwaita requirement to 1.2

cmkohnen:
Since the flatpak target is GNOME 43, which ships with libadwaita 1.2 and a few components in use are 1.1 or newer, there should be no downside to raising the requirement. Leaving it as is might lead to future errors in non-flatpak builds.


SjoerdB93 (approved):
Since Flatpaks are self-contained, and is the primary platform where Graphs is distributed, this should not break compatibility either. So I don't see any downsides either to keeping this up-to-date.

PR #25, opened 2023-01-21
Proposal: UI changes

cmkohnen:
- Use AdwComboRows instead of AdwActionRows with a child
- Use AdwEntryRows
  - Open for suggestions as this removes the ability to add subtitles. However existing subtitles were pretty self-explanitory.
- Use GtkSwitch instead of GtkCheckBox to better fit into other apps, as CheckBoxes are used for a pick-between-options selection. In this case they toggle one option separate from the others.
- Make AdwActionRows activate their respective widget (except sliders)
- Move Toast regarding invalid equation into the popup window itself

Planned:

- [✔️  ] Settings
- [✔️ ] Add from Equation Window
- [✔️ ] Add from Data advanced Window
- [✔️ ] Transform Window

This broadens the surface area of each option, not requiring aiming on a small widget
grafik
grafik
grafik

### DIFF
--- a/src/add_data_advanced.ui
+++ b/src/add_data_advanced.ui
@@ -34,55 +34,39 @@
             <property name="margin-top">20</property>
             <property name="margin-bottom">20</property>
             <child>
-              <object class="AdwActionRow">
+              <object class="AdwEntryRow" id="name">
                 <property name="visible">True</property>
                 <property name="can-focus">True</property>
-                <property name="title" translatable="yes">Name</property>
-                <property name="subtitle" translatable="yes">The name that is to be used for the data set (single file only)</property>
-                <child>
-                  <object class="GtkEntry" id="name">
-                    <property name="placeholder-text">Leave blank to use filename</property>
-                    <property name="max-width-chars">10</property>"
-                    <property name="valign">center</property>
-                  </object>
-                </child>
+                <property name="title" translatable="yes">Name (single file only)</property>
+                <!--property name="subtitle" translatable="yes">The name that is to be used for the data set (single file only)</property!-->
+                <property name="max-width-chars">10</property>""
               </object>
             </child>
             <child>
-              <object class="AdwActionRow">
+              <object class="AdwEntryRow" id="delimiter">
                 <property name="visible">True</property>
                 <property name="can-focus">True</property>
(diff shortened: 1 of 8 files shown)


cmkohnen:
Missed a commit

SjoerdB93:
> Use AdwComboRows instead of AdwActionRows with a child

I agree, this looks better and less cluttered, good change.


> Use AdwEntryRows. Open for suggestions as this removes the ability to add subtitles. However existing subtitles were pretty self-explanitory.

Yes, this was the only point I was not 100% sure. On the one hand, I do like the AdwEntryRows, I was honestly not aware of the widget initially, and they're quite nice. It's also satisfying how the text turns small in an animation when something is entered. However, it does lose the subtitle indeed.

I think it should be fine as long as the title is self-explanatory. With the "Transform Data" window, it no longer shows the "Y =" on the same page, making it look less like one equation where you enter the right-hand side. Instead the "Y = " looks more like a title, describing what you can enter there. Maybe it's an option to make the title a bit more descriptive then for instance "Equation: Y = " or something like that.

In the end I prefer to have it all consistent than to use a different widget here just so that it looks like one equation.

> Use GtkSwitch instead of GtkCheckBox

Good, this was actually on my own to-do-list.

> Make AdwActionRows activate their respective widget (except sliders)

Again, great!

> Move Toast regarding invalid equation into the popup window itself

Also, probably better like this yes, I agree.

SjoerdB93:
I'll have to think a bit about whether the title for the equations should be changed (instead of having it at "Y =" as it is right now), but I'll likely end up merging this

SjoerdB93:
We will also need to need to change the "Preferences" window to follow the same design. You are free to do that yourself, otherwise I will do that, probably tomorrow or later otherwise later today this evening.

cmkohnen:
An option would be to use prefix widgets for the "X =" part as seen below, but then they would not align properly

grafik

cmkohnen:
> We will also need to need to change the "Preferences" window 

Completely slipped past me, thanks for noticing

SjoerdB93 (approved):
Looks good to me.

I think at the moment it is fine to leave the "Equation" titles as is. They don't bother me that much.

Issue #26, opened 2023-01-21
Consider removing "allow_duplicate_filenames" config option


cmkohnen:
As the title says.
Whilst discovering and trying to fix a logic error in add_equation.py, I discovered that although multiple files can have the same name, only the latest added would be considered in the graph or the datadict, overriding any previous one. 

The related code is in add_equation.py at line 23 and datman.py at line 152

The function of just adding a number in brackets works just fine, maybe adding a toast after addition saying the name was changed would be optimal.

Correct me if I'm wrong

SjoerdB93:
I've just checked this, and it seems something has been broken along the way. 

Originally, this started as a tool that I just used myself to make my life easier, and then grew into a bit more proper hobby project. 
In the very original design, it only loaded new data if the filename didn't exist. So if I tried to open a datafile that was already open, it didn't open it again. This was the default option.

Along the way, I added the option to open the same data multiple times anyway, and just made it add a number in brackets afterwards. As this is probably the most sensible default for most people, I made this the default setting. So by default, "allow_duplicate_filenames" is true. It's also how I have it configured myself usually, so I'm not sure when the "false" functionality broke. 

So the idea is that if you set this config option to false, and then import a dataset twice, that it doesn't load this dataset again. But this feature mainly grew out of the fact that in my initial prototype, I ran into issues with duplicate filenames and then just made it so that it doesn't load the dataset again. Then I added the option as well to just change the filename instead ("Allow duplicate filenames on").

But I do have to admit that the use-case of this particular setting is very narrow. I'm not completely opposed to removing it, or maybe change the wording a bit. I did fix the bug however, although the fix is a bit ugly code-wise, it does work for now (on a similar note, I will start to add docstrings and comments to the code-base soon as well).

 If I decide to keep it in, I should also add a toast in the "Add equation" window. But again, I'm a bit torn about keeping it in at all. I do see arguments for just removing the functionality all-together.

SjoerdB93:
In fact, the more I think about it, the more I'm leaning towards getting rid of this config setting. I have difficulties finding a clear use-case, it may introduce confusion, and it only adds to clutter (both in the settings and in the code-base).

SjoerdB93:
Pushed a slightly less hacky fix. Now it works as it did previously. Not sure when I broke it.

cmkohnen:
Thanks for the explanation!
> as I use the filename as dictionary key
[...]

PR #28, opened 2023-01-22
use unique identifiers for data management

cmkohnen:
This allows for multiple items / files bearing the same visual name to be displayed simultaneously without overriding each other. I'm trying to update the config to implement changes discussed in https://github.com/SjoerdB93/Graphs/issues/26#issuecomment-1399338374 
grafik
Still needs testing, as I haven't tried every option yet

### DIFF
--- a/src/add_equation.py
+++ b/src/add_equation.py
@@ -22,15 +22,15 @@ def on_accept(widget, self, window):
         new_file = create_data(self, x_start, x_stop, equation, step_size, str(window.name_entry.get_text()))
         name = new_file.filename
         handle_duplicates = self.preferences.config["handle_duplicates"]
-        if not handle_duplicates == "keep":
+        if not handle_duplicates == "Add duplicates":
             for key, item in self.datadict.items():
                 if name == item.filename:
-                    if handle_duplicates == "auto-rename":
+                    if handle_duplicates == "Auto-rename duplicates":
                         new_file.filename = datman.get_duplicate_filename(self, name)
-                    elif handle_duplicates == "ignore":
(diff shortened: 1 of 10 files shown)


SjoerdB93:
Hey,

This is looking great! 
I was planning to add this concept of using an identifier instead of a filename later today, or potentially tomorrow, but you were really fast :)

I've played around with it, and it does seem to work well. I noticed the "delete" hotkey didn't work, as it wants both the filename and the id. I fixed this in the latest commit, now it just takes the filename from the data object instead, which is probably better anyway.

cmkohnen:
I have now added four options:

- auto-rename appends a (1), (2), etc. (default for now)
- keep doesn't do anything and just adds the file / equation
- ignore skips the file (giving a toast)
- override overrides any existing file with that name

SjoerdB93:
Yes, that's a nice set of options. I like the addition of the override option as well, that's nice to have. 
I now fixed the delete button now as well, and everything seems good to go.

Only thing I'm unsure about is whether we should use more descriptive labels for the options. Either just by changing the keys to more descriptive ones directly, or rename the dropdown options and couple those to the keys in-code. An option such as "keep" for example may not be directly obvious for a user what it does. Something in the vicinity of: 
"Auto-rename duplicate files"
"Add duplicate file" 
"Ignore duplicate files"
"Override duplicate files"

Not 100% sold on those labels (maybe "override existing files" would be more correct), but I was thinking of something in that vain.

cmkohnen:
I agree, having them this short is a bit unintuitive

SjoerdB93 (approved):

PR #29, opened 2023-01-22
remove redundant/empty files

cmkohnen:
1. remove "empty file" probably created by mistake in GNOME Builder
2. remove "__init__.py" which was probably created at project init, but never used
3. remove two icons moved into subfolder but picked up as duplicate by git
4. remove "open_special.ui" an earlier version of "add_data_advanced.ui"?


SjoerdB93 (approved):
You're spot on with regard to the origin of these files.

PR #33, opened 2023-01-28
clean window.ui

cmkohnen:
- Have two GtkBoxes for buttons in the headerbar: one for actions at the start (add, undo, redo, select) and one at the end (save & primary)
         - move everything into these Boxes
         - change spacing to 5 pixels
- remove redundant GtkGrid, since all buttons formerly in it have been (re)moved. Also remove respective layout property of equation_scroll
- give the item list 50 more pixels


SjoerdB93 (approved):
Good changes, works good here.

Issue #32, opened 2023-01-28
UI: Use GNOME icons


TheEvilSkeleton:
I noticed the toolbar does not use GNOME icons:

![image](https://user-images.githubusercontent.com/50847364/215273772-f6993eda-aaaf-4e05-96dd-1856cda3fc7d.png)

I suggest to use GNOME icons whenever possible, as they look more familiar and better on GNOME. I recommend using [Icon Library](https://apps.gnome.org/app/org.gnome.design.IconLibrary/).

SjoerdB93:
Thanks a lot for the heads-up. Indeed it would be nicer to use the GNOME icons, when making the toolbar icons I didn't consider there may be existing icons out there that I could use. I will swap them out, seems I have a good alternative for all of them in the Icon Library.

SjoerdB93:
Fixed with PR #39

PR #39, opened 2023-01-28
Use GNOME icons in toolbar & change project structure

cmkohnen:
- Move GResource Manifest to data and rename to match appid
- Move .ui files into data/ui
- Move meson file out of items directory
- Use GNOME icons in toolbar (https://github.com/SjoerdB93/Graphs/issues/32) - I left change-x/y scale because I could not find a match in IconLibrary
- Move all icons into data/icons
- Remove icons no longer in use (!)
- deduplicate code in toolbar.py
- change meson configuration

The cleanest solution would be to compile all icons in the GResource and use Gio.File to make them accessible for matplotlib to read, however I couldn't get it to work, maybe someone else can, for now they remain installed using meson.

Regarding the save icon: The floppy icon currently in use, is what many people learned to associate with saving, however GNOME tends towards the following:  
grafik
Feedback appreciated, which to choose

Screenshots:
grafik
grafik


SjoerdB93:
Thanks, that's quite nice work. Regarding the save icon, personally I'm more inclined to go for the non-floppy logo with the arrow in this case, despite nostalgic feelings towards the floppy icon. Simply because I'm a bit unsure how obvious this specific floppy icon will be at a glance. 

For the scaling icons, I noticed that as well when I was digging into this (I was trying to get them from GResource, but that didn't really work out). There is a horizontal and vertical arrow icon (like this <---> and 90 degrees flipped for the vertical one), but that icon is typically used for mirroring items so it is not really the correct icon for this.

I was thinking of trying to remake the arrows in the icon so that they look closer to the arrows in the "Pan" icon we do got from GNOME. Not sure if the fonts need additional work, I had a hard time finding a font that looked readable at this small size. But I'll at least try and fix the arrow a bit later in the day.

cmkohnen:
Welp, seems like I broke the gtk icon cache somewhere down the line

SjoerdB93:
Just wanted to mention the same thing, and then got a notification that it's not just my machine.
See image below that I had open:

--------------------
It doesn't build on my machine:

```
Running custom install script '/usr/bin/glib-compile-schemas /app/share/glib-2.0/schemas'
Running custom install script '/usr/bin/gtk4-update-icon-cache -q -t -f /app/share/icons/hicolor'
--- stdout ---

--- stderr ---
gtk4-update-icon-cache: Failed to open file /app/share/icons/hicolor/.icon-theme.cache : No such file or directory

FAILED: install script '/usr/bin/gtk4-update-icon-cache -q -t -f /app/share/icons/hicolor' exit code 1, stopped
FAILED: meson-internal__install 
/usr/bin/meson install --no-rebuild

```

cmkohnen:
fixed it.
Also moved the config file to data since it's not really source code

SjoerdB93:
Yes, that makes sense to do. Just checked and it still succesfully loads a new config file from this template in case it's missing, so that still works.

I tried to just take the arrows from GNOME's "pan/move" icon and put it in the "Change scale" ones. But then I need to resize the text a bit. Which is not an improvement, the text is already small as-is:
image

I think I'm fine with either keeping those icons as-is for now. Or alternatively using the "flip vertical" and "flip horizontal" icons:
image

But those have the disadvantage of not being the correct icon, and may therefore be a bit more confusing.

cmkohnen:
I think the current icons don't look out of place, I wouldn't mind keeping them as is

SjoerdB93:
Yeah, I think they look fine. And as they're build in meson, they won't look at out of place for people using different icon themes either. Gonna merge this PR.

By the way, if you'd like feel free to add your name to the  `developers=['Sjoerd Broekhuijsen <contact@sjoerd.se>']` line in the main.py file, if you want your name in the "Code by" section in the about dialog.  You've contributed quite a bit to Graphs so I'd definitely merge it.

SjoerdB93 (approved):

Issue #40, opened 2023-01-29
Update the appid?


cmkohnen:
The current appid still reads as se.sjoerd.DatMan, however all visual references, as well as Github and the linked website do name the application as Graphs.

Should this be updated?
The main implication, apart from changing all references in code and build configuration, would be flatpak. A new appid would need to be requested on flathub and all user configuration would be lost.

It would however be more intuitive for users as well as contributors and should, if chosen to, be done before the project enters a nautilus/files state, where changing it would be a to monumental task.

SjoerdB93:
Yes, it would also be much nicer to find the application from the CLI. 
The appid that it has right now, was because the initial build was called DatMan, short for Data Manipulator. I agree that se.sjoerd.Graphs would make much more sense.

As far as I understood, to change the appid on Flathub, I need to submit the application again to Flathub as a PR. 
I don't think user configuration being lost is the end of the world. As it is right now, it resets user configuration also when the keys no longer match with the templates for instance (otherwise you run into key errors). 

I'm not sure if existing installs just "migrate" over to the new app id. I will probably also lose the statistics (Graphs is on 1731 total installs right now), but the latter is not that important. I'll ask around on the Flatpak matrix channel.

SjoerdB93:
I'm still not sure if manual intervention is necessary for people who have installed the application, but this indeed needs to be done by just pushing a PR to Flathub with this request, similar to get an app published on Flathub in the first place:

From another issue:
> As mentioned before, just open a new merge request with the new app-id and one of the reviewers will merge it and end of life the old app-id. 

I think it would probably be nicest if we can rename it from se.sjoerd.DatMan to se.sjoerd.Graphs. So once we've got those changes prepared, I can go ahead and request the change on Flathub.

SjoerdB93:
I have submitted [a PR to Flathub](https://github.com/flathub/flathub/pull/3865). It's combined with a very minor update towards v 1.4.1, as the latest Flatpak build needs to be done from the commit with the latest changes.

Will close this issue as soon as it's merged, in the unlikely event it gets denied I would switch back to the old app-id.

SjoerdB93:
The flathub PR got merged, so I'm closing this issue. Our appid is now officially changed.

SjoerdB93:
[...]

PR #41, opened 2023-01-29
Update appid

cmkohnen:
This renames all files and performs a simple string replacement on all textfiles. Compiles and seems to work just fine. 
See https://github.com/SjoerdB93/Graphs/issues/40


SjoerdB93:
Thanks a lot, I'll see when I got an answer in the Matrix channel to see if there's more that needs to be fixed. When that all seems clear, I'll merge this and put a PR on Flathub to change the appid.

Issue #36, opened 2023-01-28
Make graphs list sidebar retractable


mossymossmoss:
Making the graphs list sidebar retractable would make it so that you can still use the app when the window is scaled down horizontally, as currently it just disappears entirely
![image](https://user-images.githubusercontent.com/56167669/215284220-5602c5d0-b89b-4f00-8842-6ef09e1c159d.png)

It would also allow you to get rid of the "Open In New Window" button, as you would be providing the user with essentially the same view
![image](https://user-images.githubusercontent.com/56167669/215284186-f1365e20-73b3-463d-aaba-edc7c7607991.png)
![image](https://user-images.githubusercontent.com/56167669/215284199-08210462-e792-41c0-a818-a422bafc0d40.png)

cmkohnen:
I have actually tried to implement this about a week ago.
The foundation to do this already exists, since the sidebar is currently using libadwaitas Flap, which takes away ~50% of the work needed.
I'll try tackling this in the next few days.

SjoerdB93:
@cmkohnen Thanks, I also played around trying to achieve the same thing after you implemented the Libadwaita Flap, but my attempts didn't work out. So it was indeed on the list of things that should be done.

PR #43, opened 2023-01-29
implement rudimentary toggle_sidebar_button

cmkohnen:
This adds basic functionality to toggle the sidebar. It also gets rid of the ScrolledWindow containing the graph, as showing only portions at small window sizes doesn't seem like a good design choice. goes a long way in tackling https://github.com/SjoerdB93/Graphs/issues/36 
Use is (in my opinion) quite intuitive: press the sidebar button and the thing retracts (also hiding the select mode button, since there would be nothing to select), press again, it reappears
One caveat however: rendering the graph takes noticeable time, thus making the transition animation quite sluggish.


SjoerdB93:
Nice, this works quite well. 
The animation is quite sluggish, but not so sluggish that it's unworkable.

I have also disabled the coordinates on the bottom-right corner. Those only show the coordinates of data in the "top-right" axis, which is the data that is plotted using the right-hand y-axis and the top-axis, meaning it's utterly broken in 98% of the cases while giving an awful experience for small window sizes (it tries to increase window size to accomodate for this coordinate label).

This latter can be fixed, I could for example limit the precision. But as of right now it doesn't work anyway and it has to be re-implemented at one point. Perhaps a dedicated "Data selection" button in the toolbar (I've done that before in an older project, but I'll have to check  how it works when using more than just the regular axes on the left and the bottom), and then make a little pop-up when one hovers over the graph. Like MatLab does:

image

But that's a seperate issue than this side-bar of course. I just mentioned it here, because it was when testing this sidebar button that I got annoyed by these coordinates that try to resize the window.

SjoerdB93:
I also removed the size requests for drawing_layout. Those existed to give the window a minimum size so that there was always room for these coordinates. But as I disabled them, there's no reason to limit the window size by anything else than the size of the toolbar.

SjoerdB93:
I've played around a bit more with this, and I really like the ability to retract this sidebar. Feels really nice, and not only does it allow for small window sizes, but also for larger graphs. Nice work :)

I also feel like it has become more smooth since removing these coordinates, but that could just as well be a placebo effect.
[Screencast from 2023-01-30 07-50-14.webm](https://user-images.githubusercontent.com/68477016/215408137-8d81b978-2c78-41c7-a8dd-9520645a7930.webm)

SjoerdB93:
It seems like it's smoother _after_ loading a dataset. I think it also slows down again when using the top axis. So my guess is that it is slower because the top axis is loaded during initial setup (hence all labels being visible in initial setup), and then faster again after all unused axes are hidden when a single dataset is loaded.

I noticed a slow-down in the selection-tool before when using the top axis. Not exactly sure why, because the other axes behave nicely. It's something I should investigate at one point. It could be a bug in matplotlib, their usage of multiple axes is a bit hacky (each axis is a pair of X-Y axes, and to create anything that is not left-bottom, I need to clone and mirror the axis using the twinx and twiny function). It could also just be that I implemented something slightly wrong there.

It's not really breaking (everything still works fine, but moving things are a bit more choppy), and the usage of a top axis is quite rare. But it is on my list of things to investigate at one point.

PR #44, opened 2023-01-29
use meson version string for about window

cmkohnen:
As the title says


SjoerdB93 (approved):
Thanks, that's a lot more convenient indeed.

Issue #48, opened 2023-01-30
PIP doesn't update when adding new data


cmkohnen:
As the title says.
Opening a PIP-window and adding data doesn't update tha graph inside the PIP window.

SjoerdB93:
Yeah, the thing is that the pip window gets rendered on it's own canvas, and all actions are plotted on the original canvas. It is probably possible to just refactor self.canvas towards a list that contains all canvases, and each time we plot on this canvas, we just loop over all these canvases. That does feel pretty ugly though. 

I think we should also consider the option to just remove the button as well, as suggested in issue #36, given retracting the sidebar has the same effect. I'm not 100% decided on this.

PR #49, opened 2023-01-30
Redesign UI Elements

cmkohnen:
Simplifies UI and gets rid of unneeded components. Now the set_child/content functions are used to add toolbar & graph to the window

Adds a button just next to the toggle_sidebar button to toggle the toolbar (implemented using AdwFlap again)


SjoerdB93:
Also, I see now that in this current implementation the PIP-graph loses the toolbar. If we decide to keep it (see issue #48), I think it should have the toolbar as well.

cmkohnen:
I've added back the toolbar to PIP.
I also changed the Add Data Button to a split button, to have the "normal" option more accessible.
I added a little spacing to group the buttons in accordance to the following:
>  Buttons can be grouped within a header bar, in order to communicate a functional relationship between them.

I also moved the two toggle buttons to the right as that separates data and action based buttons from UI related ones. In the process, I also moved the save function to the primary menu.

SjoerdB93:
Thanks, that sounds good. In the very first build, I actually tried to use a split button for the Add data function but couldn't get it to work. But that was a while ago, and this was my first GTK4 project 

 I'll take a look later tonight when I've got access to my computer.

bertob:
FWIW my general feeling with the toolbar on the right side is that it should probably just go away and those options should be integrated in other ways. For example:
- Reset view could go into a view menu (like e.g. in Text Editor)
- Log/Linear axis toggles could go in a view menu as well
- Plot Settings could go in the headerbar
- Save graph could go in the primary menu (maybe as "Export"?)
- I'm not sure what the use case for opening the graph as an independent window is, depending on that it could either live in a view menu, primary menu, or in the headerbar
- Zoom and Pan should perhaps be top-level tools in the sidebar, like selection? That's how other graphics apps do it.

cmkohnen:
> FWIW my general feeling with the toolbar on the right side is that it should probably just go away and those options should be integrated in other ways. 
I agree, the design you described here and [here](https://github.com/SjoerdB93/Graphs/issues/14#issuecomment-1409206495) would be an improvement.

If we were to introduce a view menu, the current sidebar toggle should go in there as well.

As for now I've implemented a quick (and quite hacky feeling) way. As of now, a solution has to be found to replace pan, back, forward, zoom & save

SjoerdB93:
I quickly tried this latest commit, and this custom toolbar seems mostly
functional. I agree that moving this to some kind of "View" mode on the
side panel would be a much more elegant solution, and would also feel much
more modern than a toolbar.

There's some nice widgets in GTK to get exactly this kind of behavior. I
think a Stackswitcher (if it's still called that, I used it once in an old
GTK3 project at least) would work nice.

Den mån 30 jan. 2023 21:24Christoph Matthias Kohnen <
***@***.***> skrev:

cmkohnen:
I have now moved all the actions invoked by the toolbar to global app actions, which allows to invoke these anyway UI wise we might want to chose. Sadly this broke interaction in the PIP mode. I also implemented @bertob's suggestions both here and [here in parts](https://github.com/SjoerdB93/Graphs/issues/51) (only the header icons for preferences are switched). Feel free to request any design changes, as I am no expert in UI design.

The question now is, do we want to keep the Toolbar? And do we want to keep the PIP, as it can be replicated by simply hiding the sidebar (and toolbar) now.

Screenshot of current look:
grafik

bertob:
Nice, I think this is going in a good direction!

Some notes:

- With the view menu the toolbar is no longer needed, so I'd just drop it rather than having a toggle option
- The sidebar toggle setting should be an actual toggle menu with a check and a label like "Show Sidebar"
- The four tool buttons in the sidebar should have a pressed state when they're selected (not all of them do at the moment)
- I'm not sure about "Cut" being one of those four, since it's not really a mode, while the other three are. I think it'd fit in with the "Adjust Data" group.
- I'd try and have the tool buttons fill the column width by making them hexpand
- I assume "Change xscale"/"Change yscale" are the logarithmic toggles? If so, I'd have them have in-menu checks and labels like "Logarithmic X Scale"
- "Enter PIP" is not very clear. My personal feeling is that this feature could just be removed, but if you want to keep it it needs a better label
- All labels should be title case (e.g. "Go Forward")

cmkohnen:
> With the view menu the toolbar is no longer needed, so I'd just drop it rather than having a toggle option
> "Enter PIP" is not very clear. My personal feeling is that this feature could just be removed

I agree with both, but would want to get an opinion from @SjoerdB93 before removing.

> I'm not sure about "Cut" being one of those four, since it's not really a mode, while the other three are. I think it'd fit in with the "Adjust Data" group.


The two are exclusively linked together, so separating them might not be the best option, having them as buttons or combined into a single one as described [here](https://github.com/SjoerdB93/Graphs/issues/14#issuecomment-1409198193)

> The sidebar toggle setting should be an actual toggle menu with a check and a label like "Show Sidebar"

Is there an easy way to do this using xml? If so I have failed to find it, since this is what I also had in mind.

cmkohnen:
Regarding the select and cut, I've got an idea, but will try to implement that tomorrow

SjoerdB93:
This is starting to look really nice, great work. Regarding the PIP mode, the feature had its purpose in the old design but now with the retractable sidebar it's kinda redundant. I think it can be removed.

SjoerdB93:
So my personal access token on Github expired a while ago, and I thought I could quickly made some changes in review mode. So that was a mess, apologies for that. Set up my access token again, and pushed my changes as a proper commit.

Didn't want to change to much of what @cmkohnen was working on, as that would probably be ineffective if we were both doing the same thing, but I thought I'd reorder the Equation row a bit to address issue #54. It's now pretty much as suggested, which pragmatically works quite well. But I couldn't get the color button to shrink. See image.

image

I also removed the "Selection mode" button (as that doesn't makes sense with the toggle always being visible), and I removed the pip-mode (which then also fixes issue #48) . Internally, I also changed the way it handles selection. It used to check whether a specific row has a selected status, to determine if a data-set is selected. And then I kept the toggle status in sync with that (this behaviour was inherited from the fact that the toggles didn't use to be there). Now it just works with the toggle status directly, which is much cleaner.

I didn't remove the toolbar yet, as not to be too much in the way of @cmkohnen  here, but I agree that it should move to the sidebar instead.

SjoerdB93:
> I assume "Change xscale"/"Change yscale" are the logarithmic toggles? If so, I'd have them have in-menu checks and labels like "Logarithmic X Scale"

Yes, these are the toggles to change the scaling on the respective axis. Either way, I think it's good to have them some toggles. In the longer term I'm thinking of adding a square root scaling as well (there's some advantages in certain disciplines of square root scaling), so personally I kinda see them as "Change scaling" toggles.

Also, for me personally, a logarithmic Y-axis is kinda the standard. So in my use-case, I see the toggle as a "change to linear scale" button rather than a "change to logarithmic scale" button. The point I'm trying to make here is not that it should be "change to linear scale" instead, but that it either should indicate something like "change scaling", or the label should be dynamic and change label depending on the active scale. 

I didn't mind the alternative icons made though where it just had the <--> arrow with LOG on top of it. As that icon kinda signals me that the button is directly coupled to the scaling of that axis, regardless of what my current scaling is. The little LOG label changing on depending on the status could be nice, but for an icon I think a static symbol still works potentially.

SjoerdB93:
This is turning out to be quite an ambitious PR. 
As this will get rid of the toolbar, this will solve issues #50 and #51, and I feel like adding a padding on the sidebar should be implemented here as well. So I coupled those issues to these PR, so they'll close once this is merged.

Not sure if @cmkohnen has the rights to couple issues to his own PR, but in case you do. You can do this with the "Development" tab on the right.

cmkohnen:
> Not sure if @cmkohnen has the rights to couple issues to his own PR, but in case you do. You can do this with the "Development" tab on the right.

I do in fact not have permission

cmkohnen:
I've changed the way selecting and cutting are displayed:
By default it looks like this:
grafik
And when going in select mode it looks like this:
grafik

SjoerdB93:
> > Not sure if @cmkohnen has the rights to couple issues to his own PR, but in case you do. You can do this with the "Development" tab on the right.
> 
> I do in fact not have permission

Ah okay, I'll try and figure that out soon at some point then (right now I should mainly get back to work, or at least the thing I'm actually getting paid to do). At this point I do consider you a trusted contributor to this project. Also, just wanted to say that I really dig the way the "Cut" action now appears contextually when selection mode is active. Which is in general also more in-line with modern design where options get out of the way when irrelevant, and are mainly shown contextually when they make sense to be there.

I just made a screenshot as a reference, but you literally just pasted a screenshot while I was writing this message. This really tackes the issue I described in my post in issue #16. The select/cut behaviour is really distinct from all other actions, and having it there as a regular button is not consistent. My initial thought was to couple it to a single button, but that doesn't give an option to cancel. This is a really nice solution, I really like this!

cmkohnen:
I've increased the padding as suggested in https://github.com/SjoerdB93/Graphs/issues/57

grafik

bertob:
Nice, getting there!

I'm still not totally sure I get how "Cut" is different from other actions. It basically just deletes the selected points, so that area becomes a straight line, no? If so, how is that different from applying any other action to the selected area?

By the way, I was just trying to apply actions to just a part of the functions, and it didn't seem to work, they were always applied to everything and the selection was lost. I assume that's a bug?

Also, it seems to be possible to select more than one of the three tool buttons at the same time currently:

image

cmkohnen:
> I'm still not totally sure I get how "Cut" is different from other actions. It basically just deletes the selected points, so that area becomes a straight line, no? If so, how is that different from applying any other action to the selected area?
> By the way, I was just trying to apply actions to just a part of the functions, and it didn't seem to work, they were always applied to everything and the selection was lost. I assume that's a bug?

That is actually intended (as of now). Select and cut select a span of data (from selected items) and cut them.
All other actions apply to the item as a whole and not specific to a select span of data

bertob:
> That is actually intended (as of now)

Aaah ok, then it makes sense.

SjoerdB93:
> I'm still not totally sure I get how "Cut" is different from other actions. It basically just deletes the selected points, so that area becomes a straight line, no? If so, how is that different from applying any other action to the selected area?

Just to clarify, it actually does delete the data points that are within the selected area. There's no data within the straight line. It's just that a typical "Line plot" draws a line through all data points. But if you make a "normal" scatter plot by just turning the line style to "none" and turn on markers, then you see that these data points are deleted. The cutting action is indeed different by it's inherent coupling to the select function. Either function cannot exist without the other. Selection is specifically only used (as of now) to cut data. And in order to cut data, you need to have highlighted the are you want to cut.

So the highlight itself is specifically only applied to the selected area. In principle it could be possible to make it so that all other actions also only apply to a selected area, but that's not something I have considered. I am unsure if it's useful either. But if there's a demand for that, then I can imagine implementing that. As of now it is possible to only apply a transformation on a specific part of the data using the "Custom transformation" function, but that would require a bit of python knowledge and a list comprehension so it's not something I expect an average user to do. 

None of the actions are applied to unselected data sets, which is also by design. If you deselect a data set, it is "safe". See attached screencast. Only exception is the "shift vertically" function, but as discussed before that one should probably be redesigned.

[Screencast from 2023-01-31 12-20-52.webm](https://user-images.githubusercontent.com/68477016/215748073-629dafbf-1264-4fc9-a994-c940423daf0b.webm)

SjoerdB93:
In the current menu it has the two options "Save as text" and "Export", the former saves the data as text files, and the latter brings up a save dialog for the plot itself.

It would probably make more sense to rename these for clarity sake so it's clear what the difference is. My proposal would be:
- Export Data
- Export Figure
But other alternatives could probably work as well

image

cmkohnen:
I've reordered the buttons inside the expandable categories:
grafik
However, when trying to add images, the text simply vanishes. And since gnome.org and the gtk documentation are currently down, I couldn't get it to work

SjoerdB93:
I was just randomly thinking to myself that we can implement the pan/zoom functions, by just borrowing these from the NavigationToolbar class without actually adding the toolbar itself. Only to discover that this is _exactly_ what this implementation of dummy_toolbar does, so we kinda were thinking of the same solution there. I poked around a bit why the implementation didn't work, and defined the dummy_toolbar in the `load_empty `function instead.

The reason why the old implementation doesn't work, is because "reload_plot" completely reloads the entire plot including the toolbar. This function is called on some specific circumstances where a refresh doesn't do it. One of these circumstances is when adding (not just editing) new data to the plot. So when new data was added, this functionality broke. By defining the toolbar in the `load_empty` function, the dummy toolbar gets correctly coupled to this newly (re)loaded canvas.

cmkohnen:
> Only to discover that this is _exactly_ what this implementation of dummy_toolbar does

There is one more thing, the current dummy toolbar does, though be it not visible in code here: It provides rubberband support for the zoom mode. This only works with the gtk backend and not the default Toolbar2 backend

SjoerdB93:
> grafik I'd say, there is enough space to house the button

Yeah personally I'd be in favour of having the button there. 
To me, I feels like options of changing the view should be accompanied by an easily visible way to revert these changes.

cmkohnen:
As saving the saved directory path doesn't work well in flatpak and returns a /run/user/... path ever so often, there is no need to implement such functionality

cmkohnen:
Any suggestions on what Icons to use for the actions? I struggle to find fitting ones for most of them.

SjoerdB93:
That's a very good question, I've been thinking about this a bit before back when I originally implemented a complete grid UI.
The logic of just using labels was basically that there's no way to implement an icon for e.g. Fourier transformations without it being confusing for most people.

Where possible, we should try and use icons available in the GNOME icon library, but this is not really an option for many (if not most) actions. So I guess most icons will have to be custom made. 

If I have time this evening (which I'm not sure about), I'll take a look if I can make some drafts for the icons.

Specifically for Fourier transformations, what it does is to take a signal in a time domain (e.g. a sinoid wave) and transforms it to the components that are in frequency space. So a sinoid with a frequency of 20 Hz, will just show a sharp peak at 20 Hz in the Fourier transform. There's no icon in the icon library describing that. One icon I found describes it like this:
bild

This is from the Noun project, the icons are free to use but require attribution. I think we should use exclusively open licensing though, so I just post is as specific example, not as suggestion to use exactly that specific icon. 
For "Derivative"  I was thinking either a stylised dy/dx, or a curve with a tangent to it. (Derivatives give the slope of a curve). For "Integral, I guess the S-shaped integral sign with dx next to it. Like this, but without the dx:
bild

I'll have to think a bit about each specific action, and see if I can find something from the icon library that comes close. But again, I think for most of them it has to come down to custom icons unfortunately.

cmkohnen:
For the Fourier transformations maybe we could use a bar chart or histogram, as one couldn't really tell any difference on small scales.

For Derivative and Integral it should either be `f'(x)` and `F(x)` or `dy/dx` and `∫ydx` in my opinion

For Transform maybe something like `x -> x'`

I don't think it would be to bad to choose icons, that don't fit 100% since these might be more advanced features

cmkohnen:
Side note: the FFT functionality is the reason, I chose to work with this app in the first place, so no need to explain the mathematics, but thank you anyway

SjoerdB93:
I quickly added an icon to the FTT button, just to see how that works. We need to add a GTKBox in the button, with a label and a GTKImage appended horizontally. That works quite well.

Not really sure why it doesn't center yet. Also the icon I use is not really appropriate here, I just used it as proof-of-concept.

SjoerdB93:
Moving the icon to the left-hand side as in the mock-up, and putting it more in the center (did this quickly by adding a margin), it would look like this for one button:

image

SjoerdB93:
I've added icons (not labels, despite the commit message) to all of the buttons. But all of them are placeholders (they're identical), so we still need to fix the icons themselves. But that's just a matter of changing the icon name itself.

Here with all categories expanded:
image

SjoerdB93:
Haven't worked on the icons themselves now, but I did add the feature that double-click opens the plot settings for the relevant row. It's a thing I've wanted to implement for a while now, but with the previous implementation of the rows (as GtkListBox) I got stuck in a bug where the row kept stuck in an active state. It works wonderfully now, see video:

Also, this probably should have been a separate PR that I should have merged with main, but as we're working on this branch anyway...

[Screencast from 2023-01-31 21-34-43.webm](https://user-images.githubusercontent.com/68477016/215877768-6b1351b6-809c-4143-93e1-ba6e0295c73a.webm)
.

SjoerdB93:
As I look in the video, I see that bug actually persists. In my initial testing, I double clicked slow enough that the bug didn't get triggered.

 I have some ideas on how to potentially fix it but I'll check that out tomorrow, also when the GNOME servers are up so I can look in the documentation as well.

cmkohnen:
Maybe, as to not broaden the scope of this pr any further, it is best, to finalize the icons and fix any bugs we might spot, that have been introduced due to these changes, but not implement new features, as the changelist is already quite vast.

Since the current UI design is something I'm quite happy with, I would suggest merging this and do any small tweaks in a separate pr.
I've got a few features (only small additions) in mind, that would depend on this, but would not want to introduce them here. That would also make it easier to work on thinks without getting in each others way.

SjoerdB93:
Yeah I agree, in general a PR should be limited to one feature/issue anyway.

SjoerdB93 (approved):

SjoerdB93:
Regarding the UI, I share the opinion that the current design is at a good spot. We can fix the icons potentially in a separate PR. As we really shouldn't treat this PR as just a development branch.

SjoerdB93:
We could have kept the icons as a blocker for this PR to avoid unfinished design in the main branch, but I'd suggest to have those separately as well. 

The size of this PR had gotten quite big, to the point that any changes to the main branch should keep this PR in mind as well, which was also why I did my latest commit on this PR as I was using it as a base for development. Basically the majority of files had at least some changes at this point. Hence me merging it already now.

cmkohnen:
An option would be to use Githubs milestone feature

cmkohnen:
It would allow to keep track of changes, thinks left to do, and issues to solve. I have a bit of time tomorrow morning, I could set one up for an upcoming 1.5 release. That would be a lot cleaner and effective in my opinion.

SjoerdB93:
Yeah that would be probably be a good idea. It would also allow for a slighty more organized flow, so that it if is clear what needs to be done before next release.

PR #63, opened 2023-02-01
Fix stuck rows

SjoerdB93:
This fixes a the bug where the rows get stuck in an active state on double click.

This is a known upstream GTK issue where rows can get stuck upon a double click action. I'm not completely sure where the problem lies, but it gets triggered when you double click an item and immediate pop over a modal window. I believe it has something to do with the window getting triggered on click, while the fade back to the non-active state for the row gets triggered upon release. But due to this modal window freezing the UI in the parent window, this release state is never triggered.

The workaround was to just trigger the modal window upon double release instead of double click. Where I had to use another workaround to emmit a release signal on GtkGesture. See bug report [there](https://gitlab.gnome.org/GNOME/gtk/-/issues/4939). Either way, this feature works perfectly now. But just wanted to merge it with a PR instead of comitting straight to main.


SjoerdB93:
See video, double click actives plot setting menu for the data set that is clicked on. Streamlining many workflows significantly. Single click does nothing.

[Screencast from 2023-02-01 09-37-52.webm](https://user-images.githubusercontent.com/68477016/215992384-9fa104d2-60ac-49f8-a9a7-0ff82113f2da.webm)

Issue #45, opened 2023-01-30
Selection tool doesn't work when x-axis is logarithmic


SjoerdB93:
When data is selected using the selection tool, it obtains the wrong data-span if the x-axis is logarithimic.

The cause of this bug is because the selection tool has to be coupled to a specific axis, which happens to be the top x-axis. Even when no x-axis is shown or used, it is still there (albeit hidden) with coordinates going from 0 to 1 in the entire range.
It then calculates what the corresponding coordinates are on the bottom x-axis and uses those data points for selection on data that are plotted on the bottom x-axis. This allows to cut data-sets simultaniously even if both datasets are plotted in on different x-axes.

I just realized however that the calculation here assumes that both axes are scaled linearly, and the wrong set will therefore be selected if any of these axes is not linear. See attached video:
[Screencast from 2023-01-30 09-12-32.webm](https://user-images.githubusercontent.com/68477016/215422702-4fd672c9-1fd9-4321-89f7-95172a66a569.webm)

I will redo the mathematics for the special  cases that any of these axes are not linear.

PR #65, opened 2023-02-01
WIP: Perform item operations only on highlighted area

SjoerdB93:
This is still WIP, but as a proof of concept I've implemented a solution that only performs item operations on the selected area. It also contains a slight bugfix with the selection mode, which didn't get applied as attribute to the dummy toolbar. (That didn't lead to additional bugs, only with this selection mode).

The code has some pretty rough edges, and I have some ideas how to improve this. But again right now it's only a proof of concept. But I'd thought I'd throw it out here. I've commented in-code how it works, and also where I see a potential bug. Basically it will only work right now if the x-data is in an ascending order, but I've got ideas to fix this. This was just a quick and dirty implementation as a proof of concept. I've also only implemented this for the Translate Y functionality, but I'm kinda hoping to see if I can find a way to do this without too much code replication. Finally, this is also still affected by bug #45. Which is probably next up on my priority list, and I consider it definite a blocker for version 1.5.

Feel free to work on this, I'll work further on this this evening at the earliest. 

[Screencast from 2023-02-01 12-20-36.webm](https://user-images.githubusercontent.com/68477016/216029847-dcaa1471-0fe3-4309-8a05-40287fa76343.webm)

### DIFF
--- a/src/plotting_tools.py
+++ b/src/plotting_tools.py
@@ -6,6 +6,7 @@ import matplotlib.pyplot as plt
 from matplotlib import colors
 from matplotlib.backends.backend_gtk4agg import (
     FigureCanvasGTK4Agg as FigureCanvas)
+from matplotlib.backend_bases import _Mode
 from . import graphs, utilities, rename_label, toolbar
 from matplotlib.widgets import SpanSelector
 from cycler import cycler
@@ -18,9 +19,6 @@ def define_highlight(self, span=None):
     Create a span selector object, to highlight part of the graph.
     If a span already exists, make it visible instead
     """
-    if (not self.highlight == None):
-        self.highlight.set_visible(False)
-        self.highlight.set_active(False)
     #This lamdba function is not pretty, but it doesn't accept a "Pass"
     self.highlight = SpanSelector(
         self.canvas.top_right_axis,
@@ -33,15 +31,13 @@ def define_highlight(self, span=None):
         drag_from_anywhere=True)
     if span is not None:
         self.highlight.extents = span
-        self.highlight.set_visible(True)
-        self.highlight.set_active(True)
 
-def toggle_highlight(shortcut, _, self):
+def select(shortcut, _, self):
     """
     Toggle the SpanSelector.
     """
     if self.main_window.select_data_button.get_active():
-        set_mode(self, "")
+        set_mode(self, "none")
     else:
         set_mode(self, "select/cut")
 
@@ -189,9 +185,13 @@ def reload_plot(self, from_dictionary = True):
     win = self.props.active_window
     graphs.load_empty(self)
     if len(self.datadict) > 0:
-        define_highlight(self)
         hide_unused_axes(self, self.canvas)
         graphs.open_selection(self, None, from_dictionary)
+        if (not self.highlight == None):
+            self.highlight.set_visible(False)
+            self.highlight.set_active(False)
+            self.highlight = None
+        set_mode(self, self._mode)
         set_canvas_limits_axis(self, self.canvas)
     self.canvas.grab_focus()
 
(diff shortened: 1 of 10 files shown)


SjoerdB93:
So I worked a bit more on this anyway. It seems to be working quite well for almost all operations with two exceptions:
- Shift vertically. But this function has to be rethought a bit, in principle we could just apply the entire logic to this one a well though. Apply to the highlighted area only, and for the selected datasets only
- Custom transformations. I just haven't gotten around to this yet.

Right now, I apply to the selected area for most actions such as translations and multiplications etc.. However in the case for FTT and inverse FTT, it doesn't make sense to apply to a specific area, while keeping the rest intact. So instead what that does, is that it applies to Fourier Transform to that area, and discards the non-highlighted data. 
The same is true for Integral and Derivative, it takes the derivative of the selected area and discards the rest of the data. That feels to me like the most logical way to use this, but I'm not sure if it's completely consistent. So opinions on how the highlighted area should affect each button is appreciated.

Finally, the code itself probably needs some work, and comments are a bit ugly now as well. Again, still WIP and it probably needs lots of testing as I am expecting bugs in this code. So quite more extensive testing will be needed anyway before merging.

src/plotting_tools.py
```diff
@@ -413,6 +413,7 @@ def set_mode(self, mode):
         cut_button.set_visible(True)
         highlight.set_visible(True)
         highlight.set_active(True)
+        self.dummy_toolbar.mode = "select/cut"
```
cmkohnen, line 416: setting the mode this way is a bad idea.
matplotlib's NavigationToolbar2 uses an enum to set its mode. These values can only be `_Mode.PAN` ("pan/zoom"), `_Mode.ZOOM` ("zoom rect") and [_Mode.NONE](url) (""). This enum allows for string comparison e. g. `if (mode == "pan/zoom"`, but setting the mode to any string can cause issues with some internal functions.
Rebasing on https://github.com/SjoerdB93/Graphs/pull/67 and using self._mode for all checks would be a better solution.

cmkohnen (changes requested):

SjoerdB93:
[Screencast from 2023-02-01 12-20-36.webm](https://user-images.githubusercontent.com/68477016/216294530-66eeca20-b393-4a04-9d7c-6deb4d42d205.webm)

The other design question we should ask here, should data also be sorted such that the x-axis data is continiously increasing? I do this for the Fourier Transforms.

This doesn't change the values of the data itself, and the difference will only be visible with line graphs (which is the default) instead of scatter plots. But if we don't sort it, then it means that lines will be drawn criss-cross through the dataset, as it just draws the line "chronologically". This may make sense as it's closer to the original data, but I'm inclined to sort them. See video with the current behaviour where it doesn't sort.

SjoerdB93:
So this now works for all operations. For all operations (including "Shift Vertically") the behaviour is the same:
- The operations are only applied on data sets that are active (so selected)
- The operations is only applied on the selected area.

In the case for FTT, IFTT, derivatives and integrals, the operation is applied on the selected area and the rest of the data set is discarded. This is because otherwise these operations make little sense. You can't both have fourier space and real space in the same x-axis, this is physically undefined. 
Likewise with derivatives it becomes pretty weird, but in the case of derivatives and integrals I am open to the idea of keeping the old data there anyway.

In the current implementation, any operation that can change the x-position of the coordinate, the data set is re-sorted afterwards so that it's continiously increasing. See previous video and compare it with the current one.

[Screencast from 2023-02-02 13-19-43.webm](https://user-images.githubusercontent.com/68477016/216324306-4c0b8669-6535-476d-a3a9-4faaaafa5a2d.webm)

SjoerdB93:
To add here, a peak "bouncing" back to the previous state before the operation was applied, is not a bug. That's just me pressing ctrl+z. Should probably have pressed the undo button for sake of clarity in the video.

cmkohnen (approved):

PR #67, opened 2023-02-01
change mode switching logic

cmkohnen:
Switching modes and everything related to that is now handled by plotting_tools.set_mode().
We no longer simulate button presses on the dummy_toolbar but rely on functions deeper down the stack.
Thanks to this, mode switching is now a tad bit faster and reliable. 
Selecting a mode persists through graph.reload as well now.


SjoerdB93:
Yes, this seems cleaner than string comparisons. I've tested this (with the updated ui files which I broke :() and all modes also seem to still work.

SjoerdB93 (approved):

PR #70, opened 2023-02-01
WIP: Fix selection on log scale

SjoerdB93:


SjoerdB93:
This fixes issue #45. But still has some issues with edge cases with multiple axes (on top and bottom) while both have different scaling. Will look into that later.

SjoerdB93:
I think I know why it doesn't work in those edge cases, I'm assuming the top axis goes from 0 to 1, should convert it to actual percentage of the axis instead. Again, will take a look at it tomorrow.

SjoerdB93:
This works now for all combinations, also when translating/multiplying individual axes only.

Issue #62, opened 2023-02-01
Make primary toolbar buttons exclusive


bertob:
An important aspect of toolboxes in creative tools is that tools are modes, and that there's always exactly one of them that's active. This way it's always predictable what's going to happen when you click on the canvas. Currently this isn't the case, because
- Reset is not a mode, but just a singular action
- It's possible to have none of the modes be active

I'd make the tools exclusive buttons where one is always toggled, ideally using [this widget/pattern](https://gitlab.gnome.org/GNOME/libadwaita/-/issues/212) but if that's too hard, a regular stack switcher would work for now.

SjoerdB93:
There's basically two modes there then, zoom and pan. And then there's the reset action. Would you then suggest to have one switcher with "Neutral mode", "Zoom mode" and 'Panning mode", plus the regular reset button put there somewhere?

You could make the neutral mode to reset as well, but I'm not sure how intuitive that is to most people.

SjoerdB93:
Just to add to this. The current implementation with just regular toggle buttons for zoom mode and pan mode is the standard design language in these kind of applications. In fact, some applications I've seen have even worse implementations for this where this needs to be triggered by a right-click menu, and then selecting a mode. (Which I'm not a fan of)

But I'll be honest. Most similar applications have a god-awful design. Even expensive proprietary graphing applications such as OriginLabs have a design that is straight from the early 2000's, and we've paid a $800 license fee to use it. See attached screenshot:

![image](https://user-images.githubusercontent.com/68477016/215993855-b6e3a4ed-3882-4b5c-a7e6-e8e308ee03d1.png)

 Which is no disrespect to the respective developers, but my theory is that it's a result of these applications being built by scientists and engineers who often have a more pragmatic approach to this. Also, in particular OriginLabs is very powerful and has a lot of features. It's inherently difficult to add so many features while keeping the design itself sane. This is also why my view with Graphs is not to make a competing product to these kind of tools, as they have their place, but have something more of a middle ground that fits with most use-cases. So I'm really not trying to talk down on these kind of applications. What I am just trying to say is that I'm happy with borrowing some of the design language from creative tools rather than engineering tools, as the latter are not really what I consider the design we should be aiming for.

cmkohnen:
> There's basically two modes there then, zoom and pan. And then there's the reset action. Would you then suggest to have one switcher with "Neutral mode", "Zoom mode" and 'Panning mode", plus the regular reset button put there somewhere?

[...]

Issue #61, opened 2023-02-01
Rethink Cut placement


bertob:
I'm still not sold on the "cut" action being added to the primary tools. Even if it has a different scope from other actions, it makes more sense to me to group it into one of the categories than having it bolted onto the selection button.

![image](https://user-images.githubusercontent.com/1908896/215919827-50961109-98ab-4687-aa5c-f72ca081e096.png)

How about this: We add the button to the bottom of the first group, but have a very explicit label explaining the scope, e.g. "Cut Points in Selection" and have the button disabled when there's no selection.

SjoerdB93:
I'm pretty open to moving it to the first group. Mostly because it's quite distinct from these toolbar options. Reset, pan and zoom are all tools to interact with the plot view. Select and cut is a way to manipulate the data. So i see the argument of moving it.

We could still make the but button contextually appear,  with "select and cut" as label but I think that would make the button too wide so just disabling it when you can't use it is probably a better solution.

SjoerdB93:
Looking at the "Adjust Data" group. Adding this one there would provide an odd amount of items, which will look awkward in a 2-column grid panel. One option would be to add it as a 2-column wide row, giving it plenty of space to expand with a contextual cutting symbol upon activation, while having the label "Select and cut".  

I'll may experiment with it this evening or so after work.

bertob:
I was just thinking, if "Cut" is the only thing you can actually do with the selection, perhaps it doesn't make sense to have "Select" as a distinct action and it should just be a single "Select and Cut" action where you click the button, do the selection, and it deletes those nodes?

bertob:
> However, I've been working on an implementation of having all actions applied to only the selected area when Selection mode is activated. See PR https://github.com/SjoerdB93/Graphs/pull/65. If we decide to implement that behaviour, I think it makes sense to have Selection mode as it's own mode, and move the cut action to the top group ("Adjust data"), and just deactivating the button when not in selection mode..

That makes sense, yeah. In that case the 3 tools/modes idea from #62 would work cleanly.

SjoerdB93:
Yes exactly. Then for the "reset view" button, we can probably have it only in the "View" menu (the one next to the primary menu) and couple it to a keyboard shortcut as @cmkohnen suggests, coupling keyboard shortcuts to an action is quite trivial to do. 

[...]

PR #74, opened 2023-02-02
redesign ui elements

cmkohnen:
This redesigns UI, and underlying code, around mode switching.
Cut data is now moved to the manipulate category as discussed in https://github.com/SjoerdB93/Graphs/issues/61
It is now no longer possible to have none of these buttons selected, see https://github.com/SjoerdB93/Graphs/issues/62
Reset view button is removed, thinking of improving a lot of actions via keyboard shortcuts in a different pr

It looks like this:
grafik

Feedback appreciated, if it should resemble something like [this](https://gitlab.gnome.org/GNOME/libadwaita/-/issues/212) like @bertob suggested in https://github.com/SjoerdB93/Graphs/issues/62 as discussion around this widget seems to be ongoing and we would have to implement our own version

On the backend side:
I've done a little cleanup and regrouping, to keep related code closer together.
I've also used an enum for the mode and updated all references


SjoerdB93:
I think this works quite well, I've looked a bit at what the Stack Switcher looks like these days with LibAdwaita, but it's basically the same as these toggles but without the padding in between. Personally I feel like this works just as well UI-wise.

I'm still a bit conflicted on whether the "Normal mouse pointer" mode should be coupled to a "Restore View" action. This would give a quick and easy way to restore the view, while still giving a consistent experience ("Mouse pointer mode" always shows the graphs in a full zoomed-out mode). On the other hand, I'm not 100% sure if having the graph reset view when going into neutral mode is the best course of action, I can also imagine some might find that annoying. And binding the "Restore View" to a keyboard shortcut would indeed get the job done as well (also it's available from the menu). But it feels natural to me to have a quick way in GUI to reset view, and we could just as well bind that to the Restore view..

I'm open to either comments on that.

src/main.py
```diff
@@ -37,6 +38,11 @@ def load_preferences(self):
     def connect_actions(self):
         self.create_action('quit', self.on_quit_action, ['<primary>q'])
         self.create_action('about', self.on_about_action)
+        self.create_action('toggle_sidebar', self.toggle_sidebar, None)
+        self.create_action('mode_none', self.set_mode, None, InteractionMode.NONE)
+        self.create_action('mode_pan', self.set_mode, None, InteractionMode.PAN)
+        self.create_action('mode_zoom', self.set_mode, None, InteractionMode.ZOOM)
+        self.create_action('mode_select', self.set_mode, None, InteractionMode.SELECT)
         self.create_action('preferences', preferences.open_preferences_window, ['<primary>p'], self)
```
SjoerdB93, line 46: We should probably bind some shortcuts to some of these actions, but as said it's probably best to do that in a separate PR

src/main.py
```diff
@@ -88,10 +90,11 @@ def do_activate(self):
         graphs.load_empty(self)
         graphs.disable_clipboard_buttons(self)
         graphs.enable_data_dependent_buttons(self, False)
+        self.set_mode(None, None, InteractionMode.NONE)
         win.maximize()
         win.present()
 
-    def on_about_action(self, _widget, _):
+    def on_about_action(self, widget, shortcutself):
```
SjoerdB93, line 97: I'd go for just shortcut instead of shortcutself

SjoerdB93 (approved):

SjoerdB93 (approved):

cmkohnen:
We could use an AdwSplitButton for the view menu und have the primary function be to reset the view. Then there'd still be a quick option to reset the view in UI, but it wouldn't confuse or annoy users

SjoerdB93:
> We could use an AdwSplitButton for the view menu und have the primary function be to reset the view. Then there'd still be a quick option to reset the view in UI, but it wouldn't confuse or annoy users

I think that would be a really nice solution, at least for me that would work well!

SjoerdB93:
This works really well on my machine, and fits nicely in my workflow at least without needing to bind an action to a specific mode (mouse pointer mode in this case). For my part this can be merged.

PR #79, opened 2023-02-02
Add combine data feature

SjoerdB93:
I don't think we should add too many features before the 1.5 release, but this is something that I've kinda wanted to add for a while and it works quite well. Basically it lets you combine multiple data sets to one, which has a fair amount of use cases. For instance when we do neutron experiments, we typically divide our measurements up into different smaller ones with different parameters. (Scanning slower where there's less signals). This lets you stitch these loose measurements together easily. See video
[Screencast from 2023-02-02 21-46-49.webm](https://user-images.githubusercontent.com/68477016/216446407-378a384b-79ee-4d6f-a1cc-459f7bba0c57.webm)

This is still a WIP, as it doesn't work on highlighted data yet, but that's easy to add


cmkohnen:
Nice, this doesn't make the cut button look awkward. I wouldn't mind adding a few more features to the 1.5 release, as long as we do intensive testing, which we should do either way.

SjoerdB93:
There's no specific due date for the next release, so I won't block any new features if good ideas pop up. Adding new features is just not on the list of must-haves either

SjoerdB93:
This should be good to go now. It also fixes a bug that got introduced yesterday where it ran into issues when opening two datafiles at ones. At one point it tried to hide the select_data button, which no longer exists. But it also called set_mode from plotting_tools, but set_mode lives in the main function now, which was the reason multiple data sets couldn't be imported at once.

SjoerdB93:
Regarding the very last commits in this branch. I noticed that if you center part of the x-data, it didn't sort it the data points so that the x-coordinates are continiously increasing upon centering. It does so for all other actions where the x-coordinates are affected, so for consistency's sake I changed this behaviour for centering feature as well.

SjoerdB93:
I did quite a bit of testing, and while the previous commit (before the last one) did work quite well, there were some minor kinks. Mainly related to the undo/redo buttons, but also when data sets where combined with identical x-data coordinates, it threw out one of those points. Or in other words, it was not possible to combine data sets if the x-coordinates were identical. (Which is super niche and probably not something you want to do anyway, but still a bug nonetheless)

 This is fixed now in the latest commit.

Issue #80, opened 2023-02-02
The shortcuts for the different modes conflict with translate/multiply entries


SjoerdB93:
When trying to enter a number in the entry for "translate" or "multiply" X or Y, it selects a different mode instead. I suggest changing the shortkeys to shift+1, shift +2 etc. It's a quick fix, I can do it tomorrow myself as well, but I thought I'd file an issue quickly for now.

PR #82, opened 2023-02-02
change keyboard shortcuts

cmkohnen:
fixes https://github.com/SjoerdB93/Graphs/issues/80. Use Function keys, because shift + Number doesn't seem to work


SjoerdB93:
My guess is that it translates shift plus number to the alternative key behind it (e.g. exclamation mark on qwerty keyboard for shift + 1).

Binding to those keys doesn't work since the alternative key depends on keyboard layout.

SjoerdB93 (approved):

PR #83, opened 2023-02-02
Add CI workflow

cmkohnen:
Basic CI workflow, that simply builds Graphs into a flatpak ref, could be expanded for tools like linting and security analysis in the future.
Triggers on commits to main and pull requests


SjoerdB93 (approved):
Seems good to at least have a basic set-up rubbing

cmkohnen:
welp, seems the current manifest is broken

PR #84, opened 2023-02-02
fix manifest

cmkohnen:


SjoerdB93 (approved):
I could be wrong, but I think using the dir directly like this if the default on Gnome builder. The previous version with git was an adaptation of the manifest on Flathub

cmkohnen:
Just tested and Builder uses git in conjunction with a file:// uri for the project, which is weird

Issue #35, opened 2023-01-28
Make dark mode plot background default adwaita dark background color


mossymossmoss:
It would look much nicer to have the background of the plot the same as the background of the window by default
![image](https://user-images.githubusercontent.com/56167669/215284085-a922bafd-7974-46b3-8aa0-e86d4555bb83.png)

SjoerdB93:
The current background chosen is from the Matplotlib styles that are available by default in the matplotlib module, which drives all the graphing functionality in Graphs. It is possible to define own styles, so it's certainly possible to make it blend in with the background colour.

Having a purely white background is the general standard for presentations and publishing, and purely black would be the logical anologue for dark mode. I also do personally prefer a purely white background in light mode in looks, it blends in nicely with the white colours of the rows on the side-panel.

Having said that, I agree that the purely black looks a bit off in dark mode. Perhaps, to keep consistency with the light mode, it would make sense to use the same colours as the action rows again. Then it would look kinda like this:

![image](https://user-images.githubusercontent.com/68477016/215289310-6cb834f5-2fd5-44d3-9919-dab5e33b0ee5.png)

Purely black would then be offered as a seperate stylesheet.

SjoerdB93:
For reference, here's with the black background, and again with the same background as the rest of the application. These are quick mock-ups I made in GIMP, but it's pretty achievable to implement:

![image](https://user-images.githubusercontent.com/68477016/215289343-019c67ec-10a7-4115-983c-c1a453421f6f.png)

![image](https://user-images.githubusercontent.com/68477016/215289353-7ee7d604-b6d2-489c-977e-5a7c8f6de3f4.png)

cmkohnen:
To suggest something further:
Have styles, that fit GNOME as default and look visually appealing and when saving the data use a style that is suitable for publishing. This should allow for using the application in Dark Mode but exporting to a white style sheet.
This could be achieved either via popup window, or a preferences option.

SjoerdB93:
Yes, I think that may be quite a good solution. The save button is by default coupled to the "Save" command that is build into the navigationbar, so it's a bit of a black box (I can make a custom function as well, but this one works quite well). This command uses the rcparams in matplotlib for the saving parameters, the ones related to saving figures are as follows:

```
## background white
#savefig.dpi:       figure      # figure dots per inch or 'figure'
#savefig.facecolor: auto        # figure face color when saving
#savefig.edgecolor: auto        # figure edge color when saving
[...]

PR #88, opened 2023-02-03
add custom stylesheets

cmkohnen:
fixes https://github.com/SjoerdB93/Graphs/issues/35
My suggestions on how the stylesheets should look.
They are currently installed in /app/lib/python3.10/site-packages/matplotlib/mpl-data/stylelib as this is where all matplotlib styles are stored and checked for. As we are only targeting flatpak this shouldn't affect any other application, and even if so, it shouldn't disrupt anything.
I also updated default config settings.
Screenshots:
grafik
grafik
Feedback appreciated


SjoerdB93:
This looks nice, and especially the dark mode looks so much nicer. The only thing I'd like to add is that these styles fix the used font size for some reason. It still respects my choice of font, but the font size seems fixed. If that can be fixed this would be good to go, as the style sheet themselves are quite nice. I think the original light mode was fine as well, but this one has some added clarity and I think is good as well. But again, dark mode is a huge improvement. Also it's a big improvement to have grids on both of them (or on neither of them, at least have it consistent).

data/styles/adwaita.mplstyle
```diff
@@ -0,0 +1,61 @@
+# adwaita
+patch.facecolor: FAFAFA
+patch.linewidth: .3
+
+axes.axisbelow: True
+axes.edgecolor: black
+axes.facecolor: white
+axes.grid: True
+axes.labelcolor: black
+axes.labelsize: 11
```
SjoerdB93, line 10: I think this maybe the line that overrides the chosen font size. If we want to use 11 as default font size, we can set that in the config file.

data/styles/adwaita.mplstyle
```diff
@@ -0,0 +1,61 @@
...
+axes.linewidth: .8
+axes.prop_cycle: cycler('color', ['4C72B0', '55A868', 'C44E52', '8172B2', 'CCB974', '64B5CD'])
+axes.titlesize: 14
+
+font.family: sans-serif
+font.sans-serif: Cantarell
+font.weight: normal
+font.size: 12
```
SjoerdB93, line 18: Or these... 
Basically the fonts family, sans-serif, weight and size properties can all be set in the config file. For some reason the stylesheet seem to override at least the font sizes that are chosen. Making it impossible to change the font size.

SjoerdB93 (changes requested):

cmkohnen:
The problem was, that the axes.fontsize option was never set from config, it is fixed now.

SjoerdB93 (approved):

PR #94, opened 2023-02-04
code cleanup

cmkohnen:
- Remove pip-mode.ui
- update build configuration for localization (localization still needs to be adressed properly)
- move uuid creation of data to the data class
- cleanup imports
- add license identifier


src/data.py
```diff
@@ -1,9 +1,11 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+import uuid
+
 from dataclasses import dataclass, field
 
 @dataclass
 class Data:
-    id: str = ""
+    id: str = str(uuid.uuid4())
```
SjoerdB93, line 8: I like this, it's much more elegant than creating the uuid seperately.

SjoerdB93 (approved):

Issue #54, opened 2023-01-30
Simplify equation list


bertob:
There are a few oddities with the equation list as-is, namely:
- [x] The selection mode and non-selection mode states are very similar (basically just replacing a check with an interactive check, and the color with the delete button), making the two modes feel unnecessary. I'd move the interactive check to the left and always have it visible, and have color chooser and delete always visible on the right.
- [ ] The list has card styling, but doesn't behave like a real card (e.g. it doesn't grow with the content). I'd drop the card styling and add a clear separator between the two scrolling areas or make the list height dynamic and make the entire sidebar a single scrollview.
- [ ] The color dot is too large, I'd make it about 20px

![image](https://user-images.githubusercontent.com/1908896/215569460-a0eb4c4c-e64a-46b6-a68f-b62c7e9f899e.png)

![image](https://user-images.githubusercontent.com/1908896/215568820-49cb7416-e5d0-48ff-92a4-515c5bb1a969.png)

bertob:
Of the three points mentioned in the issue only the first was fixed by #49, so I'd keep this open or open separate issues for the other two things.

bertob:
> I'm personally inclined to just make the entire sidebar a single scrollview, simply because I like the look of the cards. But it's something we should experiment with anyway.

+1 for a single scrollview, that's probably the simplest solution here.

SjoerdB93:
> Of the three points mentioned in the issue only the first was fixed by #49, so I'd keep this open or open separate issues for the other two things.

Agreed, this was an oversight in merging the PR and opting for fixing the open issues in seperate PRs. This was still on the to-do list.

cmkohnen:
Sorry for responding not at an earlier time.
I would keep the sidebar as two scroll wheels, or at least keep the option to enable it that way.
The reason being consistency. If handling multiple lines for the graph, thus making the list expand, would move the quick navigation buttons, and one would always have to search a (maybe very) short moment to find them. Quickly selecting lines for the different operations to perform would also prove difficult, as one would have to scroll up and down more often.
I would suggest, letting the user resize the item list by themselves, so if they wish to expand it they have the power to do sow and do it exactly how large they want.

SjoerdB93:
I believe the issue raised is mainly that a static size is not how these cards generally behave and is therefore not consistent with typical GNOME design. So if we use this current styling, the cards should dynamically expand based on the amount of open data sets. Having the cards resizable by the user doesn't really change that.

[...]

PR #95, opened 2023-02-04
grow list view dynamically

cmkohnen:
Like suggested in https://github.com/SjoerdB93/Graphs/issues/54, the list box now grows dynamically with its content. When no content is present it remains hidden.
grafik
grafik

### DIFF
--- a/data/ui/window.ui
+++ b/data/ui/window.ui
@@ -92,7 +92,7 @@
             <property name="fold-duration">1</property>
             <property name="fold-policy">ADW_FLAP_FOLD_POLICY_NEVER</property>
             <child type="flap">
-              <object class="GtkBox" id="selection_box">"
+              <object class="GtkBox">
                 <property name="orientation">vertical</property>
                 <property name="width-request">370</property>
                 <property name="hexpand">false</property>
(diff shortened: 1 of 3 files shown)


SjoerdB93:
This works quite well, and I do like the look of the card itself without the whitespace in the bottom. I wonder if it isn't a bit empty when there's no/few data sets loaded. I'm also not 100% sure if we need a GtkSeperator on both sides. 

I tried to play with it a bit. Here's what it looks like with only one GtkSeperator and with little less space for the data list. (200 pixels instead of 250, leaving room for only three data sets instead of four):

image

I'm not 100% sure which looks best by the way. But I was curious about opinions on this.

cmkohnen:
To adress the feeling of emptiness: Perhaps we should add a label, when no data is loaded, simply telling the user to do so. This would go a long way in general UX as well.

Regarding the separators, I thought It'd be consistent if they were on the end of both scrollviews, but we can perfectly get away with either having just the top one or no one at all.

SjoerdB93:
> To adress the feeling of emptiness: Perhaps we should add a label, when no data is loaded, simply telling the user to do so. This would go a long way in general UX as well.

Yes, I was actually thinking of something along the same lines earlier today, that it probably should say something suggesting to add/open new data in this blank stage indeed.

Regarding the separators, I didn't really think about the fact that the bottom view is also a scroll view. I can see the point of being consistent and keeping them on both sides, as it does separate this mode toolbar from the action scrollview. So I think it's okay to keep them as is actually. As it are distinct areas that behave independently (the toolbar doesn't scroll, the action rows do).

SjoerdB93 (approved):
I think we can take such a message in a separate PR perhaps, as it is a separate issue that probably would be good regardless of what we do with this list. So am approving the merge now

PR #97, opened 2023-02-05
use button with icon for colorchoosing

cmkohnen:


SjoerdB93:
Just a reference, this is what it looks like with a smaller dot instead. It's a bit hacky as I just replaced the fill image with a black symbolic dot.

image

SjoerdB93:
And here's the buckets:

image

cmkohnen:
See https://github.com/SjoerdB93/Graphs/pull/98 for further development

Issue #34, opened 2023-01-28
Remove select items menu


mossymossmoss:
The select items menu is fairly unintuitive, the options to unfocus or delete a graph could be placed alongside the change colour option. It may also be easier to understand if you make the change colour option a rectangle, as is often the case with other GNOME apps, as it is pictured here in a screenshot from Paper
![image](https://user-images.githubusercontent.com/56167669/215284011-3da0657e-344e-47e5-bc78-e2d6e4564a08.png)

SjoerdB93:
It was actually suggested in the past to make the colour chooser circular, as it was a bit overbearing in a previous iteration. See attached picture for reference. 

![image](https://user-images.githubusercontent.com/68477016/215289580-9bb1953f-a91a-4b41-b55a-ec0d5bc5c465.png)

My intention was to make it so that if you double click on a data set, it would pop up the plot settings for this specific data set. Making it a lot easier to change things like the label, line styles, markers etc.. This worked well enough, but there's a nasty bug where the css of the item would get stuck in "selected" grayed mode for some reason so I didn't push that to production just yet. To have this behaviour, a separate selection mode would be needed.

In principle there's three actions that can be performed on an item. Change the colour, delete the item, and activate/deactive it. Having the selection and deletion behind a "selection" mode seemed like a logical way to achieve this while reducing the clutter. This is also consistent with design such as used in GNOME contacts:

[Screencast from 2023-01-28 21-39-38.webm](https://user-images.githubusercontent.com/68477016/215290180-d4c71828-e901-4402-87e7-c15cdae347b3.webm)

If you have specific improvements in mind to make the behaviour more intuitive I'm open to suggestions. But it was a concious decision to not add three different buttons to a single row. And the old behaviour of clicking anywhere on the row to select/deselect the dataset would conflict with the plan of opening the settings by double clicking on it.

mossymossmoss:
The new setup is definitely an improvement, much more intuitive. I would like to still propose the idea of having the color selector be rectangular, as that would be in line with similar features in other apps and would further differentiate it from the graph selector

SjoerdB93:
This is closely related to issue #54. The current GtkColorChooserWidget we use will be deprecated from Gtk 4.10 and on (which ships with GNOME 44), so we are gonna have to get rid of it. The alternative (GtkColorButton) is also deprecated, and they'll be replaced with GtkColorDialog from Gtk 4.10 and onwards (so not available yet).

[...]

PR #98, opened 2023-02-05
Change to 20 px color button

SjoerdB93:
With a color chooser the size of 20 px as disccused in issue #54. Also contains a fix to how the next color in the cycle is chosen. Behaviour is a same, but it fixes a bug where it would choose existing color anyway whenever the colorchooser widget has been activated for that particular sample. 

20 px:
image

Here's the color picker at two larger sizes as well, just for reference:

25 px:
image

30 px:
image


cmkohnen:
Uhm, you seem to have included all _build files and none of the actual source

SjoerdB93:
> Uhm, you seem to have included all _build files and none of the actual source

Woops, that's sloppy. Should be fixed now.

cmkohnen:
I quite like the 25px version

SjoerdB93:
Yeah, I think most are an improvement. For what it's worth, your PR #97 was not in vain, as this one is a derivative of that branch. I couldn't figure out how to resize a GtkButton to smaller sizes, but using an icon as a work-around (which is what this does) worked perfectly.

Just as a quick references, here's some of the options:
A = 20 px
B = 25 px
C = Fill bucket
D = 30 px
E = GNOME Color Picker
bild

I kinda like the circular dot, especially version A or B. Either are an improvement of the big dot in the current main branch.
But I do see an argument in using E as well, as @mossymossmoss mentions in issue #34 this is the standard GNOME widget for the Color Picker.

SjoerdB93:
So I found the following passage in the GNOME HIG:

>GTK and Adwaita can be used as the basis for custom UI elements, which are not included in the GNOME platform. The appearance of preexisting widgets can also be customized.
>
>As a general rule, it is better to keep the amount of this custom styling to a minimum, as this will reduce maintenance overhead, >will reduce bugs, and will help to ensure compatibility with accessibility and internationalization features.

So I'm tempted to merge this with a pixel size of 20 px. But I'll wait for @cmkohnen before force-merging this :)

cmkohnen:
I think I'm happy with either 20px (A) or 25px (B), with a slight tendency towards 20px as well, so that it fits with the close button right next to it.

cmkohnen (approved):

Issue #99, opened 2023-02-05
Graphs is missing a "Save Project" function


SjoerdB93:
To truly make Graphs useful as an application where you can also edit your graphs, and edit it at a later date. It would be extremely useful if we could save our session and load this at a later time. 

I haven't had a lot of free time this weekend, so I haven't had the time to implement this as I was hoping. But I've played around with it for the past 30 minutes, and this is really easy to implement with the `pickle` module, which can just save entire dictionaries with objects in a simple file. The idea would be to save all `plot_settings` as well as `self.datadict` and load that when one chooses to load a project. This would then load all the relevant settings from that session, including labels, axes scaling, fonts, linewidths, linestyles, colours etc... 

As a quick proof of concept, I stored such a file (containing only `self.datadict` for now) and coupled the `open` function to "Open Preferences". This works quite well, see video.
[Screencast from 2023-02-05 21-22-25.webm](https://user-images.githubusercontent.com/68477016/216843105-b4a90e51-1ef8-4711-8fe6-2a33f42172f7.webm)

I'll create a full PR when I've got the rudimentary features implemented (saving a project, and opening one using dedicated buttons in the primary menu and the "add data" menu respectively), which will be somewhere in the first half of next week. But I think it would be a really nice addition before releasing version 1.5.

cmkohnen:
Agreed, this would be nice functionality, we should however make sure, that the format we use won't brake on a future update.

PR #100, opened 2023-02-06
Add "Save as project" functionality

SjoerdB93:
This adds the "Save Project" functionality as discussed in issue #99. See video at the bottom.

Feedback welcome. I think we may need to add a separator in the menu, but I'm not 100% sure. Also, I feel like the `graphs.py ` file is maybe becoming a bit bloated. Maybe some functions should be moved to a different new file to have a more clear categorisation. But I haven't looked into that at all, perhaps all functions do fit the same category (something like "core functionality" should be there I guess). 

Also as of now, it saves the project without any extension. In principle file extensions are kinda meaningless and mostly just guidelines, but it could be worthwhile to give it some custom extension, (e.g. .graphs) so that it's easy to find and filter them from the "open file" dialog when opening a project.

The implementation of the feature is relatively simple. Basically when saving it, it saves everything as a single dictionary containing two keys. project["data"] contains all data itself (which is a dictionary), and project["plot_settings"] contains the plot settings (which is an object in the PlotSettings class) that are used to set the plot settings from that session.

I did add a `color` attribute to the data objects, as this information was otherwise not stored in the project file meaning that information was lost when saving/loading projects. Now when loading data from this project file, it uses this attribute to give the color picker the appropriate colour. When plotting the data, it then takes the color of the color picker as a reference for which colour to use (just like it was before, so that part hasn't changed).

[Screencast from 2023-02-06 10-08-25.webm](https://user-images.githubusercontent.com/68477016/216932445-235cbc44-9363-4921-8afd-5d6218856330.webm)

### DIFF
--- a/data/icons/color-picker-symbolic.svg
+++ b/data/icons/color-picker-symbolic.svg
@@ -1,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   height="20"
+   viewBox="0 0 20 20"
+   width="20"
+   version="1.1"
+   id="svg4"
+   sodipodi:docname="fill-tool-symbolic.svg"
+   inkscape:version="1.2.2 (b0a8486541, 2022-12-01)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+  <defs
+     id="defs8" />
+  <sodipodi:namedview
+     id="namedview6"
+     pagecolor="#ffffff"
+     bordercolor="#000000"
+     borderopacity="0.25"
+     inkscape:showpageshadow="2"
+     inkscape:pageopacity="0.0"
+     inkscape:pagecheckerboard="0"
+     inkscape:deskcolor="#d1d1d1"
+     showgrid="false"
+     inkscape:zoom="6.9291229"
+     inkscape:cx="-24.822766"
+     inkscape:cy="41.275065"
+     inkscape:window-width="1920"
+     inkscape:window-height="1011"
+     inkscape:window-x="0"
+     inkscape:window-y="0"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="svg4"
+     showguides="true">
+    <sodipodi:guide
+       position="-9.9142462,18.583679"
+       orientation="1,0"
+       id="guide2624"
+       inkscape:locked="false" />
+  </sodipodi:namedview>
+  <circle
+     style="fill:#000000;fill-opacity:1;stroke:none;stroke-width:9.78662;stroke-linecap:square;paint-order:stroke markers fill"
+     id="path2630"
+     cx="10"
+     cy="10"
+     r="10" />
+</svg>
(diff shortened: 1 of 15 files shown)


SjoerdB93:
Regarding the markers dissapearing when choosing tri_down, apparently that's a bug. The way I implemented this marker list, is just that I ask Matplotlib for all options and populate the chooser with those options. (Same is done with the graph style sheets, hence the weird naming convention there)

However the tri_direction series seem broken, and it can't set it to those values (hence it going back to "point") for some reason. From the name they seem identical to triangle_direction though. Maybe it's best if we set these values by hand instead of dynamically filling this list based on the matplotlib options.

SjoerdB93:
I think at the least we should rename "Save project" option to "Export project", again to remain consistent with the naming of the other options.

cmkohnen:
Looks nice, will take a look when I'm on my PC

cmkohnen:
This seems to work rather nicely. I do not know much about the pickle module, so this might be a redundant question:
Does this scale with adding new options? I think we kind of need a system that checks, if a queried value is actually present in the format and load default if necessary. E. g. adding a plot setting which is queried on load might result in bugs, if trying to load a project from a previous version. on that note we might want to store the version, the project was generated in too.

SjoerdB93:
So basically what the module does is save and load entire Python objects into a regular file. So in this case, I put both `self.plot_settings` and `self.datadict` into a dictionary, which I save using the Pickle module. Then when loading it, I get this stored dictionary with my objects, and can access them using the key I gave to it. Adding the current version number of Graphs is probably a good idea, if only for troubleshooting in the future. This is also quite trivial by simply adding `version` into this dictionary.

Regarding issues in the future. It simply loads `self.datadict` and `self.plot_settings` as-is from the previous session. So if new attributes are added to either the data set or plot_settings, we will probably run into an AttributeError. So I agree that there should be a system in place that fixes that. Perhaps something similar to what we do with the config file in preferences, that just goes through the attributes in a template, and add it if it's missing from our actual object.

cmkohnen:
Yeah, I think having the same system in place like checking for config is a good idea. One could use pythons [inspect](https://docs.python.org/3/library/inspect.html#inspect.getmembers) method.

SjoerdB93:
Yes, it's good to have something in place to keep compatibility over the longer term.

I've changed the `data` object a bit, and moved some code to this class. Basically, the initial clipboard data and position are now generated in the init function. And `set_data_properties` is now a class specific function that is applied on the dataset. Import_settings is an optional argument that can be given. A new dataset can then be created using data_set = Data(self, xdata, ydata, import_settings). Where the import_settings is optional, if not given it will just take the default import_settings from the preferences.
This all reduces quite a bit of replication, as you no longer explictly have to create new clipboard data all the time, this is just done on init as it should be.

I was thinking of using attributes that are present in the data objects to generate new Data objects when importing the data. This way, we explicitly use the same template/Class as for regularly newly added files. This doesn't stop the necessity of such a key synchronization as we have in preferences I think (definitely not for import_settings), but it should be a bit more robust. Will take a look at an implementation later.

SjoerdB93:
This should work, but I should test it a bit more thoroughly probably. 
Also, I may want to add some separator in the menu...

SjoerdB93:
I also changed the variable name id to key as id conflicts with a built-in Python function.
Also removed the dataclass decorator in Data as it lead to some troubles, basically the attributes that aren't defined during init are never assigned and therefore cannot be copied over. Figured it's probably best to define this during init instead.

cmkohnen (approved):

PR #106, opened 2023-02-07
Add notice, when no data is loaded

cmkohnen:
image
I'm not 100% certain with the wording / line break and definitely open for suggestions

### DIFF
--- a/data/ui/window.ui
+++ b/data/ui/window.ui
@@ -122,6 +122,31 @@
                                 </style>
                               </object>
                             </child>
+                            <child>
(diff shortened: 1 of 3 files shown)


SjoerdB93:
This is great, and seems to work just as it should!
Regarding the wording, I was thinking perhaps "Press the button in the top-left corner to add data", but I'm not hard opposed to the current wording either.

SjoerdB93 (approved):

cmkohnen:
grafik
Changed the wording and added the note about files and equations for the layout's sake

Issue #108, opened 2023-02-08
Improve "Add Equation" dialog


bertob:
The layout of the "add equation" dialog could be better, a list of grouped AdwEntries isn't ideal here, especially when some of them don't need to be quite so large.

![image](https://user-images.githubusercontent.com/1908896/217405468-9de02fe9-afa8-45d7-b6c3-b724dee14e69.png)

Here's a quick sketch for what a more appropriate layout could look like:

![image](https://user-images.githubusercontent.com/1908896/217405633-94346070-c4e4-4182-a59c-d5393ee83691.png)

SjoerdB93:
Thanks a lot for the feedback. 
I made a quick variant with this feedback in mind, and I do have to say it's already quite an improvement. See screenshot:
![image](https://user-images.githubusercontent.com/68477016/217449625-dff76b07-a679-405d-883f-ddbb2487484d.png)

I've also tried to set the `halign` to `center` as in the sketch, but it looked quite off. Maybe decreasing the margins there will improve ti but I was a bit unsure.
![image](https://user-images.githubusercontent.com/68477016/217449755-1b6f657f-57bd-46b5-9916-26c07e8cb89c.png)

I may also want to decrease the default size, then it looks like this (with the centered button):
![image](https://user-images.githubusercontent.com/68477016/217449873-12cd291c-ea87-4f43-99dc-62ea6d2ab8f8.png)

I've uploaded this as a separate branch and a draft PR.

bertob:
Looks pretty good, a few notes:
- I'd drop the second "Add Equation" title above the formula
- It'd look better with a flat header bar, like e.g. AdwAboutWindow
- The button at the bottom should have the `.pill` class and maybe have a slightly wider minimum width, but I wouldn't go full-width
- I'd double the spacing between the name and the 3 entries below
- The spacing between the 3 entries at the bottom should be the same in both gaps (12px)

SjoerdB93:
Thanks, I'll take a look and try apply the points in this list later today.  Also agree about the flat header bar

On a similar note, I'll probably change the "Transform Data" dialog and the "Add data (advanced)" dialog a bit as well to match the updated design language in "Add equation" a bit more.

SjoerdB93:
That's already an improvement

![image](https://user-images.githubusercontent.com/68477016/217625137-ce99cc69-9d8a-433c-9e31-f8e29f1f8b12.png)

[...]

PR #111, opened 2023-02-08
Improved modal dialogs

SjoerdB93:
Improves the "Add equation" window as discussed in issue #108 

image

Feedback welcome, I'm particularly not really sure yet about the margins to use everywhere.


cmkohnen:
This doesn't run on my machine:
 ` File "/app/share/graphs/graphs/preferences.py", line 35, in check_config
   config = utilities.remove_unused_config_keys(config, template)
TypeError: remove_unused_config_keys() missing 1 required positional argument: 'template'`

SjoerdB93:
Hmm, interesting. Seemed to work here. Can't find a remove_unused_config_keys without positional argument being given.

I've merged the current main branch into this PR again, maybe something got out of wack the previous time I did this. Try again with the latest commit maybe.

cmkohnen:
Yeah I've found the reason, this is a thing in all other open PRs

SjoerdB93:
Changed the above two dialog's confirm button to a pill as well to match the new style of "add equation"

image
image

SjoerdB93:
Changed the margins a bit, also in the other modal dialogs. 
Removed the label as well for the other dialogs, to be consistent with "add equation". Feedback is welcome either way :)

image

image

SjoerdB93:
I now put the information with more information about the Transform Window in an ExpanderRow, however the window doesn't resize back when I collapse it again. Not sure if this is a behaviour that's wanted. See video either way.

[Screencast from 2023-02-09 08-59-21.webm](https://user-images.githubusercontent.com/68477016/217752010-1f8df920-5940-4ba0-a147-f01619170ef8.webm)

cmkohnen (approved):

cmkohnen:
I'd suggest just leaving the text for the transform window always visible. It is supposed to be for advanced uses anyway and doesn't really hinder the experience in my opinion. Having it under the two entry rows is quite a good change. Another solution would be to add a "?" button into the headerbar, which then opens either another dialogue or a AdwLeaflet showing this text

SjoerdB93:
Yeah I think I'll move it back to always visible for now. Longer term it needs to be revamped a bit with some more proper documentation. There I'd like to have some specific examples, and perhaps activate scipy tools as well.

Right now it imports numpy as * for that specific module. Which is why I had this module as a seperate py file, as I don't want to do blank * imports in the rest of the code for obvious reasons. However, I didn't want users to have to write np.sin(X) to do a sine, hence the * import for numpy. I could add scipy to that mix with a single line of code, but it wouldn't surprise me if some functions overlap. So for scipy I was thinking to have it import scipy as sp, and explain that in documentation at one point. 

But again, that's future ideas that are beyond the scope of this PR, which is just about the UI design. For now having it as always visible would be fine I think (or at least better than having the window expand, and not go back to the original size as in the video). And then when we expand the documentation of it, we can tuck it away in a different widget. (We could just add a link, but I'm not always a fan of locking documentation away behind internet access)

I like the idea of putting it in an AdwLeaflet or something similar. Not sure about a new dialog. I've thought about that as well, but I didn't want a cascade of several dialogs popping up. Even though it is kinda pragmatic, it doesn't feel super nice design-wise.

SjoerdB93:
Current version

image

Issue #89, opened 2023-02-03
No white surfaces (apart from external media etc)


cmkohnen:
Resizing the graph currently shows a white background even in dark mode, this should be changed

SjoerdB93:
I think the issue is that Matplotlibs FigureCanvas tries to redraw immediately each time the window is resized. I've gotten the entire widget to hide when it's being resized by connecting a hide() function to the `resize`  signal of the widget. Unfortunately I don't know how to get it back again.

Best would probably to get it to stop redrawing completely while the window is being resized, and enable this behaviour again when we stop resizing but I'm not sure yet if this is possible or how. There's some hacky ideas I'll may try tomorrow but not sure if I'll find a great solution.

SjoerdB93:
So I think this is an upstream Matplotlib issue, here's a minimal reproducible example using purely matplotlib:

```
import numpy as np
import matplotlib.pyplot as plt


plt.style.use('dark_background')

fig, ax = plt.subplots()

L = 6
x = np.linspace(0, L)
ncolors = len(plt.rcParams['axes.prop_cycle'])
shift = np.linspace(0, L, ncolors, endpoint=False)
for s in shift:
    ax.plot(x, np.sin(x + s), 'o-')
ax.set_xlabel('x-axis')
ax.set_ylabel('y-axis')
ax.set_title("'dark_background' style sheet")

plt.show()
```

See attached screencast as well. I'll try and ask around in Python/matplotlib circles if it's possible to stop Matplotlib from redrawing while the window is being resized, but I'm not sure about this. This could be that it's some GTK thing to redraw widgets upon resizing, which happens to be matplotlib in this case.

I guess the workaround again would be to hide the widget when resizing and showing it again while not resizing. Perhaps it's possible the change this white background we get to black (which I guess would be a matplotlib thing again), but I would not really know how. I will keep this issue open, and ask around to see if there's something that can be done.

SjoerdB93:
Indentation didn't really work on pasting it on the terminal, so the lines where not shown, here's a screencast where I just open a script with the same example code:

[Screencast from 2023-02-08 09-28-32.webm](https://user-images.githubusercontent.com/68477016/217475638-69a360c1-d868-4c1c-aa59-b4158b1b2c0a.webm)

SjoerdB93:
Here's a screencast of the general problem described in this issue, just for reference:

[...]

PR #113, opened 2023-02-08
Switch to cairo backend for figurecanvas

SjoerdB93:
Fixes issue #89. Seems this is an upstream issue in the regular gtk4 backend, but it works fine for cairo.
No extra imports or packages are needed, Cairo needs requires [PyGObject](https://wiki.gnome.org/action/show/Projects/PyGObject) and [pycairo](https://www.cairographics.org/pycairo/) but it seems both are included in the standard runtimes (PyGObject obviously is, but I wasn't sure about pycairo before).
[Screencast from 2023-02-08 11-59-30.webm](https://user-images.githubusercontent.com/68477016/217512278-5cd9ef9e-8d3d-4b7b-8dc2-0d47ccfaaf17.webm)

Closes issue #89

### DIFF
--- a/src/plotting_tools.py
+++ b/src/plotting_tools.py
@@ -7,7 +7,7 @@ import matplotlib.pyplot as plt
 
 from gi.repository import Gtk, Adw
 from matplotlib import colors
-from matplotlib.backends.backend_gtk4agg import FigureCanvasGTK4Agg as FigureCanvas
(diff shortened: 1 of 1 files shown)


cmkohnen:
Hmm, I genuinely think, this improved rendering performance as well. The sidebar animation feels much snappier for example.

Issue #114, opened 2023-02-08
Unique tool icons


bertob:
There are icons for some of the tools in devkit now: https://gitlab.gnome.org/Teams/Design/icon-development-kit/-/issues/71

You can export the individual icons from icons.svg by opening it in Symbolic Preview and exporting the whole sheet.

SjoerdB93:
Wow, this is great! I'm honestly amazed by the help and support I've been getting with this application. It's much appreciated! Here's what it looks like with the icon-set, for the translate and multiply icons, I've used horizontal arrows and vertical arrows (which are just the horizontal ones rotated 90 degrees), but perhaps I could go for a + sign for Translate, and a X sign for Multiply instead.

This is a huge improvement. I'm not sure if all metaphors are perfectly appropriate. Mainly:

- The "Shift" icon looks like it signifies a gear box shift or something similar, rather than a lines in a graph shifting vertically. But perhaps there's another existing icon that could be used.

- Center is also a clear metaphor that shows a dot being centered in the middle. Which is fine when refering to the abstract term of centering stuff: it puts something in the center. But I am not sure if we'd need something that specifically refers to a graph being centered. (This is more of a question than a comment) Same comment about Transform icon.

- I'm not 100% sure what metaphor that is being used for the Fourier transform. Typically I either think of the mathematical expression ([image](https://user-images.githubusercontent.com/68477016/217531501-c3cf14f5-6d0c-4b93-8285-20b0482cb7a2.png)), a transform from time to the frequency domain ([image](https://user-images.githubusercontent.com/68477016/217531307-552f6c9f-21a1-4be5-8be7-bf338fc9a203.png
)), or the transformation of real spaces to reciprocal space in the case of e.g. crystal lattices ([image]([image](https://user-images.githubusercontent.com/68477016/217531926-dedacb52-6df5-42e0-9762-1c3b0179fee6.png)). I guess it could represent e.g. a hexagonal crystal lattice in this case. Either way, people using a Fourier Transform will know what it is anyway, so I'm not pushing for specific changes, I was just not 100% sure what it represents in the icon set. I also really like the inverse version just having inverted colours by the way. 

Regarding the second point about icons not being graph-specific. I really don't know if that's an issue. Using a floppy diskette in a text editor has very little to do with saving stuff nowadays. It's quite common to have a metaphor to represent a specific issue, hence my remark that it is more of a question than a comment. Is it okay to use such a metaphor, or should we strive for icons that specifically target a graph?

[...]

PR #115, opened 2023-02-08
Add unique icons

SjoerdB93:
See issue #114, adds unique icons to each button. Open to discussion and feedback on the icons used in this PR.

image


cmkohnen (approved):
I think the icons are fitting

SjoerdB93:
Yeah, Jakub Steiner (who made these) mentioned on the GNOME Design Matrix he'd create a more appropriate shift icon. For the others we would keep them as is, as they work quite well.

PR #116, opened 2023-02-08
deduplication

SjoerdB93:


SjoerdB93:
This fixes a bug where the "none" option was not given in plot_settings. Also did some deduplication. Some functions in preferences.py where already present in utilities.py

PR #118, opened 2023-02-08
fix startup

cmkohnen:

### DIFF
--- a/src/preferences.py
+++ b/src/preferences.py
@@ -32,8 +32,8 @@ class Preferences():
         with open("config.json", 'r') as f:
             template = json.load(f)
         if set(config.keys()) != set(template.keys()):
-            config = utilities.remove_unused_config_keys(config, template)
-            config = utilities.add_new_config_keys(config, template)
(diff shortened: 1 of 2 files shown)


src/preferences.py
```diff
@@ -32,8 +32,8 @@ def check_config(self, config):
         with open("config.json", 'r') as f:
             template = json.load(f)
         if set(config.keys()) != set(template.keys()):
-            config = utilities.remove_unused_config_keys(config, template)
-            config = utilities.add_new_config_keys(config, template)
+            config = utilities.remove_unused_config_keys(self, config, template)
+            config = utilities.add_new_config_keys(self, config, template)
         return config
```
SjoerdB93, line 37: Ah, now I see the issue. Neither function actually uses `self` though, so perhaps it's better to remove self from the input arguments in `utilities.py` instead.

SjoerdB93 (changes requested):

SjoerdB93 (approved):
I now see why this error didn't get triggered for me. As my config file didn't show any differences with the template, it never triggered these functions.

PR #120, opened 2023-02-08
implement own rubberband zoom

cmkohnen:
This has a few changes to make this feature work:

- `DummyToolbar` Class based on `backend_bases` version of `NavigationToolbar`. This eliminates any unnecessary overhead from having unrendered gtk widgets, and gtk-releated stuff, that we don't use anyway. Since `backend_bases` does not implement the `draw_rubberband` function this is implemented as well (even more optimized than the upstream matplotlib implementation)
- A `_post_draw` classfunction in `PlotWidget`. This overrides matplotlibs implementation of a rubberband by hooking into existing code and just changing the way its actually rendered: using a full rectangle with a style-matching colour instead of a dashed rectangle like before.

Screenshot:
grafik

### DIFF
--- a/src/graphs.py
+++ b/src/graphs.py
@@ -6,10 +6,10 @@ import numpy
 import pickle
 
 from gi.repository import Gtk, Adw
-from matplotlib.backends.backend_gtk4 import NavigationToolbar2GTK4 as NavigationToolbar
 from . import plotting_tools, samplerow, colorpicker, utilities
 from .plotting_tools import PlotWidget
 from .data import Data
+from .utilities import DummyToolbar
 
 def get_theme_color(self):
     win = self.props.active_window
@@ -412,7 +412,7 @@ def load_empty(self):
     for axis in [self.canvas.right_axis, self.canvas.top_left_axis, self.canvas.top_right_axis]:
         axis.get_xaxis().set_visible(False)
         axis.get_yaxis().set_visible(False)    
-    self.dummy_toolbar = NavigationToolbar(self.canvas)
+    self.dummy_toolbar = DummyToolbar(self.canvas)
     win.toast_overlay.set_child(self.canvas)
 
 def disable_clipboard_buttons(self):
--- a/src/plotting_tools.py
+++ b/src/plotting_tools.py
@@ -494,7 +494,7 @@ class PlotWidget(FigureCanvas):
     """
     Create the widget that contains the graph itself
     """
-    def __init__(self, parent=None, xlabel="", ylabel="", yscale = "log", title="", scale="linear", style = "seaborn-whitegrid"):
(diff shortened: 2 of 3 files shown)


cmkohnen:
Oh and matplotlib uses cairo for the gtk4 backend in most ways anyway, which I find weird

SjoerdB93:
The code seems fine, it performs well (basically the same as the default rubberband in my test), and it looks fine.

I like that this allows us to change the styling of the zoom. Any particular reason for going for a solid line with a blue-ish colour? I quickly played around with it, and you can set a dash with a sequence. For example, if you want to alternative the line with 5 px off line and 1 px nothing you can use `context.set_dash([5,1])`, so this is very customizable:
image

Or here it is with [3,3], so equal widths:
image

I'm honestly not sure what works best. But playing around with it, a solid line feels nice as well. So I'm approving the PR.

SjoerdB93 (approved):
I think these implementations can be good in the longer run also if we ever decide to change any of the functionality, or styling.

cmkohnen:
I was mainly trying to replicate Adwaitas Rubberband styling as seen here (nautilus), but couldn't get the inner part to work (yet)
grafik

SjoerdB93:
I've got the inner part to fill, but then it doesn't show the border. I got the fill with the following code snippet, maybe that helps:
```

        context.rectangle(x0, y0, w, h)
        context.set_source_rgba(120 / 255, 174 / 255, 237 / 255, 0.2)
        context.fill()
        #input are floats so divide rgb value by 255
        context.set_source_rgba(120 / 255, 174 / 255, 237 / 255, 1)
        context.stroke()
```

SjoerdB93:
By the way, I do really like the idea of trying to replicate Adwaitas rubberband 👍

cmkohnen:
> I've got the inner part to fill, but then it doesn't show the border. I got the fill with the following code snippet, maybe that helps:
> 
> ```
> 
>         context.rectangle(x0, y0, w, h)
>         context.set_source_rgba(120 / 255, 174 / 255, 237 / 255, 0.2)
>         context.fill()
>         #input are floats so divide rgb value by 255
>         context.set_source_rgba(120 / 255, 174 / 255, 237 / 255, 1)
>         context.stroke()
> ```

Thanks somehow completely missed fill()

SjoerdB93:
We may want to check if the selection tool uses the same colour scheme and border. It would be nice to make that consistent.

At the time I just used a random blue facecolor with no border I believe. So I think we should use the same style. The new rubberband for zoom is nice, that's a good addition :)

cmkohnen:
Reference:
grafik
grafik

SjoerdB93 (approved):

SjoerdB93:
Those are looking really nice!

PR #123, opened 2023-02-09
reorder menu items

cmkohnen:
Move "Save Project" and "Open Project" to the top of the primary menu. Also update keybinds to more fitting ones. (Ctrl+S, Ctrl+O for load/save and Ctrl+E, Ctrl+Shift+E for both export tools)
Also rename some actions and functions to better reflect their name in UI


SjoerdB93 (approved):
Renames look good to me. I actually had the Save/Export Project options on top in an earlier build before comitting, but was a bit indecisive on that. I'm also happy to have those on top

Issue #109, opened 2023-02-08
"Normal mouse pointer" tool is weird


bertob:
All the other tools are a mode of interacting with the graph, but this one seems to just do nothing, so I'm not sure why it's needed.

![image](https://user-images.githubusercontent.com/1908896/217406010-14c82c14-7b4c-4e04-95a5-368c0b0428d2.png)

SjoerdB93:
While I've considered the point that this mode does literally nothing, that is kind of the idea of this mode. To have some sort of neutral mode, where mouse clicks and actions are "safe" and don't do anything. Like a way to turn off all other modes. I am open to removing it if that's considered better, as the same result can be achieved by simply not clicking on the Graph, and there's not much point to wanting to click on the Graph in Normal pointer mode either.

bertob:
> the same result can be achieved by simply not clicking on the Graph, and there's not much point to wanting to click on the Graph in Normal pointer mode either.

Other toolbox apps (e.g. GIMP, Inkscape, and the like) don't have a "do nothing" mode as far as I'm aware. It's usually select, move, or some drawing tool I think.

My feeling is that the "neutral" mode in this case could be either move or select, depending on which one you think is more frequently used. Neither of the modes is destructive, so I don't think it'd be weird for them to be on by default.

SjoerdB93:
I've been using Inkscape quite a lot today (for graphing purposes actually), and indeed all modes actually specifically do something. Even the "mouse pointer mode" is used for selection and transformation of objects. I think it's a fair point that such a neutral mode is not consistent with that kind of design at least. 

I've opened a PR to go back to three different modes.
![image](https://user-images.githubusercontent.com/68477016/217917594-39a7cd22-8b52-430b-9e1f-864bd4f5b5e2.png)

I can merge it later, but I'll test it a bit more to see if there's no unintended bugs.

PR #125, opened 2023-02-09
Remove mouse pointer mode

SjoerdB93:
Fixes issue #109

### DIFF
--- a/src/item_operations.py
+++ b/src/item_operations.py
@@ -10,9 +10,9 @@ from .data import Data
 from .utilities import InteractionMode
 from .plotting_tools import PlotWidget
 
-def export_data(widget, _, self):
+def save_data(widget, _, self):
     """
-    Open the export data dialog.
+    Open the save file dialog.
     """
     delete_selected_data(self)
     graphs.save_file_dialog(self)
--- a/src/main.py
+++ b/src/main.py
@@ -32,15 +32,15 @@ class GraphsApplication(Adw.Application):
     def connect_actions(self):
         self.create_action('quit', self.on_quit_action, ['<primary>q'])
         self.create_action('about', self.on_about_action)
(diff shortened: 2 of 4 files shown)


src/item_operations.py
```diff
@@ -10,9 +10,9 @@
 from .utilities import InteractionMode
 from .plotting_tools import PlotWidget
 
-def export_data(widget, _, self):
```
cmkohnen, line 13: Any reason this is changed back?

src/main.py
```diff
@@ -32,26 +32,25 @@ def load_preferences(self):
...
         self.create_action('undo', item_operations.undo, ['<primary>Z'], self)
         self.create_action('redo', item_operations.redo, ['<primary><shift>Z'], self)
         self.create_action('restore_view', plotting_tools.restore_view, ['<primary>R'], self)
         self.create_action('view_back', plotting_tools.view_back, ['<alt>Z'], self)
         self.create_action('view_forward', plotting_tools.view_forward, ['<alt><shift>Z'], self)
-        self.create_action('export_figure', plotting_tools.export_figure, ["<primary>E"], self)
-        self.create_action('export_data', item_operations.export_data, ['<primary><shift>E'], self)
-        self.create_action('save_project', graphs.save_project_dialog, ['<primary>S'], self)
-        self.create_action('open_project', graphs.open_file_dialog, ['<primary>O'], self, None, True)
+        self.create_action('save_data', item_operations.save_data, ['<primary>S'], self)
```
cmkohnen, line 51: same hear, this breaks currently since the action called in window.ui is still the old one

SjoerdB93: No good reasons for this, except that I was not working in a valid git directory and my branch was out of sync. Thanks for noticing.

cmkohnen (approved):

PR #126, opened 2023-02-09
Improve responsiveness

cmkohnen:
For now, this just makes the sidebar automatically retract if the window size gets too small
grafik

### DIFF
--- a/data/ui/window.blp
+++ b/data/ui/window.blp
@@ -72,41 +72,42 @@ template GraphsWindow : Adw.ApplicationWindow {
 
         Box {
           visible: bind sidebar_flap.folded;
-          spacing: 10;
-          margin-top: 12;
-          margin-bottom: 12;
-          margin-start: 12;
-          margin-end: 12;
+          margin-top: 6;
+          margin-bottom: 6;
+          margin-start: 6;
+          margin-end: 6;
           Adw.SplitButton {
             icon-name: bind add_button.icon-name;
             action-name: bind add_button.action-name;
             tooltip-text: bind add_button.tooltip-text;
             menu-model: bind add_button.menu-model;
-            hexpand: true;
+            styles ["flat"]
           }
-          Box {
-            styles ["linked"]
-            Button {
-              action-name: bind undo_button.action-name;
-              icon-name: bind undo_button.icon-name;
-              tooltip-text: bind undo_button.tooltip-text;
-              sensitive: bind undo_button.sensitive;
-              hexpand: true;
-            }
-            Button {
-              action-name: bind redo_button.action-name;
-              icon-name: bind redo_button.icon-name;
-              tooltip-text: bind redo_button.tooltip-text;
-              sensitive: bind redo_button.sensitive;
-              hexpand: true;
-            }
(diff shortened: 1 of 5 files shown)


cmkohnen:
Idea is to move headerbar buttons in here as well (when modal), to free space (so the title could display for example)

SjoerdB93:
Just the sidebar retracting by default when the window becomes too small is a nice touch, although we would need an easy way to bring it back up again on demand. I haven't checked this build as it's still using the old libraries which I had to compile again, and I don't really have that much time right now.

If we would include responsive buttons in the header bar, then maybe adding such a button in there to bring it up would be nice, like with nautilus where such a button pops up in the top left-corner when the window becomes too small. (Also in Nautilus some headerbar buttons dissapear to make space)

image

cmkohnen:
That was the eventual idea

cmkohnen:
I have a proper Idea now on how to implement this, but some of that is dependent on blueprint syntax, so will rewrite this when https://github.com/SjoerdB93/Graphs/pull/137 is merged.

cmkohnen:
Rebase on blueprint.

Current implementation works almost exclusively through xml (blueprint).
preview

grafik

SjoerdB93:
This looks and works good, I think it's nice to put these extra buttons in the sidebar in the compact mode. Perhaps we can change their alignments a bit and give them a flat style to make it more consistent with the top bar?

That makes it look like this when I played with it:
image

Using the following blueprint code:

```
        Box {
          visible: bind sidebar_flap.folded;
          spacing: 10;
          margin-top: 12;
          margin-bottom: 12;
          margin-start: 12;
          margin-end: 12;
          Adw.SplitButton {
            icon-name: bind add_button.icon-name;
            action-name: bind add_button.action-name;
            tooltip-text: bind add_button.tooltip-text;
            menu-model: bind add_button.menu-model;
            hexpand: false;
            halign: start;
            styles ["flat"]
          }
          Box {
            Button {
              action-name: bind undo_button.action-name;
              icon-name: bind undo_button.icon-name;
              tooltip-text: bind undo_button.tooltip-text;
              sensitive: bind undo_button.sensitive;
              halign: start;
              hexpand: false;
              styles ["flat"]              
            }
            Button {
              action-name: bind redo_button.action-name;
              icon-name: bind redo_button.icon-name;
              tooltip-text: bind redo_button.tooltip-text;
              sensitive: bind redo_button.sensitive;
              halign: start;
              hexpand: false;
              styles ["flat"]
            }
          }
          Adw.SplitButton {
            icon-name: bind view_menu_button.icon-name;
            action-name: bind view_menu_button.action-name;
            tooltip-text: bind view_menu_button.tooltip-text;
            menu-model: bind view_menu_button.menu-model;
            halign: end;
            hexpand: true;
            styles ["flat"]
          }
        }
```
Note I removed the linked styles also, as that conflicts with the flat style it seems (also according to the Libadwaita documentation they can't be used together).

cmkohnen:
That definitely looks better, I also reduced the margin a bit to better fit with the titlebar as well
grafik

Issue #121, opened 2023-02-08
Consider updating libraries


cmkohnen:
whilst checking matplotlib documentation, I realised, that we are using an "old" version (3.6.2 instead of 3.6.3). So I checked which libraries are up to date and found a few that had newer versions:

- matplotlib 3.6.2 -> 3.6.3
- scipy 1.9.3 -> 1.10.0
- setuptools 59.2.0 -> 67.2.0
- wheel 0.37.0 -> 0.38.4
- cython 0.29.32 -> 0.29.33
- numpy 1.23.3 -> 1.24.2
- pyparsing 3.0.7 -> 3.0.9
- pillow 9.3.0 -> 9.4.0
- packaging 21.3 -> 23.0
- contourpy 1.0.6 -> 1.0.7
- fonttools 4.31.2 -> 4.38.0

Is there any specific reason to have these versions besides just not having them updated in some time? If not, it might be worthto update these before the 1.5 release

SjoerdB93:
The Flatpak Manifest for Matplotlib, scipy and numpy are build by hand. Meaning I set  the link to the PyPy wheels and the appropriate checksums myself. When creating these manifests these were the latest versions that were out and build successfully both for x86 and aarch.

All other dependencies are generated using flatpak-builder, which typically also just retrieves the latest versions using pip if I recall correctly.

If any library is not on the latest version, it's probably simply because I haven't updated them yet, often I just check the latest versions when updating runtimes but that's not because of any specific reason, generally I like to be on the latest releases. Thanks for the heads up, I'll give it a look tomorrow.

PR #127, opened 2023-02-09
Update libraries

SjoerdB93:
Solves issue #121.

The reason most of these libraries are out of date, is because matplotlib, numpy and scipy wheels specifically demand those (older) versions when compiling. However, that still leaves us with these three big libraries that were slightly out of date.

This should fix that. I was able to build this with `flatpak-builder`, but gnome-builder was complaining to me that this new version doesn't exist in cache. Of course it doesn't, but it didn't want to build. I think it's a local issue for me with Gnome builder specifically, but I'll abstain from merging before figuring that out.

### DIFF
--- a/python3-matplotlib.json
+++ b/python3-matplotlib.json
@@ -17,36 +17,36 @@
         },
         {
             "type": "file",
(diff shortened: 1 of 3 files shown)


SjoerdB93:
I removed all cache, reinstalled builder and tried again. This works fine!

cmkohnen:
Builds just fine for me

cmkohnen (approved):

PR #131, opened 2023-02-12
refactoring

cmkohnen:
- Moves many functions out of graphs into new files.
- Moves Canvas into new file
- Import Settings is now an object
- various abstractions


src/add_data_advanced.py
```diff
@@ -9,7 +9,7 @@ def open_add_data_advanced_window(widget, _, self):
     """
     win = AddAdvancedWindow(self)
     button = win.open_advanced_confirm_button
-    win.set_transient_for(self.props.active_window)
+    win.set_transient_for(self.main_window)
```
SjoerdB93, line 12: Yes, this is a good change. In the first implimentation, I just called props_active_window seperately, but using the static self.main_window is more robust.

src/canvas.py
```diff
@@ -0,0 +1,188 @@
...
+
+from gi.repository import Gtk, Adw
+from matplotlib.figure import Figure
+from matplotlib.backends.backend_gtk4cairo import FigureCanvas
+from cycler import cycler
+
+from . import rename_label, utilities
+
+class Canvas(FigureCanvas):
+    """
```
SjoerdB93, line 13: Yes, this class has become big enough that it's probably good to seperate.

src/data.py
```diff
@@ -24,8 +25,8 @@ def __init__(self, parent, xdata, ydata, import_settings = None):
         self.xdata_clipboard = [self.xdata.copy()]
         self.ydata_clipboard = [self.ydata.copy()]
         self.key: str = str(uuid.uuid4())
-        if import_settings is None:
-            import_settings = graphs.get_import_settings(parent) 
+        if(import_settings == None):
```
SjoerdB93, line 28: Using` is None`  instead of `== None` is by purpose. The difference is that `== None` checks for value equality (do two objects have the same value. While `is None` checks for reference equality (does our object refer to `None` specifically). As a general rule, you'd like to use `is None` when refering to None.

Unless you've got a good reason to use == here, I'd suggest keep using is, as it's considered best practice.

src/data.py
```diff
@@ -39,10 +40,10 @@ def set_data_properties(self, parent, import_settings):
         self.unselected_markers = parent.preferences.config["plot_unselected_markers"]
         self.selected_marker_size = parent.preferences.config["plot_selected_marker_size"]
         self.unselected_marker_size = parent.preferences.config["plot_unselected_marker_size"]        
-        if import_settings["name"] != "" and import_settings["mode"] == "single":
-            filename = import_settings["name"]
+        if import_settings.name != "" and import_settings.mode == "single":
+            filename = import_settings.name
         else:
-            filename = import_settings["path"].split("/")[-1]
+            filename = import_settings.path.split("/")[-1]
```
SjoerdB93, line 46: Have to say, this syntax looks so much cleaner as object attributes. Was unsure about this change when I read the post, but I think I prefer this.

src/file_io.py
```diff
@@ -0,0 +1,96 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+import numpy
+import re
+import pickle
+
+from . import graphs, plotting_tools
+from .data import Data
+
+def save_project(self, path):
```
SjoerdB93, line 9: Yeah I was already considering moving this to it's own function, as it didn't have a logical place anywhere else.

src/graphs.py
```diff
@@ -1,23 +1,12 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
-import gi
 import os
-import re
```
SjoerdB93, line 4: Quite a large amount of differences in this file, but that seems to be a consequence from moving away many functions to their own new files. From what I've looked at it seems good.

src/item_operations.py
```diff
@@ -7,8 +7,7 @@
 
 from . import plotting_tools, graphs, utilities
 from .data import Data
-from .utilities import InteractionMode
```
SjoerdB93, line 10: This seems to be mostly the swich from props.active_window to main_window. I think it's more robust this way, so good change.

src/plotting_tools.py
```diff
@@ -1,30 +1,26 @@
...
 
-from . import graphs, utilities, rename_label
+from . import graphs, utilities
 
 def define_highlight(self, span=None):
     """
     Create a span selector object, to highlight part of the graph.
     If a span already exists, make it visible instead
     """
+    color = self.main_window.get_style_context().lookup_color('accent_color')[1]
```
SjoerdB93, line 17: Does this get the used accent color in the GNOME theme? That would be really nice actually, as that would ensure compatibiltiy with Gradience or perhaps Ubuntu accent colors (not sure how those work, I don't use Ubuntu). While I'm not targeting non-stock GNOME, extra compatibility is always nice. Also makes it easier when the theme slightly changes.

src/ui.py
```diff
@@ -0,0 +1,163 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+from gi.repository import GLib, Adw, Gtk
+
+from . import plotting_tools, file_io, graphs
+
+def toggle_sidebar(action, shortcut, self):
```
SjoerdB93, line 6: Nice to have these seperate.

src/utilities.py
```diff
@@ -135,23 +133,27 @@ def create_rgba(r, g, b, a=1):
     res.alpha = a
     return res
 
-class InteractionMode(Enum):
-    PAN = 1
-    ZOOM = 2
-    SELECT = 3
+def get_theme_color(self):
```
SjoerdB93, line 136: Good

SjoerdB93 (changes requested):
There's one change from `is None` to `== None`. Unless you've got good reasons for this, I'd suggest changing back to `is None`, which is considered the proper Pythonic way to do comparisons with `None`. That's the only "Request changes" part of this review :)

Otherwise, some much needed refactoring, looks good overall. Should check if it still works when merged with the latest main, but this seems. Now we've got `plot_settings` as an object, we could ask ourselves if `self.config` should be changed from a dictionary to an object as well. I could do that separately in a separate PR though. Just throwing the suggestion out there.

src/plotting_tools.py
cmkohnen: I think it does, Haven't considered compatibility. My idea was just not having it hardcoded

src/data.py
cmkohnen: Will change that. At some point I removed this, not considering it breaks stuff and just reimplemented it

SjoerdB93 (approved):

PR #132, opened 2023-02-14
fix ui

cmkohnen:
Somehow between merging the UI got changed back, this just reintroduces these changes


SjoerdB93 (approved):

SjoerdB93:
The same thing has happened to me as well, with the exact same lines. I've changed my workflow a bit, and simply installed GitHub Desktop Flatpak, to make sure I keep working in a valid git directory. 

The CLI works, but wants me to authenticate with a specific token each time which gets tedious. Back when I used GitLab, I get it set-up nicely with a SSH key so that I don't have to copy that password in each time, but haven't got around trying to fix this with GitHub.

cmkohnen:
>  Back when I used GitLab, I got it set-up nicely with a SSH key so that I don't have to copy that password in each time, but haven't got around trying to fix this with GitHub.

It is actually relatively straight forward. I'm using gh-cli which can be set as authentication agent for git. As per signing commits, I pasted my public key into GitHubs field and all signed commits have gained a verified badge.

PR #133, opened 2023-02-14
More refactoring

cmkohnen:
Some more refactoring.
Main focus are actions. Functions no longer include actions and targets (wrongly named widgets and shortcuts). Also actions are defined just using their name, keybind(s) and implementing a respective function in actions.py.

Also fixes add data advanced, which I broke when changing how import_settings work.
Some samplebox changes, to move logic inside it


SjoerdB93:
```
se.sjoerd.Graphs:                  AppData problem: tag-invalid : Expected children for tag
Error loading AppData file: AppData file /app/share/appdata/se.sjoerd.Graphs.appdata.xml was not valid
Error: ERROR: appstream-compose failed: Child process exited with code 1
Error: Build failed: Error: The process '/usr/bin/xvfb-run' failed with exit code 1
```

cmkohnen:
Okay, reverted that change

cmkohnen:
Regarding the last commit, I tried looking up colors on dark/white-mode change, but for some reason it would work when loading the app, but not when changing whilst the app is running

SjoerdB93 (approved):

PR #134, opened 2023-02-15
Update CI workflow

cmkohnen:


cmkohnen:
Huh, seems like the documentation on the docker page is quite outdated.

Issue #135, opened 2023-02-16
Consider Linting


cmkohnen:
Out of sheer curiosity, I ran both flake8 and pylint against the existing codebase. A lot of things caught my attention there. 
Some things we already discussed (e. g. using `is None` instead of `== None`) showed up and a lot of other idiomatic decisions, for which we should conform to the PEP. I'll try to improve the "quality" of the codebase in an upcoming pr, based on most suggestions.

Some of them are pretty straight-forward and easily implemented (e.g. fixing whitespaces, newlines, import orders, etc.) But for some we should, if we want to discuss what settings we want to choose (e.g. max length of a line) and what we want to actually check for.

I'd suggest using [this github action](https://github.com/marketplace/actions/run-flake8-with-reviewdog) in CI, starting with Release 1.5 to have automatic annotations directly when opening prs. We could even consider setting it to fail, if code standard might not be met.

Just for transparency Here are my runs of both linters:
https://pastebin.com/W7vD0kLa
https://pastebin.com/iUuMTrjQ

I'm open for suggestions and feedback, if we want to use linting as a whole and in the way I've proposed

SjoerdB93:
Yeah in general I think this is a good idea. The very first code base comes from me starting this as a personal project to help me with my graphs in my PhD thesis. Then it became quite useable so I put it out online, and then it developed into what this is new. 

So when first writing this, I didn't really consider too many standards, apart just "intuitive taste" so to say. 
Also, my personal background is in physics, and I'm pretty much self-taught in coding so that does leave some bad habits at certain points. As far as I've understood, scientists are kinda notorious for poor coding practices. Having said that, I do care about code quality and sustainable development, meaning this kind of workflow is definitely something we should set up.

On a personal note. I'll be finishing with my PhD this June, and got accepted for a post-doc in scientific development. With that in mind, more standardized/professionalised workflows are a thing I want to get into anyway. Also apart Graphs. So it's a good thing you bring this up.

I'm not very familiar with linting, but these runs do sound useful to get the code up to a better general standard. I've actually never considered single quotes to be preferred in Python. In fact, I always thought that for strings specifically double quotes were the preferred. Looking at PEP guidelines, it seems there's no formal recommendation. Is there any reason why window.py gets so many errors about `Double quotes found but single quotes preferred`?

cmkohnen:
> In fact, I always thought that for strings specifically double quotes were the preferred. Looking at PEP guidelines, it seems there's no formal recommendation.

[...]

PR #136, opened 2023-02-16
Linting

cmkohnen:
Implements changes based on flake8 linting.
Sets up flake8 config to use for future CI runs and keep consistency across machines.

Currently, docstring related warnings are ignored, these might be more fit for a documentation-related pr.
The main problems with pylint atm are, that it is quite slow and needs to be properly set up in an environment, where all dependencies and the graphs module are properly installed. This makes up ca. 90% of the errors pylint raises, so I would mainly focus on flake8 linting for the moment.

As of now, only the following output remains (ignoring everything related to docstrings):

image

A few notes on these:

- the numpy wildcart imports should be changed, but this is something, which has already been discussed, and I didn't want to change user-facing functionality on these ones
- the actions import on graphs is actually needed, but not recognized, I'll add these as an exception to the config
- The error in preferences I haven't gotten around to debug
- The error in utilities I haven't dared to change without fully understanding the function

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [flake8]
 count = True
 #default_ignorelist: E121,E123,E126,E226,E24,E704,W503,W504
-ignore = E501, Q000
+ignore = E501, Q000, D
 inline-quotes = double
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,17 @@
 [flake8]
 count = True
 #default_ignorelist: E121,E123,E126,E226,E24,E704,W503,W504
-ignore = E501, Q000, D
+ignore =
+	# temporarily disable double quotes
+	Q000
+	# temporarily disable line too long
+	E501
+	# temporarily disable docstring related stuff
+	D100, D101, D102, D103, D107, D204, D205, D400, D401
 inline-quotes = double
+
+per-file-ignores =
+	main.py: F401
+	# As long as numpy should be available in its current state
+	add_equation.py: F403, F401
+	transform_data.py: F403, F401
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,5 @@
+[flake8]
+count = True
+#default_ignorelist: E121,E123,E126,E226,E24,E704,W503,W504
+ignore = E501, Q000
+inline-quotes = double
--- a/tox.ini
+++ b/tox.ini
@@ -2,8 +2,6 @@
 count = True
 #default_ignorelist: E121,E123,E126,E226,E24,E704,W503,W504
 ignore = U101
-	# temporarily disable double quotes
-	Q000
 	# temporarily disable line too long
 	E501
 	# temporarily disable docstring related stuff
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [flake8]
 count = True
 #default_ignorelist: E121,E123,E126,E226,E24,E704,W503,W504
-ignore =
+ignore = U101
 	# temporarily disable double quotes
 	Q000
 	# temporarily disable line too long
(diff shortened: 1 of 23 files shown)


cmkohnen:
I'd say this would be ready for review: before that, two things (both mentioned in https://github.com/SjoerdB93/Graphs/issues/135) on which I#d like feedback:

- What type of quotes do we want to use? I'm slightly in favor of double quotes, but do want feedback
- What maximum file-length should we use? I think the default is a little bit short and would suggest 100 chars

One other thing:
Docstring related checks are turned off atm, these are better fit for another pr.

cmkohnen:
> > * What maximum file-length should we use? I think the default is a little bit short and would suggest 100 chars
> 
> This is just for the file name itself right? Does this include the path? I might be confusing this with another parameter, but isn't the default 79? I personally like to be verbose, but I think that would be generous enough actually. Also seems to be recommended to use the defaults.

Oh, my bad. I meant the max character length of a single line

> 
> > Docstring related checks are turned off atm, these are better fit for another pr.
> 
> I'm not sure what it exactly checks for docstrings when turned on. Either way, as a general rule I think it's good to strive for some docstring on every function. At one point if we want to further "document" the code like this, type hinting could be a good idea as well. Although that's all not the highest priority for the coming release itself.

There are a few checks. E. g. warning when a module, class, function, etc. doesn't hava one. Also things like one-line docstrings using three lines and the general checks on ending with correct punctuation, starting capitalized and following given criteria, which I haven't looked that deep into myself

SjoerdB93:
> Oh, my bad. I meant the max character length of a single line

Ah right, that makes sense. Personally, I prefer to stay with vanilla PEP guidelines, which state 79 characters.
Although PEP8 does specify that exceptions can be made up to 99 lines:

> Limit all lines to a maximum of 79 characters.
> 
> For flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters.
> 
> Limiting the required editor window width makes it possible to have several files open side by side, and works well when using code review tools that present the two versions in adjacent columns.
> 
> The default wrapping in most tools disrupts the visual structure of the code, making it more difficult to understand. The limits are chosen to avoid wrapping in editors with the window width set to 80, even if the tool places a marker glyph in the final column when wrapping lines. Some web based tools may not offer dynamic line wrapping at all.
> 
>  Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the line length limit up to 99 characters, provided that comments and docstrings are still wrapped at 72 characters.

However, I kinda like to avoid deviating from the standard recommendations. Using 79 lines is the standard for Python libraries giving consistency with other code bases. It also makes side-by-side view (both from an IDE or side-by-side) diffs readable and encourages (not forces) good standards with respect to not using too many arguments and avoiding deeply nested loops. So while it may feel slightly cramped, if we're going to follow common standards, I personally prefer to go with the defaults.

cmkohnen:
Okay, I'll take some time later to change quotes to double and check which lines might be getting too long using plain default.

tox.ini
```diff
@@ -0,0 +1,13 @@
...
+ignore = U101
+	# temporarily disable docstring related stuff
+	D100, D101, D102, D103, D107, D204, D205, D400, D401
+inline-quotes = double
+
+per-file-ignores =
+	main.py: F401
+	# As long as numpy should be available in its current state
+	add_equation.py: F403, F401
+	transform_data.py: F403, F401
```
SjoerdB93, line 13: I like this ability to do per-file-ignores. There's a sensible reason for using star imports here (so that a user can just type sin(X) instead of np.sin(X)), but I don't want to ignore the rule all together as in general theses kind of blank imports are a bad idea. (Hence these functionalities being isolated in their own file)

SjoerdB93 (approved):

PR #138, opened 2023-02-22
Add warning dialog, when overriding data

cmkohnen:
Adds a warning Dialog if trying to open a project whilst data is loaded.
grafik


SjoerdB93:
I would go for "Opening a new project will discard all open data", but I think it's a good thing to have such a warning dialog yes!

cmkohnen:
Yes, that seems fitting, as nothing will be overwritten on the actual file system

cmkohnen:
grafik

SjoerdB93 (approved):

PR #140, opened 2023-02-23
Define color button in blueprint

cmkohnen:
Colorpicker is no longer dependant on Gtk.Button, instead taking the already compiled(?) button from samplerow.
This should help ease the transition for Gnome 44

### DIFF
--- a/data/ui/sample_box.blp
+++ b/data/ui/sample_box.blp
@@ -21,6 +21,15 @@ template SampleBox : Box {
       hexpand: true;
     }
 
+    Button color_button {
+      tooltip-text: _("Pick a color");
+      styles ["flat"]
+      Image {
+        hexpand: false;
+        icon-name: "color-picker-symbolic";
+      }
+    }
+
     Button delete_button {
       tooltip-text: _("Remove");
       styles ["flat"]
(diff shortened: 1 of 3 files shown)


SjoerdB93:
I personally think this looks cleaner as well. Will test it out after I've tested the responsive PR (which is compiling the libraries right now).

data/ui/sample_box.blp
```diff
@@ -21,6 +21,15 @@ template SampleBox : Box {
       hexpand: true;
     }
 
+    Button color_button {
+      tooltip-text: _("Pick a color");
+      styles ["flat"]
+      Image {
+        hexpand: false;
+        icon-name: "color-picker-symbolic";
```
SjoerdB93, line 29: Add the following to restore the slightly larger size of the color picker:

        pixel-size: 20;

SjoerdB93 (changes requested):

data/ui/sample_box.blp
SjoerdB93: Otherwise this looks good, a bit cleaner solution like this.

SjoerdB93 (approved):
Not sure why aarch build is taking so long

PR #141, opened 2023-02-24
miscallaneous fixes

cmkohnen:
- fix transform window, equations not showing.
- shorter way to add toast to main window
- swap print() for logging
- define log verbosity via meson (future pr for proper dev version comming soon)
- catch exception trying to load an invalid project


cmkohnen:
Ah, you already pushed a fix, fixing transform partially

SjoerdB93 (approved):
This looks good, and I've tested it and I haven't found any obvious problems. In the end it would be good to handle potential issues with logging indeed.

PR #142, opened 2023-02-25
Proper Development build

cmkohnen:
If the build type is development, enable verbose logging and change the AppID, Name and Style

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 desktop_file_in = configure_file(
-        input: application_id + '.desktop.in.in',
+        input: application_id.replace('.Devel','') + '.desktop.in.in',
        output: application_id + '.desktop.in',
 configuration: conf
 )
@@ -20,7 +20,7 @@ if desktop_utils.found()
 endif
 
 appstream_file_in = configure_file(
-        input: application_id + '.appdata.xml.in.in',
+        input: application_id.replace('.Devel','') + '.appdata.xml.in.in',
        output: application_id + '.appdata.xml.in',
 configuration: conf
 )
@@ -39,7 +39,7 @@ if appstream_util.found()
 endif
 
 gschema_file = configure_file(
-        input: application_id + '.gschema.xml.in',
+        input: application_id.replace('.Devel','') + '.gschema.xml.in',
        output: application_id + '.gschema.xml',
 configuration: conf
 )
@@ -74,7 +74,7 @@ blueprints = custom_target('blueprints',
 
 resources = gnome.compile_resources(
   application_id,
-  application_id + '.gresource.xml',
+  application_id.replace('.Devel','') + '.gresource.xml',
   dependencies: blueprints,
   gresource_bundle: true,
   source_dir: meson.current_build_dir(),
@@ -90,7 +90,7 @@ install_data(
 
 symbolic_dir = join_paths('icons', 'hicolor', 'symbolic', 'apps')
 install_data(
-  join_paths(symbolic_dir, ('@0@-symbolic.svg').format(application_id)),
+  join_paths(symbolic_dir, ('@0@-symbolic.svg').format(application_id.replace('.Devel',''))),
   install_dir: join_paths(datadir, symbolic_dir)
 )
 
--- a/data/ui/window.blp
+++ b/data/ui/window.blp
@@ -4,7 +4,6 @@ using Adw 1;
 template GraphsWindow : Adw.ApplicationWindow {
   default-width: 1200;
   default-height: 700;
-  title: _("Graphs");
   Box {
     orientation: vertical;
     HeaderBar header_bar {
--- a/meson.build
+++ b/meson.build
@@ -21,14 +21,24 @@ python = import('python')
 
 prefix = get_option('prefix')
 datadir = get_option('datadir')
+bindir = get_option('bindir')
+version = meson.project_version()
+project_name = meson.project_name()
 pkgdatadir = join_paths(prefix, datadir, meson.project_name())
 moduledir = join_paths(pkgdatadir, meson.project_name())
 localedir = join_paths(prefix, get_option('localedir'))
+debug = get_option('buildtype') == 'debug'
+
+if debug
+  version += '-' + run_command('git', 'rev-parse', '--short', 'HEAD', check: true).stdout().strip()
+  name += ' (Development)'
+  application_id += '.Devel'
+endif
 
 conf = configuration_data()
 conf.set('PYTHON', python.find_installation('python3').path())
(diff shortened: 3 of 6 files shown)


SjoerdB93:
This is definitely something that is nice to have, although I'm unsure if the app id itself should be different. Instead I think the proper way is to actually define a different branch with the same appid as most gnome projects do.

Then I can actually put up the Dev branch available in the beta branch of Flathub. (It's super easy to add a beta branch, it's literally just a second branch in the flat hub repo, doesn't require a new PR or something).

cmkohnen:
Having a different id is what (almost) all projects I've looked at use. But having a quick look at the flathub beta feature, I'll rewrite this tomorrow, to follow the same appid.

SjoerdB93:
It could very well be I’m mistaking or that Flatbub beta branch is an exception.There’s not that much documentation on it,  but to create a beta branch literally all you need to do is create a new branch on the Flathub GitHub repo where you point to the appID of the beta branch of the project. I’ve done this in the past, but didn’t bother to maintain such a branch because it’d be just another repo to maintain and I can just as well use the regular repo for that and use the releases feature on GitHub for stable releases.Skickat från min iPhone26 feb. 2023 kl. 23:16 skrev Christoph Matthias Kohnen ***@***.***>:﻿
Having a different id is what (almost) all projects I've looked at use. But having a quick look at the flathub beta feature, I'll rewrite this tomorrow, to follow the same appid.

SjoerdB93:
> Having a different id is what (almost) all projects I've looked at use. But having a quick look at the flathub beta feature, I'll rewrite this tomorrow, to follow the same appid.

Just as an example, this is literally all there is to it. Just adding a new beta branch in the flathub repo, which I have full authority to do. Here's the branch for the previous appid.
https://github.com/flathub/se.sjoerd.DatMan/tree/beta

I just checked and enabled the flathub beta repo in GNOME software, and then installed the builder nightly. It uses the same appid, but on a different branch. The branch itself is from flathub if I'm not mistaken. So you just have to point the flathub repo the correct development branch (whatever it's called), and it will automatically be a different branch than the stable version. So I don't think that's something that should be set up.

image

Maybe we should check what is the "correct" practice. But it may be possible to point the flathub beta branch to the main branch on github. And set the stable branch manually to the commit of a new release.

p.s. Apologies for the relatively low activity from my side the past weeks. It's been pretty crazy with my personal schedule, I'm getting married this thursday and on the same time my PhD thesis deadline is looming as well. Am expecting to have some more time next week when I'll be away on a trip for measurements. The amount of available time, as well as the priority list will always come and go, just saying that you shouldn't see it as a sign of lessened interest or lower ambitions with this project. In fact, the fact that I will continue with a postdoc in the same field, and now specifically with scientific development of python code kinda ensures I will keep have a personal interest in this project for the coming years as well. (Apart from just fun to work on, Graphs is a tool I actually use in my daily work)

cmkohnen:
I am sorry for not replying sooner,
as we are targeting purely flathub, I think keeping it in line with other apps on there and just pointing the beta build to this repos main  branch would be a good solution. We'd just have to look out for runtime changes, etc. which might then break upstream beta or just tag beta versions on this repo too and point the beta branch to these commits. 

Regarding low activity:
Getting married definitely has priority in life. Congratulations to the happy couple by the way!
My graduation is comming up and I myself will probably do less in the next month. So no worries! It is a project in our free time none the less...

SjoerdB93 (approved):
This works great, even includes the striped title bar as nightly builds should have!

SjoerdB93:
Just as an FYI, this latest build (including xrdml support) is on the Flathub beta now. Builds fine there, also for aarch64.

You can get it on the flathub beta branch, although it usually takes a while (up to six hours) before it actually reaches Flathub itself. All seems well, including the seperate development name, icon and striped headerbar. Even the version number gets a suffix to signify to which commit it belongs.

Given the development build is clearly differentiated from an actual release, I think we can push the version number up to 1.5.0 (and then it adds a suffix automatically as well of course, which is good), signifiying this is the dev build for the 1.5 release.

At one point we should also plan when to release the 1.5.0 build. Minor fixes and additions can be pushed in smaller point releases then. However, there's probably still some stuff to be done, including testing of course. There's also a bunch of features I want to add in at some point, but not sure if I'd push them in 1.5.0 or in a later update as I still have plenty of other stuff to do for the foreseeable future. Main things I'm thinking of still adding to the 1.5.0 release are a way to specify the axis ranges manually by typing them out (probably put in plot settings), and a way to change the order of the items.

I will probably put that order option in the item edit, thinking of having a drop down option there with the number of where the item is placed. So imagine there's five items, it will be ordered like this: [A, B, C, D, E]. Item A will be number 1 in the drop down, item B will be number 2 etc... If the user then has the settings open for item number C, and changes the 3 in the drop down to number 2, the new order should be [A, C, B, D, E]. If they would change it to number 5 instead, it should be [A, C, D, E, B].  The main purpose is that this allows to change which line is in front when plotting multiple data sets on top of each other. Will see when I get the change to implement that, but it shouldn't be that difficult. To change the axis range shouldn't be that difficult either, the foundations to change this programatically are already there. But again, I will probably need a few free hours then.

 Also before releasing we need to write a "proper" change list for the appdata, make new screenshots for Flathub, and I'll write up a small piece to get it featured in This Week in Gnome. But those are all planned to be done near the very end of the release of course.

cmkohnen:
I already tried out the beta branch. It works quite well.

Having the 1.5 release fairly soon, might be a good idea. Having the current codebase, which is in a good state imo atm as a basis to iterate on, introducing features might be easier to implement.

What about dragging the items in normal list view to change their order?

SjoerdB93:
> What about dragging the items in normal list view to change their order?

Yes, that would ultimately be the cleanest solution, far better also than doing it via a dropdown menu. Not 100% sure how doable it is to implement this nicely.  Of course we can probably just let it re-order based on a "cllicked" and "released" signal, but that would maybe a bit rough and unanimated. 

I quickly looked online, and found one thread with someone looking for the same thing and claiming they found a way to fix this. However, the edit in the first post made me a bit unsure if they did solve it. I send this person a message, and asked if they managed to do this and have an example code somewhere.
https://discourse.gnome.org/t/drag-and-drop-items-in-gtk-listbox-with-gtk-4/9083

Of course, it's not an essential feature for 1.5. But it is one thing that I want to have in there eventually. (Same with the manual axis limits)

SjoerdB93:
I got a reply from the person in this thread, and they made a repo with a demo demonstrating this:
https://gitlab.gnome.org/pervoj/vala-drag-n-drop-demo

It works really well, but it's written in Vala. I've tried to apply the same logic to Python code, but couldn't get it really working today. Feel free to try yourself, or otherwise I might give it another look next week or so.

PR #144, opened 2023-02-26
Refactor v3

cmkohnen:
Scope is a bit broader this time:
rather than surface level refactoring I'm trying to streamline the code as much as possible.
Deduplicate as often as possible and keep things consistent


cmkohnen:
Just a quick description of the new as is regarding operations (former item_operations)

Performing an action has changed drastically. To generalize and deduplicate code, any transformation is performed by calling `operations.operation` with the parameters self, the actual operation callback and any arguments passed into it (e. g. texts from translate/multiply entries)
The actual operations themselves all take in the self instance, an item, on which data should be performed and any passed arguments. They return the new x- and y-data for that item. They are completely agnostic to whether we are in selection_mode as the items only contain selected data.
The actual looping over selected keys and applying modified data is handled in the `operation` function. 
This enables the `transform` action to be applied like this as well.
The only current exception to this is the combine action, which interacts with the graph, adding new items, etc.

cmkohnen:
To elaborate further on what I've done:

- See previous comment for changes to operations
- graphs.py is now home only to loading projects/files and adding/selecting/deleting items
- To add an item simply use `graphs.add_item()`, logic regarding auto-renaming, adding to the list and refreshing the plot is handled there
- Canvas related widgets, such as the dummy toolbar and span selector have been moved to canvas.py
- canvas can now plot an items data by just using the `plot()` class function
- all windows now have relevant functionality as class functions
- various small fixes and changes


I know this is a huge changelist (again) but now I'm happy with the codebase. It feels much better and cleaner (in my opinion) and it should be easier to maintain and develop features going forward.
This also eliminates the problem, which I had doing development, that many things were duplicated across functions (e.g. adding data from files versus from equation) and changing something at one place leaves the very same thing unchanged at the other location.

cmkohnen:
I've pulled in fixes from https://github.com/SjoerdB93/Graphs/commit/944cad33933cde7c62125c9e8eef7f6358ab7ef0 so operations should work as expected now

cmkohnen:
No idea why aarch64 build is failing rn

SjoerdB93:
Just a comment on the slow merge here. Most operations still don't work for me on this build. I've worked a bit on this yesterday and fixed some of them in a seperate branch: https://github.com/SjoerdB93/Graphs/tree/pr145_bugfixes_wip2

Most noticably, there's some offset in the selected data. Seems worse when the x-axis doesn't start at 0. This is not there in the current main branch. Will look closely when I have time, but right now there's an article I need to get out asap so still a bit stacked with work. 

Once this all works, I think it's a nice improvement code-wise. It allows everything to be consistent, making it all very maintainable.

cmkohnen:
Made a few bugfixes (again). After I pulled fixes from https://github.com/SjoerdB93/Graphs/tree/pr145_bugfixes_wip2 some weird issues started popping up. This fixes them. The offset might be related to the span_selector, but will look further

SjoerdB93:
The thing that confused me last time is that span_selector does seem to work properly. However, it seems when it actually puts in the data so to speak, it puts it back in the wrong data. See attached video where I've got a print for appropriate index. It correctly identifies that it should cut from 1.7 or so, but in the graph you can see it cuts from 2 and upwards.

[Screencast from 2023-03-15 11-33-21.webm](https://user-images.githubusercontent.com/68477016/225283561-4c3844b2-73d2-41e9-be61-4da3741e87f3.webm)

SjoerdB93:
Fixed it, this was such a mysterious little thing that was hidden in a small unrelated part!

The error is in the plot function of the canvas class. It overrides the axes it plots graphs on that should be on bottom-left. One conditional was lost along the way, the correct code there should be like this:

```
    def plot(self, item, selected):
        x_axis = item.plot_x_position
        y_axis = item.plot_y_position
        if y_axis == "left":
            if x_axis == "bottom":
                axis = self.ax
            elif x_axis == "top":
                axis = self.top_left_axis
        elif y_axis == "right":
            if x_axis == "bottom":
                axis = self.right_axis
            elif x_axis == "top":
                axis = self.top_right_axis
```

SjoerdB93:
I've tested other functions as well, including with different data sets. And now everything seems to work! 
If you can fix that code snippet, I can finally mark this as approved for merging. Then the other open PR's seem to be okay for merger as well as the bugs here were the only remaining blockers for those as far as I could see.

SjoerdB93:
Regarding the aarch64 build failing, I've pushed the latest master commit (not this one, but the latest commit there has the same issue) to the beta channel on Flathub, where it also builds the aarch64 build. So far it seems to go okay. If it builds correctly it may be something in the workflow itself.

We could ask ourselves if aarch64 build is neccesary for these checks, given flathub does it as well before I can push a release. But it might be useful anyway given it works.

https://github.com/flathub/se.sjoerd.Graphs/pull/1

SjoerdB93:
Finally, and it may be redundant to add, but I've become a bit more familiar now with how this refactor worked after troubleshooting. And I have to say that I really like how you changed the way operations are handled. It's really elegant how the operations are created dynamically using functional coding, nice work! Really happy how this looks like now, adds a lot to longer term maintainability.

cmkohnen:
I've literally JUST fixed the last pushed bug

cmkohnen:
Thanks for fixing these two bugs, I wanted to implement one small change (Moving self out of each operation), which I will adress later.
Also I'm still unsure whether to disable linting for operations.py or explicitly naming everything _item in there.
Will work further on things, when I have time later

SjoerdB93:
> I've literally JUST fixed the last pushed bug

Weird, looking at the commit history it seems to have appeared in commit db15b96. So probably when the changes from my branch where merged in this code. It also seems that it's there where the linting started to complain. And indeed I see you also had fixed this by using the underscore for the unused variable. 

The issue was probably that the branch that I made earlier with some bugfixes was slightly out of date, and didn't contain this fixed linting. Then when those fixes got merged into this, the "linting bug" that was present overwrote the fixed code.

This last bug fix that I pushed seems to be specifically from my bugfix branch. It worked fine a few hours ago, so not sure what happened there. Probably when I tested I was on a slightly older commit maybe?

cmkohnen:
Oh, my bad, I was about to push the exact same fix, but you beat me to it is what I meant.

SjoerdB93:
> Also I'm still unsure whether to disable linting for operations.py or explicitly naming everything _item in there.
> Will work further on things, when I have time later

Yes, I was a bit unsure about this as well. Using an underscore for unused variables is the proper way to do if I recall correctly, so that seems reasonable. I think flake8 may give a U101 (unused argument with underscore warning), but we could set it to ignore that specific error-code. I think it's probably good to keep it checking for U100 (unused argument without underscore) anyway, as to avoid having unneeded arguments in there in general. 

No rush of course with further work or chances, as I'm also still pretty loaded with work and will/should get back to my actual work now as well. Again, now I'm a bit more familiar with the refactored codebase I've really started to appreciate it. It's a lot more elegant than the old structure where *lots* of stuff had to be repeated for each single operation. Now it's just a matter of adding only the relevant part of the operations when introducing new functionality (and coupling it to an action), and if we want to change the way operations are handled we can do this centrally. It's really in a better place like this, great work!

PR #145, opened 2023-03-01
Transform discard

cmkohnen:
based on https://github.com/SjoerdB93/Graphs/pull/144

Adds ability to discard unselected data, when using transform

### DIFF
--- a/data/ui/transform_window.blp
+++ b/data/ui/transform_window.blp
@@ -24,12 +24,17 @@ template TransformWindow : Adw.Window {
       margin-end: 26;
       margin-top: 12;
       Adw.EntryRow transform_x_entry {
-        use-underline: true;
         title: _("X =");
       }
       Adw.EntryRow transform_y_entry {
-        use-underline: true;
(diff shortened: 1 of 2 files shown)


SjoerdB93:
There were quite a bunch of features broken with this PR (most of the operations). I started digging into a a few of them, and solved some but not all this morning but didn't have the time to fix all I could find. I've pushed those to a different branch: https://github.com/SjoerdB93/Graphs/tree/pr145_bugfixes_wip

cmkohnen:
Oh, that is a massive oversight on my end. These were introduced somewhere in https://github.com/SjoerdB93/Graphs/pull/144. Some of these I haven't given much thought whilst changing to be honest.

SjoerdB93:
Just to add on previous comment, as I didn't respond to this PR itself. I do like this, in particular how this button appears contextually. That's a nice touch.

SjoerdB93:
This works well now for me!

SjoerdB93 (approved):

PR #146, opened 2023-03-01
Item edit

cmkohnen:
Based on https://github.com/SjoerdB93/Graphs/pull/144

Changes the way of how to edit an item:

- Use button instead of double click
- Use own window for item settings and keep plot settings for the plot itself (and enable changing stuff, without having data loaded)


SjoerdB93:
I don't mind having the plot settings in it's own window, I've been thinking about separating these before as well. 

Changing double click action to a button would be good if this is no longer accesible through the menu, as it adds discoverability. However, I think the pencil icon is maybe a bit too prominent here right now. What about using a small cogwheel, with roughly the same size as the color chooser circl=

cmkohnen:
grafik
This does indeed look better

SjoerdB93:
I think this looks quite good! After the open bugs in #144 are merged, this can be merged as well imo

cmkohnen:
Added back functionality to change the currently selected Item

src/canvas.py
```diff
@@ -242,13 +241,18 @@ def set_legend(self):
...
             lines3, labels3 = self.top_left_axis.get_legend_handles_labels()
             lines4, labels4 = self.top_right_axis.get_legend_handles_labels()
-            self.top_right_axis.legend(
-                lines + lines2 + lines3 + lines4,
-                labels + labels2 + labels3 + labels4, loc=0, frameon=True)
+            new_lines = lines1 + lines2 + lines3 + lines4
+            new_labels = labels1 + labels2 + labels3 + labels4
+            labels = []
+            for label in new_labels:
+                labels.append(utilities.shorten_label(label))
```
SjoerdB93, line 253: Very long labels make the plot unworkable, so it may be useful to shorten the label in the legend as well (like in this code), but 20 in the legend seems a bit low. In the end, this should maybe also be optional in the settings. 

This length of 20 works perfect for the item list in the panel). As a quick solution, I'd suggest having the max length as an optional argument with default value 20 in the utilities function, and then passing e.g. 40 for the max length in this piece here.

SjoerdB93 (changes requested):

src/canvas.py
cmkohnen: Agreed, this seems like a scalable solution

PR #151, opened 2023-03-15
Bugfixes + Fill in name automatically in plot settings

SjoerdB93:
Fixes some bugs, plus automatically fills in the current name of an item in the plot settings. I've found while working with Graphs that you often just want to change a small thing in the name (fix a typo for instance, or add some information). It's annoying if you have to retype the entire name. It's also more consistent with general behaviour to have the name filled in that field already, and let it react if the user changes is. (Such as already is the case with plot settings).

image

Fixes the following issues as well:
- Change of y-scale was broken as it was looking for graphs.graphs.refresh instead of graphs.refresh
- Y_range was not accesabile in transformation tools due to a capitalization issue
- Canvas limits are refreshed now after every operation (so that the graph doesn't dissapear out of sight when nornalizing for instance)
- Centered data is now sorted as well (ran into some bugs otherwise iirc)
- Plot settings now properly refreshes when switching sample (it tried to call save_settings, but that function has been renamed on_close)

### DIFF
--- a/data/ui/plot_settings.blp
+++ b/data/ui/plot_settings.blp
@@ -28,8 +28,7 @@ template PlotSettingsWindow : Adw.PreferencesWindow {
       }
 
       Adw.EntryRow name_entry {
-        title: _("Change Name");
-        max-width-chars: 25;
+        title: _("Item name");
       }
 
       Adw.ComboRow plot_y_position {
--- a/src/calculation.py
+++ b/src/calculation.py
@@ -26,7 +26,7 @@ def operation(xdata, ydata, input_x, input_y):
     x_array = []
     y_array = []
     X_range = xdata
-    Y_Range = ydata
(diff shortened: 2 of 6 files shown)


cmkohnen:
Ah, i must have pushed the graphs.graphs commit to the wrong branch

cmkohnen (approved):

cmkohnen:
Regarding setting the name, I implemented this in https://github.com/SjoerdB93/Graphs/pull/146

SjoerdB93:
I see you've also saw the same issue with the "Change name" being blank by default, given your PR #146 also has it filled in by default. Still think it's valuable to have the dropdown selector there as well, so you don't have to back out each time.

Still the idea of #146 to split plot settings from item settings is a good one, as there's no reason to have these coupled making it a bit confusing experience.

PR #153, opened 2023-03-16
Hide info in transform window behind popover

SjoerdB93:
This hides the explanation in the transform window behind a popover, making that window a lot cleaner. Open to suggestions about the exact placement of this info button:

image

image


cmkohnen (approved):

SjoerdB93:
Yes, that sounds much better!

SjoerdB93:
Right now it's in the top-left, basically mirrored with the close button.

image

PR #155, opened 2023-03-16
Update main.yml

cmkohnen:


cmkohnen:
Updating CI might fix failing builds

cmkohnen:
Nevermind, having them up to date is a benefit either way

SjoerdB93:
For some reason it keeps giving a 404 on the Gnome sdk on aarch64.  Error: Failed to install org.gnome.Sdk: While pulling runtime/org.gnome.Sdk/aarch64/43 from remote flathub: Server returned HTTP 404

Flathub's buildbot works fine, so the application itself builds fine on aarch64 in the real world. Not sure what's wrong here.

SjoerdB93 (approved):
x86 builds fine though, so no use on staying with old versions here.

SjoerdB93:
In the meantime I'm considering if we should just disable aarch64 in this workflow, and test-build those with flathub releases. When pushing this to flathub, it builds for both x86 and aarch64 before I can merge it in the main manifest. There it works fine.

https://github.com/flathub/se.sjoerd.Graphs/pull/2

PR #157, opened 2023-03-16
use proper snake_case

cmkohnen:


SjoerdB93 (approved):

SjoerdB93:
Interesting enough, the new uncapitalized variable names do not break projects. This is due to the fail-safe we build for this a while ago. When it imports data from a project file, it actually generates a new data object using the Data class. All relevant properties (the ones that are changed) are then created in the init function. 

Then, it tries to apply the properties that are stored in the project file on the data object. If these no longer match up, it just keeps the ones that were created in the init.

So the old project files still work fine. It's just that some specific settings (e.g. on which axis the data set is plotted) are reset to the default values. This should still be avoided of course, but it's nice to see the way these chances are handled are relatively robust.

PR #158, opened 2023-03-16
Add support for xrdml files

SjoerdB93:
Adds support for xrdml files. 

xrdml is the default filetype for X-ray diffraction measurements performed by diffractometers from Panalytical (one of the bigger players in this world). The reason I've added support for xrdml specifically is because our university uses Panalytical diffractometers (so a lot of my data is in xrdml), and the file format is "easy enough" to work with. (It's open in the sense that it's basically just a xml file)

Automatically converts intensity to counts per second and detects what kind of measurement is stored, and then gives appropriate labels and axis limits.

Also had to fix some remaining config names that still had a capital letter.

[Screencast from 2023-03-16 21-17-50.webm](https://user-images.githubusercontent.com/68477016/225743038-5ce3c186-75e4-4f49-8e5a-79ea00265b9e.webm)

### DIFF
--- a/src/file_io.py
+++ b/src/file_io.py
@@ -2,7 +2,7 @@
 import os
 import pickle
 import re
-
+from xml.dom import minidom
 from graphs import utilities
 
 import numpy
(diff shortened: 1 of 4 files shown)


cmkohnen (approved):
Regarding snake case I'm a bit puzzled as to why they didn't show up in my GNOME Builder search.

In the future we might want to add even more "common" formats such as csv. A flexible system like the current one should scale quite well.

PR #163, opened 2023-03-23
migrate to GNOME 44

cmkohnen:
Update Sdk and packages to the newest version. Also fix error when adding new items.


cmkohnen:
I had to implement a quick dialog for file format when exporting the figure as FileDialog doesn't support choices. This enables to expand for future choices like transparent background, etc.

SjoerdB93:
> I had to implement a quick dialog for file format when exporting the figure as FileDialog doesn't support choices. This enables to expand for future choices like transparent background, etc.

From the docs, it does seem there's there's functions to set these choices except that they're called filters now: https://docs.gtk.org/gtk4/class.FileDialog.html 
A dialog window could still be an option by the way if we squeeze in some more options there. 

The modal ColorDialog does remove a bug as well. In the popover implementation, if you select a custom color using the "+" icon, the entire row would get stuck in a activated/selected state. This is some upstream GTK bug which people were aware off, and the solution at the moment was just to avoid the conditions that trigger this bug.

SjoerdB93:
I see, you are correct. I only checked the documentation for the new FileDialog, and never checked the old one: https://docs.gtk.org/gtk4/iface.FileChooser.html#methods

Choices were indeed a separate thing from the filters. Not sure if filters can be used to simulate the same behavior we were using for choice now. But as said, if we extend the new dialog a bit I'm perfectly happy with this solution with a separate dialog. Basically, the dialog can be a front-end to the matplotlib savefig options as documented [here](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html). 

 It makes sense to do that in a separate PR, as the scope of this PR should mainly about migrating to GNOME 44 and GTK 4.10.

SjoerdB93:
> Also the FontChooser should be migrated as well.

I played around with this quickly. We probably want to add the FontDialog in the .ui file, but either way this is an easy migration and the new FontDialogButton supports the same features with the same syntax as the old FontChooser.

Basically, I just changed the FontButton (which is now deprecated) in the blueprint to FontDialogButton. Then immediately the font button works fine, but you still need to add the FontDialog itself to the button (by default it's None). Programatically, this is easy using:

```
        font_dialog = Gtk.FontDialog.new()
        self.plot_font_chooser.set_dialog(font_dialog)
```

And that's seems to be it. I haven't tried putting the FontDialog as a child to the FontDialogButton in the Blueprint file. My initial guess is that that may be the way to do it. But I didn't really want to push to much to this PR in order to avoid clashes. So instead, thought I'd just write my findings here :)

SjoerdB93 (approved):

PR #164, opened 2023-03-25
Export figure improvements

cmkohnen:
Add transparent background option, change wording a bit and add a confirmation toast


SjoerdB93 (approved):
Looks good to me, works nicely!

PR #165, opened 2023-03-26
Refactor v4

cmkohnen:


cmkohnen:
Apologies for the delayed elaboration and activity.
This pr tackles a few minor things, that kept bugging me when trying to work with the codebase. Detailed explanations should (!) be inside each commit, as I tried to explain things in commits and use multiple commits for different things I've done.

tox.ini
```diff
@@ -1,7 +1,7 @@
 [flake8]
 count = True
 #default_ignorelist: E121,E123,E126,E226,E24,E704,W503,W504
-ignore = U101
+ignore = U101, E251
```
SjoerdB93, line 4: Any specific reason to ignore E251 as well?


```
Unexpected spaces around keyword / parameter equals (E251)

There should be no spaces before or after the = in a function definition.
```

src/ui.py
```diff
@@ -6,7 +6,7 @@
 from graphs import file_io, graphs
 
 
-def toggle_darkmode(_shortcut, _theme, _widget, self):
+def on_style_change(_shortcut, _theme, _widget, self):
```
SjoerdB93, line 9: Makes sense with the renaming in GNOME 44 from Dark Mode to "Dark Style"

image

src/transform_data.py
SjoerdB93, line 1: good

src/plotting_tools.py
```diff
@@ -200,11 +200,11 @@ def change_top_xscale(action, target, self):
 
 def change_bottom_xscale(action, target, self):
     if target.get_string() == "log":
-        self.canvas.ax.set_xscale("log")
+        self.canvas.axis.set_xscale("log")
         self.canvas.right_axis.set_xscale("log")
         self.plot_settings.xscale = "log"
     else:
-        self.canvas.ax.set_xscale("linear")
+        self.canvas.axi.set_xscale("linear")
```
SjoerdB93, line 207: Probably should be axi**s** right?

data/se.sjoerd.Graphs.gresource.xml
```diff
@@ -3,37 +3,37 @@
...
-    <file preprocess="xml-stripblanks" alias="shift-symbolic.svg">icons/shift-symbolic.svg</file>
+    <file preprocess="xml-stripblanks" alias="info-symbolic.svg">icons/info-symbolic.svg</file>
+    <file preprocess="xml-stripblanks" alias="integral-symbolic.svg">icons/integral-symbolic.svg</file>
     <file preprocess="xml-stripblanks" alias="inverse-fast-fourier-transform-symbolic.svg">icons/inverse-fast-fourier-transform-symbolic.svg</file>
-    <file preprocess="xml-stripblanks" alias="fast-fourier-transform-symbolic.svg">icons/fast-fourier-transform-symbolic.svg</file>
-    <file preprocess="xml-stripblanks" alias="integral-symbolic.svg">icons/integral-symbolic.svg</file>    
-    <file preprocess="xml-stripblanks" alias="list-compact-symbolic.svg">icons/list-compact-symbolic.svg</file> 
+    <file preprocess="xml-stripblanks" alias="list-compact-symbolic.svg">icons/list-compact-symbolic.svg</file>
+    <file preprocess="xml-stripblanks" alias="loupe-symbolic.svg">icons/loupe-symbolic.svg</file>
+    <file preprocess="xml-stripblanks" alias="move-tool-symbolic.svg">icons/move-tool-symbolic.svg</file>
```
SjoerdB93, line 30: Will check these while testing the actual code on a live environment. Not sure why some of these seem changed. "Fast-fourier-transform" is gone?"

src/graphs.py
```diff
@@ -21,13 +23,16 @@ def open_files(self, files, import_settings):
                 import_settings.path = path
                 xdata, ydata = file_io.get_data(self, import_settings)
                 if xdata == []:
-                    filename = import_settings.path.split("/")[-1]
+                    filename = pathlib.Path(import_settings.path).stem
```
SjoerdB93, line 26: Not 100% what stem does, will check how functional it is but if it works I'm fine with this as I'm not 100% sold on using splits either.

SjoerdB93 (commented):
No need to apologise for delays. I've had way less time to work on this project than I'd like to as well. I need to hand in my PhD thesis at the end of April, so right now it's all kinda full stress-mode on that :) 

Either way, overall sensible changes. Few minor remarks before merging:

- At one point in the code it uses canvas.axi, I think that should be canvas.axis.
- Is there a specific reason to surpress E251?
- While we're changing some of the naming conventions (which I agree with), I propose to change the naming of "sample" through some places in the code as well (e.g. sample_box.py). The reason for this naming is that the vast majority of data that I am working with are measurements done on specific scientific samples I've grown, so all data is kinda the data representing a specific sample. But as a general naming convention it makes less sense. As in most places we talk of a data-set as an "item" right now in this PR (changing it from Data to Item also makes sense, as "Data" is kinda known as a bad name. Everything is data technically speaking). Maybe changing the mention of "sample" to "item" would make sense.

SjoerdB93:
I'll also rebase my own install to this PR, so I can test this out while also working on other stuff, simply by using it.

src/canvas.py
```diff
@@ -156,16 +159,16 @@ def set_style(self, parent):
...
+        cmap = self.parent.preferences.config["plot_color_cycle"]
+        if Adw.StyleManager.get_default().get_dark() and \
+                self.parent.preferences.config["plot_invert_color_cycle_dark"]:
             cmap += "_r"
-        color_cycle = cycler(color=pyplot.get_cmap(cmap).colors)
-        self.color_cycle = color_cycle.by_key()["color"]
+        self.color_cycle = \
+            cycler(color=pyplot.get_cmap(cmap).colors).by_key()["color"]
 
+    # Overwritten function - do not change name
```
SjoerdB93, line 171: These kind of comments can be very useful :+1:

src/canvas.py
```diff
@@ -349,7 +350,7 @@ def on_define(self, canvas):
 
     def get_start_stop(self, bottom_x):
         if bottom_x:
-            xrange_bottom = max(self.canvas.ax.get_xlim()) \
+            xrange_bottom = max(self.canvas.axis.get_xlim()) \
                 - min(self.canvas.ax.get_xlim())
```
SjoerdB93, line 354: Line 354, should be self.canvas.ax**is**

SjoerdB93 (changes requested):
See some bugs in the comments

tox.ini
cmkohnen: Just that I got it confused with something else...

src/plotting_tools.py
cmkohnen: yep, typo I didn't catch

data/se.sjoerd.Graphs.gresource.xml
cmkohnen: I just reordered them alphabetically, it is indeed there

SjoerdB93 (approved):
Seems good

PR #166, opened 2023-03-30
Update CI

cmkohnen:
Use GNOME 44 and re-enable aarch64, as it seems to be fixed upstream


SjoerdB93:
Great, nice that it's working again!

SjoerdB93 (approved):

PR #167, opened 2023-03-31
plot styles

cmkohnen:
Work for plot styles
- [x] support custom styles
- [x] make them selectable
- [x] ui for viewing them, copying and resetting to default ones
- [x] ui for editing


cmkohnen:
Huge changesize is mainly due to having the styles in this repo now

SjoerdB93 (approved):

cmkohnen:
> This is nice and seems to work well. There's some tweaks it could use (e.g. the ability to rename a style maybe, right now if you want to create a custom style you can only copy an old one and change that but it still has the same name as the old one apart from the (1)), but that can be done in a seperate PR imo.

Agreed. This is by far not at the point I want it to be, but "feature parity" was my minimum goal working on this pr, so that no merge conflicts would come from to big of a changeset.

PR #168, opened 2023-04-01
Set axis limits in Plot Settings

SjoerdB93:
Rudimentary implementation of limit settings, still a WIP.

TODO:

- [X] Retrieve axis limits and display them in Plot Settings window
- [X] Implement functionality to set canvas limits by changing plot settings 
- [X] Fix UI (probably will group each axis)
- [X] Make options for unused axis dissapear (e.g. right-axis canvas limit should only show when right-axis is actually used)


cmkohnen:
Would it be an option to have the selectors in ui as spinners? You could still input manually, and it is always a valid number.

SjoerdB93:
The problem with a spinner is that it's not straight forward to predict in what intervals you want the ticks to be, and that probably depends a lot on the situation. I looked around how other applications do this, and they overwhelmly go for just an entry box. But I cleaned it up a bit, and it looks a lot better now. Unused axes are hidden as well, and I've changed the default scaling to only use three significant digits (to go for sigfigs rather than rounding ensures this also works with weird scalings). 

Also fixed some bugs that arised with the last refactor. All should work fine now.

image

cmkohnen:
One thing I noticed: When no data is loaded none of the options are shown. Maybe add in a small notice, when that happens.

cmkohnen:
Other than that, this works great for me!

src/graphs.py
```diff
@@ -100,14 +100,15 @@ def add_item(self, item, reset_clipboard=True):
     self.main_window.no_data_label_box.set_visible(False)
     self.item_boxes[key] = ItemBox(self, item)
     self.main_window.list_box.append(self.item_boxes[key])
+    self.item_menu[key] = self.main_window.list_box.get_last_child()
```
cmkohnen, line 103: Maybe a way to not have this stored two times is to store just the ListBoxRow gotten from get_last_child(). If we need the actual ItemBox we could just query get_child()

SjoerdB93:
Good call on the message, it now displays a message when no data is loaded.

image

src/graphs.py
SjoerdB93: In the latest commit, I removed item_boxes dictionary, and store it directly in the list_box of the main window. As the dictionary itself is kinda redundant, and the ItemBox itself can be retrieved from item_menu using the same key.

So far it seems to work in my initial tests without new issues.

cmkohnen (approved):

src/graphs.py
SjoerdB93: Just FYI, this did introduce some bugs and I reverted this removal of item_boxes for now with a seperate commit.

PR #170, opened 2023-04-02
Improve Plot Style Editor

cmkohnen:


cmkohnen:
To elaborate on what is changed in this pr:
- Allow renaming of styles
- Extend the style editor with options for the grid, padding and axis width

cmkohnen:
next two steps would be:
 - query linewidth from actual style and do not distinguish between selected/unselected (unselected just becomes .6 times selected or something)
 - Add options to change colors
 These however deserver their own pr imo

SjoerdB93 (approved):
This is great, I was literally just thinking earlier that we at least should add the possibility turn on/turn off the grid in this style settings at the very least for now. Works good as well!

SjoerdB93:
>     * query linewidth from actual style and do not distinguish between selected/unselected (unselected just becomes .6 times selected or something)
> 

I've been thinking a bit about this, and I agree that it would be good to have the default linestyle in the Plot Style. This also allows people to make a style for a specific purpose, and always have it the same way. Having the colour cycle in there would make sense as well then for the same reasons. 

Then add one point we can make an interface to create a custom color cycle, but that's not really a priority to have that specific part in there for release 1.5. Which probably will be soon (tm).

PR #173, opened 2023-04-02
Change order of items by dragging and dropping the items

SjoerdB93:
The proof-of-concept is there, and it works quite well actually:

[Screencast from 2023-04-02 21-16-09.webm](https://user-images.githubusercontent.com/68477016/229374154-6d14f3e0-cba8-40ee-9d22-40e7b9ab04f1.webm)

TODO:

- [X] Fix UI (The dragging preview shouldn't be the ID, ideally it should show the row but at least it should show item name)
- [X] Code (check if code is OK enough, but at least fix linting, am aware some lines are too long atm)

### DIFF
--- a/src/item_box.py
+++ b/src/item_box.py
@@ -20,8 +20,6 @@ class ItemBox(Gtk.Box):
         self.check_button.set_active(item.selected)
         self.parent = parent
 
-        self.one_click_trigger = False
-        self.time_first_click = 0
         self.gesture = Gtk.GestureClick()
         self.gesture.set_button(0)
         self.add_controller(self.gesture)
--- a/src/item_box.py
+++ b/src/item_box.py
@@ -40,30 +40,26 @@ class ItemBox(Gtk.Box):
 
 
         self.drop_source = Gtk.DropTarget.new(str, Gdk.DragAction.COPY)
-
+        self.drop_source.key = item.key
         self.drop_source.connect('drop', self.on_dnd_drop)
         self.drag_source.connect('prepare', self.on_dnd_prepare)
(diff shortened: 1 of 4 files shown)


src/utilities.py
```diff
@@ -23,6 +23,17 @@ def sig_fig_round(number, digits):
     power = "{:e}".format(float(number)).split("e")[1]
     return round(float(number), -(int(power) - digits + 1))
 
+def swap_key_positions(d, key1, key2):
```
cmkohnen, line 27: 'd' should probably be 'dict' or 'dictionary'

src/item_box.py
```diff
@@ -33,6 +34,36 @@ def __init__(self, parent, item):
...
+        self.parent.item_menu = utilities.swap_key_positions(
+        self.parent.item_menu, drop_target.key, value)
+        for key, item in self.parent.item_menu.items():
+            self.parent.main_window.list_box.remove(item)
+            self.parent.main_window.list_box.append(item)
+        graphs.reload(self.parent)
+
+
+    def on_dnd_prepare(self, drag_source, x, y):
+        data = self.item.key
```
cmkohnen, line 62: would `utilities.shorten_label(self.item.name)` suffice to change to the tooltip to the actual name or is the key required for this to function?

cmkohnen (commented):
Looks good!
Apart from the flake comments, this looks clean

src/utilities.py
SjoerdB93: Yes that's probably better

cmkohnen (approved):

PR #179, opened 2023-04-03
More minor fixes

SjoerdB93:
See commits:
 - I just tested the old version on Flathub, and saw that today's build crashed. The issue was that the config file was not properly loaded since it moved to the data folder. So it loaded the actual config file instead so the fail-safe check didn't work. It works now.
 - Adds a failsafe for the style settings. Another issue is that the old default style for the plot didn't exist anymore. Now if it tries to load a style that doesn't exist, it just loads the adwaita style
 - Fix color cycle loop. Previous PR partially made it useable again by using black lines if the color cycle is full, now it loops around again. Also works if the order is messed up due to drag and drop :)
 - Fix drag and drop. Drag and drop functionality tried to reorder item_boxes which is deleted now, so removed that line.


src/canvas.py
```diff
@@ -19,8 +19,13 @@ class Canvas(FigureCanvas):
     """Create the graph widget"""
     def __init__(self, parent):
         self.parent = parent
-        style_path = plot_styles.get_user_styles(
-            self.parent)[self.parent.plot_settings.plot_style]
+        try:
+            style_path = plot_styles.get_user_styles(
+                self.parent)[self.parent.plot_settings.plot_style]
+        except KeyError:
+            self.parent.plot_settings.plot_style = "adwaita"
```
cmkohnen, line 26: maybe we shouldn't hardcode adwaita but use the first availabe style from get_user_styles()

src/preferences.py
```diff
@@ -19,7 +19,8 @@ def __init__(self, parent):
         self.check_config(self.config)
 
     def check_config(self, config):
-        with open("config.json", "r", encoding="utf-8") as file:
+        config_path = self.parent.modulepath
+        with open(f"{config_path}/config.json", "r", encoding="utf-8") as file:
```
cmkohnen, line 23: maybe something like
```
template_path = os.path.join(self.parent.modulepath, "config.json")
with open(template_path, "r", encoding="utf-8") as file:
```
would be cleaner here?

src/canvas.py
SjoerdB93: I think either works, but using the first available style is a bit less maintenance-prone as it doesn't have to be changed upon name changes. Changed it.

cmkohnen (approved):

PR #180, opened 2023-04-03
Add line options to plot styles

cmkohnen:
Allows to edit some line properties in the plot styles editor
namely: linestyle, linewidth, markerstyle, markersize
Each item can individually change their marker, but the default is queried from the currently active style upon item creation.
Also adds the option to hide unselected items from the plot. If this is set to false, every unselected Item will have a thickness of 0.6 times the style-provided one
Also removes old config keys no longer used


SjoerdB93:
> Each item can individually change their marker, but the default is queried from the currently active style upon item creation.

Which is good, but I do strongly feel that the linestyle, linewidth, and markersize should behave the same. That is, the line style, width etc.. should be queried from the currently active style upon item creation, but individual changes to specific lines should be possible using the edit_item menu. 

Otherwise it seems to work quite well, didn't see any bugs when testing this.

cmkohnen:
I was wondering which of these should be changable on a per-item basis and could only think of changing the markers this way to better separate them. But the option to change them doesn't hurt imo.

cmkohnen:
I also changed the properties Item() gets to just self, xdata, ydata and an optional name, as the rest can be queried from self (style, config and next color. The latter might change when I implement color cycle from style)

PR #181, opened 2023-04-03
Plot styles

cmkohnen:
Small changes:
- rafactor to use pyplot.rcParams instead of canvas.style
- query the prop_cycle from the current style


SjoerdB93 (approved):
The ability to change the colour cycle can later be added back into the plot settings menu. But it's a bit of a niche thing anyway, so it's fine that this is not in this PR. Also ideally, we may want to come up with an interface to make a custom color cycle anyway, which lends itself quite well with this code.

PR #183, opened 2023-04-04
Add Line colors to style editor

cmkohnen:
As the title says, adds a submenu to the style editor for specifying which colors to use when plotting
grafik
pink color for reference
grafik


SjoerdB93:
Yes, I really like this! Maybe it would be nice to add the orange colour back in as default secondary choice. Otherwise this is fantastic. The style editor is really turning out to be a powerful tool :+1:

SjoerdB93 (approved):

PR #184, opened 2023-04-04
Small changes

cmkohnen:
- apply textcolor for tick.labelcolor. This allows having the ticks in a different color than their labels
- Use hex notation for grayscale color cycle
- Set patch color as first line color (all styles already behave this way)
- Tweak default adwaita color cycle

### DIFF
--- a/data/styles/adwaita-dark.mplstyle
+++ b/data/styles/adwaita-dark.mplstyle
@@ -89,8 +89,8 @@ axes.facecolor: 242424
 figure.facecolor: 242424
 figure.edgecolor: 242424
 
-axes.prop_cycle: cycler('color', ['4C72B0', '55A868', 'C44E52', '8172B2', 'CCB974', '64B5CD'])
-patch.facecolor: 4C72B0
+axes.prop_cycle: cycler('color', ['1A5FB4', '26A269', 'E5A50A', 'C64600', 'A51D2D', '613583', '63452C', '9A9996'])
+patch.facecolor: 1A5FB4
 
 # other
 axes.linewidth: 0.8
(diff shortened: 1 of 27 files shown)


SjoerdB93:
In general it's nice, but I feel the color cycle colors are a bit too close to each other. Someone on the matplotlib github page in an issue actually did some research on this, and had some pretty nice suggestions, see full article: https://arxiv.org/pdf/2107.02270.pdf with the corresponding reply: [link](https://github.com/matplotlib/matplotlib/issues/9460#issuecomment-875185352)

I'd suggest using one of those suggested color cycles as standard as they're both aesthetically pleasing and really nicely distinguishable in general:

> 
> Six color:
> #5790fc, #f89c20, #e42536, #964a8b, #9c9ca1, #7a21dd
> 
> ["#5790fc", "#f89c20", "#e42536", "#964a8b", "#9c9ca1", "#7a21dd"]
> 
> Eight color:
> #1845fb, #ff5e02, #c91f16, #c849a9, #adad7d, #86c8dd, #578dff, #656364
> 
> ["#1845fb", "#ff5e02", "#c91f16", "#c849a9", "#adad7d", "#86c8dd", "#578dff", "#656364"]
> 
> Ten color:
> #3f90da, #ffa90e, #bd1f01, #94a4a2, #832db6, #a96b59, #e76300, #b9ac70, #717581, #92dadd
> 
> ["#3f90da", "#ffa90e", "#bd1f01", "#94a4a2", "#832db6", "#a96b59", "#e76300", "#b9ac70", "#717581", "#92dadd"]

cmkohnen:
I'm a bit conflicted on this. For the colors I've used all adwaita colors, for each color their 5th variant (see in color picker the last item). These fit in well with the adwaita styling we're using in general.
grafik
One option I could imagine would be to ship alternative styles (adwaita-b and adwaita-b-dark for example) with these color cycles

SjoerdB93:
Nah, for the main color cycles I think it indeed makes sense to use adwaita coloring, I didn't think about it's that where these came from. 

It also makes it easier/more logical with the color picker if the default prop colors are actually there.

SjoerdB93 (approved):

PR #186, opened 2023-04-05
Extend flake

cmkohnen:
Adds a few extensions to flake8 that improve code quality quite a bit. Also fix most of the newly reported issues right away.


SjoerdB93 (approved):
I'll have to see how the extra flake modules will turn out, but I feel it can't hurt. Approved.

PR #189, opened 2023-04-06
Change maintainer name

Sjoerd1993:
Sometimes I push typo's and "small" linguistic stuff directly to main (I know, bad habit), but this is a a major change to some extend, as the Github repo URL got changed, so I thought it's important to have this change visible.

To keep it short, I have taken my wife's last name. There's no particular reason for this, other than that my original last name (Broekhuijsen) is both difficult to spell and to pronounce in the country I'm living in (Sweden), and in Sweden it's very common to take the woman's name when you get married. So when we decided to take the same last name, we quite quickly landed on Stendahl purely for practical reasons.
Either way, my GitHub account had a B in the username, which is probably a bit confusing in the long-run now my initials have changed. So I changed my GitHub account to sjoerd1993.

This PR just changes these names to reflect this.
Github automatically sets redirects from the old URL to the new one, so that should all still work fine for the foreseeable future. But it's good to change all references from https://github.com/SjoerdB93/Graphs to https://github.com/Sjoerd1993/Graphs elsewhere.


SjoerdB93:
If someone opens a new account with my old ID, and then creates a repo called Graphs, the redirect will stop functioning. Just as an FYI, I have squatted the old Github account for this reason.  https://github.com/SjoerdB93/

Also, profile pages themselves don't automatically redirect (while the full repo url does), so I thought it's good to keep hogging that old ID for a while first.

meson.build
```diff
@@ -11,7 +11,7 @@ copyright = '2022 - 2023'
 homepage_url = 'https://www.sjoerd.se/Graphs'
 github_url = 'https://github.com/SjoerdB93/Graphs'
```
cmkohnen, line 12: Should this be changed too?

cmkohnen (approved):

PR #191, opened 2023-04-08
Rework preferred style

cmkohnen:
Rework preferred style.

Instead of choosing default styles for light and dark, the user can now choose to follow the system (default) or use a custom style. In Plot Settings the user can choose whether or not to follow the global setting (system or custom) or specify a custom style just for the project. The logic for differentiating which style to use is in a new function which is called upon canvas creation. This eliminates the need to update the style at different points in the codebase.

If the user deleted the system styles (adwaita and adwaita-dark) and they are queried (only when following the system) they are just copied to the users config folder again. If the user specified another pyth, which's file doesn't exist anymore, the system preferred is queried.

This also fixes some edge cases. e.g.:
- Setting custom plot style for a project and changing the system style (dark/light) 
- Style not available due to the file being deleted in some cases

Also the get_style() function now checks against every param in pyplot.rcParams to make sure every valid key is represented, be it the default.

This also allows to easily add high-contrast styles in the future by creating them and adding
```
if pseudo_get_high_contrast():
  system_style += "-high-contrast"
```
to the get_system_preferred_style() function


Sjoerd1993:
Sorry for the late-ish response, I was away during the eastern-weekend, but I've taken a look at this and tested this out. 


> Setting custom plot style for a project and changing the system style (dark/light)

This was an edge-case that annoyed me in the past honestly. As for my personal use-case Graphs has two kind of distinct use-cases:
1 - Plot data and visualize my own data. Look at the data, compare data-sets with each other and do some basic manipulations (translations, normalization, cutting data etc...). For this specific use-case, I like it to follow my system theme. As I do tend to switch back and forth regularly between light- and dark mode (mostly depending on my outside environment, I prefer light in most cases, but I prefer dark in evenings or in poorly lit rooms). When I then load data, and it's all white it's sticks out like a sore thumb and it's kinda blinding. So while working with data like that I want it to follow my system theme.
2 - Plot data, and make decent good-looking graph that I can use for presentations and publications (for publications, I typically do some final touch-ups in Inkscape, which works very well after exporting as svg). In these cases, I want to "design" the plot for my specific purpose and therefore the design should be static. If I reopen my plot in the evening when I have dark mode on, I don't want it to suddenly change colors, as it's not like my article will be published in dark mode.

So some balance had to be found to facilitate both mode of operations if that makes sense. Typically in the second use-case I save my graph as a project, so I can make other adjustments to it or continue where I left off. Also it's useful if I'm showing something to colleagues. For me it just made sense to be load the previous plot setting, as to just allow people to continue where they left off with the same settings. But I was aware that it did become weird when changing system theme. Partly it didn't adapt when opening new data, but also it override the old style when switching back and forth which is not good.

A bit of a needlessly long comment, but my point was that the behavior was something I have thought about. And in the end, it was kind of a case of the "lesser of two evils". I feel like this PR really solves this dilemma. Adapting to system theme (light/dark) should be the default mode of operation. But having a custom button to use a custom stylesheet still gives the power to fully work in the second use-case I described. 

This is also clearly the most intuitive behavior for the end-user in my opinion. So this is a real nice PR. I've tested this a bit and it works well. This really ties stuff much more nicely together also, given the new Plot Styles a much more intuitive behaviour. These new settings with the Plot Styles are really turning up to be among the highlights of the upcoming release!

Sjoerd1993 (approved):

PR #192, opened 2023-04-10
Misc. Changes and cleanups

cmkohnen:
- clean list creation
- update config keys and sort alphabetically
- plot_styles: update name
- preferences: update names & sort according to ui layout
- window: remove uneeded widget names
- window: bind no data status page's visibility to the item list
- window: renamelist_box to item_list
- toolbar: use builtin functions (codewise identical & cleaner)
- canvas/toolbar: just pass if trying to call save (this should never be the case)
- fix view actions
- add plot styles keybind
- plot styles do not check against pyplot (it broke things)


cmkohnen:
That would be the last bit of code cleanup for the comming-soon™ Release 1.5 from my side. With the plot styles editor completed, there aren't any features I'd want to introduce for now. However a small tweak to most dialogs would be to clamp them down. This will be in a following pr.
Still, quite a bit of testing is required for that release, which I'll start on next. 
Two things I'd want to move to an 1.5.1 version would be docstrings (documenting every function & module and stop supressing flake warnings) and proper localization support.

Sjoerd1993 (approved):

Sjoerd1993:
> That would be the last bit of code cleanup for the comming-soon™ Release 1.5 from my side. With the plot styles editor completed, there aren't any features I'd want to introduce for now. However a small tweak to most dialogs would be to clamp them down. This will be in a following pr.
Still, quite a bit of testing is required for that release, which I'll start on next.
Two things I'd want to move to an 1.5.1 version would be docstrings (documenting every function & module and stop supressing flake warnings) and proper localization support.

Yes, I was thinking similarly. For my personal point of view, the main things I wanted to have in the upcoming release was the ability to change the axis range numerically from plot settings, and the ability to reorder loaded data so that you can choose which ones comes in front. Both are implemented a while ago.

Then you also implemented the Plot Styles, which moved the goal posts a bit for my perspective, since if that were to be implemented it needed some for of polish before actually shipping it). I'm very happy that you introduced that functionality, as it is very powerful. It's very nice to just once make a nice layout that I like for my presentations and whatever, and then just pick that whenever I need to make a graph for that implementation. Then for general data visualization going for the default system style by default makes sense. So I'm very happy how that turned out, and it's definitely at a stage that we should ship it in the next release. It's also really nice in general. It cleaned up the other settings a bit, and the overall GUI of editing a plot style is just nice. 

But at this point I'd say it's mostly just testing, and ironing out kinks when we find them for the next release. Neither do I have time for writing big new things myself either as I need to hand in my thesis at the end of this month, but I've been doing some more minor stuff and bug-fixes in between work. I'll also try to see if I can create some new screenshots for the Flathub page, and I'll write a bit of text to get it featured in This Week in GNOME. For the flathub screenshots I was thinking of having a single line of description as well, highlighting some features. Something like this (just a quick mock-up, not an actual screenshot I saved):
image

Not sure how many Flathub screenshots are reasonable, but the features I was thinking to highlight in such screenshots those were:
- Import data from a wide range of filetypes (.txt, .csv, .dat, .xy, .xrdml and more)
- Generate new data using equations
- Manipulate data with ease
- Create your own plot styles
- Supports both light and dark mode 

So the idea was to have the bullet points above in the screenshot. But I'm open for suggestions, both in terms of exact phrasing and what to highlight there. I was also thinking of moving the screenshots to the Github repo. Right now they're hosted on my personal website, but if I host them on Github instead we can just update them with a PR. I will also probably update the website (graphs.sjoerd.se) at one point. I was thinking of at least having a description there, and to show how it works etc... Maybe even with some video. But I won't have time for that before next month, and it's not a priority for the next release either.

And of course some other formalities need to be done before that release, such as updating/creating a changelist, and we probably want to update the readme description as well.

PR #193, opened 2023-04-10
add file filters to file dialogs

cmkohnen:
Add file filters to file dialogs. Also implement a function to easily create a listsstore of file filters from just names and suffixes


cmkohnen:
> This is nice, but the `on_confirm_discard_response(_dialog, response, self):` function in ui.py is missing the set_filters. 

Oh, an oversight on my part. I see you already fixed that in https://github.com/Sjoerd1993/Graphs/pull/195 however. thanks!

PR #199, opened 2023-04-11
Small fixes & changes

cmkohnen:
- save data files to /app/share/graphs instead of /app/share/graphs/graphs
- Pass on datadir instead of querying on runtime
- name widgets in dialogs uniformly and sort in order of visual appearance
- use os.path.join instead of string concatination


Sjoerd1993 (approved):
Looks good

PR #200, opened 2023-04-11
Remove path extension

cmkohnen:
Fixes two bugs regarding projects.
Sadly due to sanboxing appending .graphs to any project won't work, so I had to remove this

### DIFF
--- a/src/actions.py
+++ b/src/actions.py
@@ -111,6 +111,7 @@ def save_project_action(_action, _target, self):
     dialog = Gtk.FileDialog()
     dialog.set_filters(
         utilities.create_file_filters([("Graphs Project File", "graphs")]))
+    dialog.set_initial_name("project.graphs")
     dialog.save(self.main_window, None, ui.on_save_project_response, self)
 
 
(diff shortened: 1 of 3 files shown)


Sjoerd1993:
I've tried, but saving project files works fine with the automatic .graphs extension in the latest build? Can you check again if it works? As I don't have any problems with my sandboxed flatpak.

Sjoerd1993:
I've done some testing, and you are correct. The .graphs extension does clash with the sandbox. My guess is that I have opened the permissions using Flatseal at one point when I tried to get everything to work with the Google Drive extension in Nautilus. 

Reopening this now. We can just go with the way this PR does it, and set the suffix by default.
The thing that doesn't work in the Flatpak sandbox is the part with `with open(path, "w") as file:`, while in numpy it is indeed possible to append a suffix to the path. (Actually it does save, but not with the correct filename basically)

I've been trying to get around this for an hour to no avail. For now indeed, it seems best to remove this automatic suffix for now.

Sjoerd1993 (approved):

PR #202, opened 2023-04-12
use set compehensions and remove redundant list operator

cmkohnen:
Just a quick change to use set comprehensions and a fix my local flake8 install suggested (I'm on my laptop currently)


Sjoerd1993 (approved):
Generally cleaner like this

Issue #201, opened 2023-04-11
Rework Clipboard?


cmkohnen:
Whilst testing stuff I noticed how the clipboard currently works:
Whenever a new Item is added (or deleted) the clipboard is reset.
A user might want to undo add or delete actions as well as reordering and this current limitation is intuitive imo.
To accommodate for that would a rework be feasible?
Another approach to implement a clipboard would be to hold a few copies of the current datadict. This would enable to store actions like adding, deleting, reordering and everything we already do.
In Addition to that the item menu would need a different implementation as well, maybe recreating it every reload. Since all information is stored inside the item at this point, this should be easy to implement.

Sjoerd1993:
I agree that the current limitation is unintuitive, and being able to undo and redo actions such as removing and reordering data would be nice.I guess an option is instead of holding a copy of all x- and y data separately on each file, we can have a single datadict_clipboard variable that gets expanded with each action.It would hold a bit more data, but honestly not that much. A single data file can have thousands of data points, a could of extra attributes is not going to have a very significant impact here. We could think about wether we should limit the amount of items in the clipboard as there is always a risk of keeping too much in memory. Then on undo and redo it can just swap out the datadict to the required position, which would be a pretty similar implementation as now. The other advantage (apart from getting rid of the reset) would be that allows to undo deletion of data, which right now is irreversible. (Accidentally adding data is a minor inconvenience, accidentally removing data is not fun)The item menu doesn’t have to be changed a lot to accomodate for this I think, but at least it needs to be reloaded on undo and redo to match the datadict. It’s not really that big of a project to rework it like that. I can have a look at it tomorrow when I’m back in my computer.Skickat från min iPhone11 apr. 2023 kl. 21:31 skrev Christoph Matthias Kohnen ***@***.***>:﻿
Whilst testing stuff I noticed how the clipboard currently works:
Whenever a new Item is added (or deleted) the clipboard is reset.
A user might want to undo add or delete actions as well as reordering and this current limitation is intuitive imo.
To accommodate for that would a rework be feasible?
Another approach to implement a clipboard would be to hold a few copies of the current datadict. This would enable to store actions like adding, deleting, reordering and everything we already do.
In Addition to that the item menu would need a different implementation as well, maybe recreating it every reload. Since all information is stored inside the item at this point, this should be easy to implement.

cmkohnen:
[...]

PR #204, opened 2023-04-12
Rework clipboard

Sjoerd1993:
Reworks the clipboard as discussed in issue #201.

Doesn't (yet) set the datadict as it's own class, but that could be a seperate PR if we were to implement this.
Also reworks the item menu as discussed, as that was needed to get the behaviour right.

Actually, this code is simpler than the previous implementation and works quite well. It correctly handles all actions, including deletions and everything. And you can even redo/undo all the way to an empty slate and back.

### DIFF
--- a/data/config.json
+++ b/data/config.json
@@ -4,6 +4,7 @@
 "addequation_step_size": "0.01",
 "addequation_x_start": "0",
 "addequation_x_stop": "10",
+"clipboard_length": "15",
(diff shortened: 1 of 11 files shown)


Sjoerd1993:
Also added a setting to change the clipboard length. Set it to 15 by default, and gave the spinner a max of 999.

cmkohnen (approved):

PR #206, opened 2023-04-12
Minor bugfixes

Sjoerd1993:
- Fix select all and select none shortkeys, the functions were still trying to set the selected mode through the item_menu variable which no longer exists. This fixes this.

- Set restore view to custom function to find limits. This gives a more consistent experience with loading data and such, but also the built-in "home" functionality has some bugs with specific edge cases that were solved in the custom function.

### DIFF
--- a/src/actions.py
+++ b/src/actions.py
@@ -50,7 +50,7 @@ def add_equation_action(_action, _target, self):
 
 
 def select_all_action(_action, _target, self):
-    for key, item in self.datadict.items():
+    for item in self.datadict.values():
         item.selected = True
     graphs.refresh(self)
     ui.reload_item_menu(self)
@@ -58,7 +58,7 @@ def select_all_action(_action, _target, self):
 
 
 def select_none_action(_action, _target, self):
-    for key, item in self.datadict.items():
+    for item in self.datadict.values():
         item.selected = False
     graphs.refresh(self)
     ui.reload_item_menu(self)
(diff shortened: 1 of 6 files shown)


cmkohnen:
>   - Set restore view to custom function to find limits. This gives a more consistent experience with loading data and such, but also the built-in "home" functionality has some bugs with specific edge cases that were solved in the custom function.

When loading data, the home should already be set to the current limits. Using this function acted up when trying to navigate forward/back in the view stack after that. What are the specific edge cases?
Just for reference the upstream implementation: [link](https://github.com/matplotlib/matplotlib/blob/bff46815c9b6b2300add1ed25f18b3d788b816de/lib/matplotlib/backend_bases.py#L2865) self.set_history_buttons() does nothing in this implementation, maybe that was the cause of the problem back when we used the GTK4 backend

Sjoerd1993:
> When loading data, the home should already be set to the current limits. Using this function acted up when trying to navigate forward/back in the view stack after that. What are the specific edge cases?

The issue arrised when resetting it back using the Home button. At one point I had data going from x = 0 to x = 12, after which I cut everything after x = 10. When pressing the icon to reset the view, it somehow reset everything back to the original state from x = 0 to x = 12. However, I haven't been able to replicate this issue at all, no matter how I try.

What I think happened is that I actually did have data around x = 12, but with a value at y = 0 on a log scale. So those data points were simply not visible.

src/actions.py
```diff
@@ -76,7 +76,6 @@ def redo_action(_action, _target, self):
 def restore_view_action(_action, _target, self):
     self.canvas.dummy_toolbar.home()
 
-
```
cmkohnen, line 79: Missing line here, otherwise this looks good!

PR #213, opened 2023-04-13
Set checkmark for active style in style editor

Sjoerd1993:
- Adds a shortcut button that allows to set the style that is being edited as custom style.

I was a bit unsure about the exact behaviour and what would be the most intuitive, but it seemed useful to have a shortcut here to quickly set the style from the editor as well. In the current implementation it hides the button when "System style" is set as default. So only if "Use Custom Style" is enabled, it shows this to allow for quick changing of the plot style. Pressing the switch sets the style directly to active style. Deselecting it sets  the active style to the system preferred style.

 image

- Sets a checkmark for the currently active style in the "Plot Styles" overview, to quickly indicate which style is currently active.

image

### DIFF
--- a/src/plot_styles.py
+++ b/src/plot_styles.py
@@ -237,21 +237,17 @@ class PlotStylesWindow(Adw.Window):
         edit_page = args[-1]
         if self.leaflet.get_visible_child() == edit_page:
             if self.set_active_style.get_active():
+                self.parent.plot_settings.use_custom_plot_style = True
                 self.parent.plot_settings.custom_plot_style = \
                     self.style["name"]
             else:
-                system_style = "adwaita"
-                if Adw.StyleManager.get_default().get_dark():
-                    system_style += "-dark"
-                self.parent.plot_settings.custom_plot_style = system_style
+                self.parent.plot_settings.use_custom_plot_style = False
             graphs.reload(self.parent)
 
     def load_style(self):
         style = self.style
         self.style_name.set_text(style["name"])
 
-        if not self.parent.plot_settings.use_custom_plot_style:
-            self.set_active_style_row.hide()
         if style["name"] == self.parent.plot_settings.custom_plot_style:
             self.set_active_style.set_active(True)
         else:
--- a/src/plot_styles.py
+++ b/src/plot_styles.py
@@ -133,8 +133,6 @@ class PlotStylesWindow(Adw.Window):
     minor_tick_width = Gtk.Template.Child()
     major_tick_length = Gtk.Template.Child()
     minor_tick_length = Gtk.Template.Child()
-    set_active_style = Gtk.Template.Child()
-    set_active_style_row = Gtk.Template.Child()
     tick_bottom = Gtk.Template.Child()
     tick_left = Gtk.Template.Child()
     tick_top = Gtk.Template.Child()
@@ -212,9 +210,6 @@ class PlotStylesWindow(Adw.Window):
         self.style["name"] = style
         self.load_style()
         self.leaflet.navigate(1)
-        edit_page = self.leaflet.get_visible_child()
-        self.set_active_style.connect("state-set", self.toggle_style,
-                                      edit_page)
         self.set_title(style)
 
     def back(self, _):
@@ -233,26 +228,10 @@ class PlotStylesWindow(Adw.Window):
         self.leaflet.navigate(0)
         self.set_title(self.style["name"])
 
-    def toggle_style(self, *args):
-        edit_page = args[-1]
-        if self.leaflet.get_visible_child() == edit_page:
-            if self.set_active_style.get_active():
-                self.parent.plot_settings.use_custom_plot_style = True
-                self.parent.plot_settings.custom_plot_style = \
-                    self.style["name"]
-            else:
-                self.parent.plot_settings.use_custom_plot_style = False
-            graphs.reload(self.parent)
 
     def load_style(self):
         style = self.style
         self.style_name.set_text(style["name"])
-
-        if style["name"] == self.parent.plot_settings.custom_plot_style and \
-                self.parent.plot_settings.use_custom_plot_style:
-            self.set_active_style.set_active(True)
-        else:
-            self.set_active_style.set_active(False)
 
         # font
         font_description = self.font_chooser.get_font_desc().from_string(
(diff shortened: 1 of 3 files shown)


Sjoerd1993:
Tweaked the switch behaviour a bit more. Now it's more of an "Override system style" button. This should avoid confusion, and makes it easier to switch back to the system theme as well. Also works better in edge cases.

image

src/misc.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, config):
         self.top_scale = config["plot_top_scale"]
         self.title = config["plot_title"]
         self.legend = config["plot_legend"]
-        self.use_custom_plot_style = False
+        self.use_custom_plot_style = config["plot_use_custom_style"]
```
cmkohnen, line 42: Regarding this my idea initially was to use this variable to identify whether or not to deviate from preferences. This had several merits. You could choose to use a custom style in the preferences and every project, which had had this disabled would automatically update. Querying from config directly would mean we'd set it upon instance creation like the other variables and keep it consistent.
While I like both ways I tend towards the latter.

src/plot_styles.py
```diff
@@ -58,9 +58,18 @@ def get_system_preferred_style_path(self):
...
 
 
+def get_current_style(self):
+    if not self.parent.plot_settings.use_custom_plot_style:
+        style = "adwaita"
+        if Adw.StyleManager.get_default().get_dark():
+            style += "-dark"
+    else:
+        style = self.parent.plot_settings.custom_plot_style
+    return style
```
cmkohnen, line 68: We're kinda having duplicate code here. I'd suggest shortening this function to
`return Path(get_preferred_style_path(self)).name`
or using this directly in line 450

src/plot_styles.py
```diff
@@ -201,6 +212,9 @@ def edit_style(self, _, style):
         self.style["name"] = style
         self.load_style()
         self.leaflet.navigate(1)
+        edit_page = self.leaflet.get_visible_child()
+        self.set_active_style.connect("state-set", self.toggle_style,
+                                      edit_page)
```
cmkohnen, line 217: would probably move these two statements in the init function

src/plot_styles.py
```diff
@@ -219,11 +233,27 @@ def back_line_colors(self, _):
...
     def load_style(self):
         style = self.style
-
         self.style_name.set_text(style["name"])
 
+        if style["name"] == self.parent.plot_settings.custom_plot_style and \
+                self.parent.plot_settings.use_custom_plot_style:
+            self.set_active_style.set_active(True)
+        else:
+            self.set_active_style.set_active(False)
```
cmkohnen, line 255: I'd boil this down to
```
active = ...
self.set_active_style.set_active(active)
```

src/plot_styles.py
```diff
@@ -124,6 +133,8 @@ class PlotStylesWindow(Adw.Window):
     minor_tick_width = Gtk.Template.Child()
     major_tick_length = Gtk.Template.Child()
     minor_tick_length = Gtk.Template.Child()
+    set_active_style = Gtk.Template.Child()
```
cmkohnen, line 136: just to keep it consistent, I'd sort these after font_chooser. 8sort in order of visual appearance)

src/plot_styles.py
```diff
@@ -124,6 +133,8 @@ class PlotStylesWindow(Adw.Window):
...
+    set_active_style_row = Gtk.Template.Child()
```
cmkohnen, line 137: Is there a reason to name this row and keep a reference? It is not accessed in code, just the switch.

cmkohnen (commented):
In general I'm not quite sure whether this addition would help avoid confusion. As of now the style editor is kinda for managing and editing the styles only, nothing which is connected to any project or loaded data or preferences. Having the option in both plot settings and the preferences might be confusing enough, and a user might not understand what exactly this new switch might override: plot settings or preferences.

I have kept the legend options out the style for now, as I see the need to wish to toggle it on and off for different scenarios or on a whim regardless of other style aspects.

Regardless of that, I quite like the little indicator. I gives a nice visual clue on which style to edit if the current appearance should be changed quickly

Sjoerd1993:
Regarding the general comment about the override button. I mainly added this to aid my personal workflow. Due to bugfixing workflow and everything, I regularly lose all my custom styles which I also use for actual work. So I often create a new style, and set the properties I want for it. Then I have to back away to plot settings, and then set this new style to apply it and see how it turns out. If I then want to tweak something within that style, I have to then go back to Plot Styles, and edit it further from there. It felt convenient to have a shortcut to quickly set a style from the Style Editor as well. It's not really needed when just changing out styles. But it's mainly convenient when creating or editing new styles, so that you can apply them directly. 

But I can see the argument to keep things neatly isolated with one function only, and activating a style is an inherently different function than editing one. So I'm open to keeping the override button in the freezer right now, and adding a shortcut in case user feedback suggests it's needed. So I'd be happy to hear other opinions on that as well.

Keeping the legend in plot settings would be fine as well. In particular since I think we should add a chooser there as well for the legend position. And that's something that's really dependent on a case-by-case basis and shouldn't be part of the styling. Will get back about the other comments. But if we were to remove the override button, then most of those comments  would be redundant of course.

src/plot_styles.py
Sjoerd1993: I was actually looking for a clean way to get the style, but ended up writing a function instead when I couldn't really figure it out. Now put a variant of this one around line 450 instead.

src/plot_styles.py
```diff
@@ -219,11 +233,27 @@ def back_line_colors(self, _):
...
+    def toggle_style(self, *args):
+        edit_page = args[-1]
+        if self.leaflet.get_visible_child() == edit_page:
+            if self.set_active_style.get_active():
+                self.parent.plot_settings.use_custom_plot_style = True
+                self.parent.plot_settings.custom_plot_style = \
+                    self.style["name"]
+            else:
+                self.parent.plot_settings.use_custom_plot_style = False
+            graphs.reload(self.parent)
```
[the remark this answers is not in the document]
Sjoerd1993: This functionality is removed for now, but the reason was basically to have the plot in the background change "life" when the switch is activated, which just gives a nicer more responsive feeling. Took a bit of effort, (hence the edit_page line), as the switch is also activated when backing out and loading another graph.

PR #220, opened 2023-04-14
Add screenshots

Sjoerd1993:
Adds screenshots directly to the GitHub page. 
I've made some preliminary screenshots, but I'm open to hear what we should highlight in these screenshots. Not sure if we should use all, and whether we should highlight other stuff as well. But here's the list of images I made:

- One screenshot showing plain data, visualize_data.png
- One screenshot showing a generated sine function, sine.png
- One screenshot showing a generated sine function, but in dark mode, sine_dark.png
- One screenshot showing the add equation window, with a cos(X) being added, add_equation.png
- One screenshot showing the generated sine and cosine, sin_cos.png
- One screenshot showing the style editor, style_editor.png
- One screenshot showing data being highlighted with the adjust_data row being folded out, manipulate_data.png

Some of these are probably a bit redundant. Also, to show of dark mode compatibility I previously combined a light and dark mode screenshot like this: 
image

I could do something similar, but it may be more clear to show the dark mode as a whole. 

To remove redundancy, we could perhaps go for:
sine_cos.png
sine_dark.png
add_equation.png
style_editor.png
manipulate_data.png

But perhaps it's useful to showcase it's possible to load arbitrary data as well. Hence the visualize_data.png. So we could use that one instead of sine_cos.png, or use such a dataset on the dark mode screenshot.

But again, I'm open to hear suggestions of what to highlight. Regarding the  transparent borders around the screenshots, that's the default in GNOME window screenshots, and it seems to be used by GNOME projects on Flathub as well. Either way, I do like the shadows behind the windows.


cmkohnen (approved):

cmkohnen:
I think the ones you suggested are good examples of the app and do showcase most major features

PR #222, opened 2023-04-14
Add new "thesis" style

Sjoerd1993:
Add the style that I use for my own thesis. Basically it's without grids,  slightly longer ticks on the inside on all borders and a more "standard" like color cycle. I also use the same style as the basis for academic articles.

Felt to me that a style like this was missing somewhat, and for me personally it's convenient to ship these as well.

image

### DIFF
--- a/data/styles/thesis.mplstyle
+++ b/data/styles/thesis.mplstyle
@@ -1,0 +1,109 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+# adwaita
+
+# font
+font.family: sans-serif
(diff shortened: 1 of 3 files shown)


data/styles/thesis.mplstyle
```diff
@@ -0,0 +1,109 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+# adwaita
```
cmkohnen, line 2: Should be 'thesis'

cmkohnen (approved):

PR #228, opened 2023-04-15
Add dedicated "add style" button

Sjoerd1993:
Instead of having a copy button on each style row, this adds one single "add style" button that pops up a dialog. It then gives a dropdown with a choice which style to base the new one on, and a field for the name of the new style. The name is automatically filled in based on the template the user chooses, but can be changed manually of course .

Other than this, there's not really something I really want in the upcoming release. Except for a warning dialog when deleting styles ("This will delete {style name}, are you sure") and for resetting the styles to the default settings, as those actions are destructive. See PR #229. 

image

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -58,12 +58,11 @@ endif
 
 blueprints = custom_target('blueprints',
   input: files(
-    'ui/add_style.blp',
-    'ui/window.blp',
-    'ui/edit_item.blp',
-    'ui/export_figure.blp',
     'ui/add_data_advanced.blp',
     'ui/add_equation_window.blp',
+    'ui/add_style.blp',
+    'ui/edit_item.blp',
+    'ui/export_figure.blp',
     'ui/help_overlay.blp',
     'ui/item_box.blp',
     'ui/plot_settings.blp',
@@ -72,7 +71,8 @@ blueprints = custom_target('blueprints',
     'ui/rename_window.blp',
     'ui/style_box.blp',
     'ui/style_color_box.blp',
-    'ui/transform_window.blp'
+    'ui/transform_window.blp',
+    'ui/window.blp'
   ),
   output: '.',
   command: [find_program('blueprint-compiler'), 'batch-compile', '@OUTPUT@', '@CURRENT_SOURCE_DIR@', '@INPUT@'],
--- a/data/meson.build
+++ b/data/meson.build
@@ -58,6 +58,7 @@ endif
 
 blueprints = custom_target('blueprints',
   input: files(
+    'ui/add_style.blp',
     'ui/window.blp',
     'ui/edit_item.blp',
     'ui/export_figure.blp',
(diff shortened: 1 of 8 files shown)


Sjoerd1993:
This also cleans up the style list somewhat:

image

src/plot_styles.py
```diff
@@ -483,3 +490,41 @@ def __init__(self, parent, color):
...
+        utilities.populate_chooser(
+            self.plot_style_templates, get_user_styles(parent.parent).keys())
+        selected_item = \
+            self.plot_style_templates.get_selected_item().get_string()
+        self.new_style_name.set_text(f"{selected_item} (copy)")
+        self.cancel_button.connect("clicked", self.on_cancel)
+        self.confirm_button.connect("clicked", self.on_confirm)
+        self.plot_style_templates.connect("notify::selected",
+                                          self.on_template_changed)
+        self.set_transient_for(parent.parent.plot_styles_window)
```
cmkohnen, line 513: parent would suffice here

src/plot_styles.py
```diff
@@ -389,22 +391,29 @@ def delete_style(self, _, style):
...
+                    loop = False
         user_path = os.path.join(utilities.get_config_path(), "styles")
         shutil.copy(
             os.path.join(user_path, f"{style}.mplstyle"),
             os.path.join(user_path, f"{new_style}.mplstyle"))
         self.reload_styles()
 
+    def add_data(self, _):
+        win = AddStyleWindow(self)
+        win.present()
```
cmkohnen, line 414: assignment and present() is redundant here

src/actions.py
```diff
@@ -104,7 +104,7 @@ def export_figure_action(_action, _target, self):
 
 
 def plot_styles_action(_action, _target, self):
-    PlotStylesWindow(self)
+    self.plot_styles_window = PlotStylesWindow(self)
```
cmkohnen, line 107: together with my comment on L513 this would be redundant

data/meson.build
```diff
@@ -58,6 +58,7 @@ endif
 
 blueprints = custom_target('blueprints',
   input: files(
+    'ui/add_style.blp',
```
cmkohnen, line 61: might not be scope of this pr, but perhaps sort this list alphabetically (Probably oversight on my end)

data/ui/style_box.blp
```diff
@@ -19,16 +19,6 @@ template StyleBox : Box {
...
-    tooltip-text: _("Copy Style");
-    styles ["flat"]
-    Image {
-      hexpand: false;
-      icon-name: "edit-copy-symbolic";
-      pixel-size: 20;
-    }
-  }
-
   Button delete_button {
```
cmkohnen, line 22: Perhaps not in scope either, but maybe we should switch the delete and edit button for consistency

Sjoerd1993:
Removed the cancel button and changed the margins to match the rest of the UI (e.g. Add Equation window): 

image

cmkohnen (commented):
I feel this is a good change

cmkohnen (approved):

PR #229, opened 2023-04-15
Add warning dialogs

Sjoerd1993:
Adds warning dialogs if the users wants to delete a style, and if they want to reset the styles to the default values

image

image


data/ui/add_style.blp
cmkohnen, line 1: Probably shouldn't be part of this pr?

Sjoerd1993: You are correct, I based this commit on the previous one, but thought I omitted this file when making this new branch.

src/plot_styles.py
```diff
@@ -197,6 +196,25 @@ def __init__(self, parent):
...
+        dialog = Adw.MessageDialog.new(self,
+                                       heading,
+                                       body)
+        dialog.add_response("cancel", "Cancel")
+        dialog.add_response("reset", "Reset")
+        dialog.set_close_response("cancel")
+        dialog.set_default_response("delete")
+        dialog.set_response_appearance("reset",
+                                       Adw.ResponseAppearance.DESTRUCTIVE)
+        dialog.connect("response", self.on_reset_button_press, self)
```
cmkohnen, line 211: I'd remove the self argument as this is a class function

src/plot_styles.py
```diff
@@ -197,6 +196,25 @@ def __init__(self, parent):
...
+        dialog.present()
+
+    def on_reset_button_press(self, _, response, __):
```
cmkohnen, line 214: In conjunction with the previous comment, the last argument can be removed

src/plot_styles.py
```diff
@@ -43,7 +43,6 @@ def reset_user_styles(self):
     for style, path in get_system_styles(self).items():
         shutil.copy(path, os.path.join(user_path, f"{style}.mplstyle"))
 
-
```
cmkohnen, line 46: Any reason for removing this line?

src/plot_styles.py
Sjoerd1993: Yes, that was unnecessary, probably had it there out of habit. Even led to a unused argument in the relevant function.

src/plot_styles.py
Sjoerd1993: Sharp eye, no. That was a mistake.

cmkohnen (approved):

PR #235, opened 2023-04-17
localization improvements

cmkohnen:
Rely on meson for graphs.pot and respective .po files
Updates can be done with setting up a new builddir (e.g. testbuild) and executing the following:
`meson compile graphs-update-po -C testbuild`
Additional languages can be added, by adding the code to the LINGUAS file and executing above mentioned command.
I also used this to add a file for french translation as requested in https://github.com/Sjoerd1993/Graphs/issues/232 
This covers all UI defined strings, I will address code defined ones later


Sjoerd1993 (approved):
Wow, that's a huge change list in terms of lines of code 😅 

Seems to work fine on rerun, so probably some fluke.

Sjoerd1993:
You are correct, it indeed generates correctly! Only thing I'm still confused about is that your generated files also include the release history, while all of my files skip exactly that. The final lines in my graphs.pot are:

```
#: data/se.sjoerd.Graphs.appdata.xml.in.in:41
msgid "4 - Edit plot stylesheets"
msgstr ""

#: data/se.sjoerd.Graphs.appdata.xml.in.in:45
msgid "5 - Manipulate data using a wide range of operations"
msgstr ""

#: src/main.py:44
#, python-format
msgid "Could not load %s"
msgstr ""
```

cmkohnen:
Huh, that is indeed weird

PR #237, opened 2023-04-18
Reset axes limit less agressively

Sjoerd1993:
This adds the ability to keep the old limits after reloading a graph. In the current main branch, when switching to dark mode for instance, the limits are reset. The same is true when reordering lines to put one in front. This has annoyed me at some occasions.
This adds the optional argument reset_limits to the graphs.reload function (true by default). If this is set to false, the old limits are applied again after reloading.

In this PR, this new function is used to make the canvas resets way less aggressive. This way the old limits are kept when changing styles, line properties, plot settings etc..

### DIFF
--- a/data/se.sjoerd.Graphs.appdata.xml.in.in
+++ b/data/se.sjoerd.Graphs.appdata.xml.in.in
@@ -55,6 +55,10 @@
           <li>Started initial work on localization support</li>
           <li>Refactors under-the-hood</li>
         </ul>
+        <p>Bug fixes:</p>
+        <ul>
+          <li>Fixed a bug where data was not saved when multiple files were exported simultaniously</li>
+        </ul>
       </description>
     </release>
     <release version="1.5.1" date="2023-04-17">
--- a/data/se.sjoerd.Graphs.appdata.xml.in.in
+++ b/data/se.sjoerd.Graphs.appdata.xml.in.in
@@ -47,6 +47,16 @@
     </screenshot>
   </screenshots>
   <releases>
+    <release version="1.5.2" date="2023-04-19">
+      <description>
+        <p>Changes:</p>
+        <ul>
+          <li>Changed the behaviour of the canvas so that the limits are not reset unnecessarily</li>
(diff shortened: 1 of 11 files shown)


cmkohnen (commented):
Good change! I feel however maybe a cleaner solution might be to use the already existing set_limits() function and change it to require a dict (basically the new set_manual) and refactor plotting_tools.find_limits() accordingly

cmkohnen:
Just a heads up:
starting tomorrow, I'll be away for a couple of days and after that write my Abitur exams.
I probably won't have time to contribute in that timespan.
I wish you the best of luck with your thesis as well!

Sjoerd1993:
I've refactored the behaviour of the way it sets the limits quite extensively, was a bit more ambitious than I initially planned but the implementation is a lot cleaner than previously.

Now set_limits directly sets the limits of the axis. The previous set_limits function has been redubbed to restore_view in order to signify what it actually does, restore the view to the initial state. I've also streamlined and simplified these functions overall a bit, and renamed set_axis_limits (which was a poor name from my side tbf) to set_axis_padding in order signify what it does: it adds padding to the axis on all sides.

Finally, I've already added these changes to the change-log in app-data. But that doesn't mean that I'm planning to release a 1.5.2 update today. I just put it there already so that we can just add the change-list as-we-go. So that we don't have to do that all the way in the end. The release date that's there is just set to today, so that I can use the commit in the Flathub-beta branch, as that doesn't build without a valid appdata date. (And it cannot be in the future)

Sjoerd1993:
> Just a heads up: starting tomorrow, I'll be away for a couple of days and after that write my Abitur exams. I probably won't have time to contribute in that timespan. I wish you the best of luck with your thesis as well!

No worries, with release 1.5 out of the way, I feel we're in a state where I don't have to push this stuff to production immediately either way. Also this is all of course just voluntarily hobby-stuff in the end, so of course personal time schedules will get in the way at times.

Good luck, and thanks for the heads up :)

cmkohnen (approved):
LGTM!

PR #238, opened 2023-04-21
small linting fixes

cmkohnen:
Just a few fixes my local flake8 pointed out


Sjoerd1993 (approved):
Looks good, regarding the added exception, at one point we can probably limit the scope of what the eval function is allowed to do here. Although the security risks are quite low for an application running in user-space like this (it doesn't have more permissions than a normal Python shell running in a Flatpak sandbox), it's probably best practice to lock it down a bit either way.

PR #241, opened 2023-04-23
Add xry support

cmkohnen:
Add support for .xry file types as gotten from Leybold X-Ray apparatus.
These are devices we use in my school. As the actual logic isn't that complicated and they are stored in a pretty weird format (ISO-8859-1), I figured, it can't hurt to include this functionality.

Sidenote: somehow the commits from my other pr slipped in here as well, these can be ignored, the relevant part are only a few lines in file_io.py

### DIFF
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -213,13 +213,13 @@ class Canvas(FigureCanvas):
         for direction, used in used_axes.items():
             if used:
                 if direction in ["top", "bottom"]:
-                    limits[f"min_{direction}"], limits[f"max_{direction}"] = \
-                        plotting_tools.find_min_max(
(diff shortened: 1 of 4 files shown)


Sjoerd1993:
Seems like a similar motivation of why I added .xrdml support, which we use at my university.

Seems good, I'm on the train now but will look at it this evening so we can merge it into main.

Sjoerd1993:
Seems like a similar motivation as why I added .xrdml support from Panalytical, which we use at my university.

Seems good, I'm on the train now but will look at it this evening so we can merge it into main.

Sjoerd1993 (approved):

Issue #240, opened 2023-04-23
Allow importing single column data


WiseroOrb:

Sjoerd1993:
I can see that being useful in some cases, I'll take a look at it and add this functionality when I've got time either today or tomorrow.

Sjoerd1993:
This should be possible in the next 1.5.2 release, which I'm planning release relatively soon as it contains an important fix. (Saving multiple files at once is broken in the latest release at the moment)

PR #243, opened 2023-04-23
Add single column support

Sjoerd1993:
Adds initial support for files with only one column. Fixes issue #240


cmkohnen (approved):
This is a way cleaner solution than I've imagined it would be!

PR #247, opened 2023-04-25
translate codebase

cmkohnen:
Initial work towards translating code-defined strings. Some elements are still missing (mainly choosers, on which's output we depend, so we can't just change it on a whim)

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -7,13 +7,13 @@ msgid ""
 msgstr ""
 "Project-Id-Version: graphs\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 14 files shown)


Sjoerd1993 (approved):
Looks good, thanks for the effort on this localization work, at one point it could be worthwhile to see if we can get a slightly more universal way to set these choosers that is not dependent on this exact string comparison. That would also allow us to use a bit better names than the arguments that e.g. matplotlib uses.

Next week I'll take a look at some other stuff, mainly adding linear regression and general fitting, and I'm planning to start on Dutch+Swedish translations. Until then it will be a bit slower from my side.

PR #248, opened 2023-04-29
make choosers translatable

cmkohnen:
Allows choosers to be translatable.
A few technical changes:
- Populating a chooser now creates a new model (removed option for keeping old ones as it wasn't used)
- Choosers need to be defined in code for set_chooser and get_selected_chooser_item (new function) to work properly
- When populating a chooser provide translatable values, querying and setting via afore mentioned values is done via the english strings
- Removed empty_chooser as it wasn't used
- Removed get_chooser_list (substituted via chooser.untranslated_items call)

All text should now be translatable, if not I'll consider it a bug.
Fixes https://github.com/Sjoerd1993/Graphs/issues/232 properly.

### DIFF
--- a/data/ui/add_data_advanced.blp
+++ b/data/ui/add_data_advanced.blp
@@ -39,9 +39,6 @@ template AddAdvancedWindow : Adw.Window {
         Adw.ComboRow separator {
           title: _("Decimal Separator");
           subtitle: _("The character that is used as decimal separator");
-          model: StringList {
-            strings [".", ","]
-          };
         }
 
         Adw.ActionRow {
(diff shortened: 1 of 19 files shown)


Sjoerd1993 (approved):
Apologies for ignoring this PR (and the other ones as well). Last week has been crazy, where I only went home to eat and sleep. Send in my thesis on wednesday, took thursday and friday off and spend the rest of my free time catching up with projects around the house. Not that my work is done now, but I'm in a much more comfortable spot now at least to check these out as well :)

This seems good in general, and works on my machine. I also see you've changed the markers names to more humanly readable wordings which looks much more polished. Looks good!

PR #250, opened 2023-05-06
Bundle whats_new and config.json inside gresource

cmkohnen:
As the title says.
This should yield marginally better performance and reduce file system overhead when launching the application.


Sjoerd1993 (approved):
I'm not sure if performance difference will be noticable in real life applications but it's more streamlined like this indeed.

PR #252, opened 2023-05-07
install to python site packages

cmkohnen:
Install using the python meson module. This defaults to the pythons site_packages directory. This is more inline with what most projects do. This also allows to use the python.bytecompile option once meson 1.2 is available for the flatpak buildchain, improving startuptime app.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -16,7 +16,7 @@ description_short = 'Plot and manipulate data in a breeze'
 
 i18n = import('i18n')
 gnome = import('gnome')
-python = import('python')
+python = import('python').find_installation('python3')
 
 prefix = get_option('prefix')
 datadir = get_option('datadir')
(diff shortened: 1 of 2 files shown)


Sjoerd1993 (approved):
This seems good, and in general I think it's best to follow the overall standards of what most projects in the general ecosystem do. 

The root meson.build changes effectively seem like to do the same thing on compilation right? I think it's a bit cleaner this way so that's a good change either way in my books.

In src `meson.build` file, it does behave a bit differently. Could you elaborate what the main differences are in either method? Just for curiosity.

cmkohnen:
The install is now using the [Meson python](https://mesonbuild.com/Python-module.html) module.
This revolves around the `python_installation` object, so instead of the module `python` is now set to the object. In the main `meson.build` this changes nothing, this is only to access this inside the src/ one. The `install_source` function behaves similarly to the `install_data` function, only that files are by default equipped with the right permissions and put in the `site-packages` directory of the selected python installation (the one we use to run the project either way). Starting in meson 1.2 this can precompile the files to bytecode as well, reducing startuptime of the application, as the interpreter doesn't have to do that on each run.

PR #258, opened 2023-05-09
Add Dutch translation

Sjoerd1993:
As title says, adds a Dutch translation to Graphs. I've added the .po file for Swedish as well (hence the giant changelog), as that's the next language I will do. (Also, the final one I will do myself by hand, as I've run out of languages I'm fluent in then)

### DIFF
--- a/data/se.sjoerd.Graphs.appdata.xml.in.in
+++ b/data/se.sjoerd.Graphs.appdata.xml.in.in
@@ -47,6 +47,14 @@
     </screenshot>
   </screenshots>
   <releases>
+    <release version="1.5.3" date="2023-05-09">
+      <description>
+        <p>Changes:</p>
+        <ul>
+	  <li>Adds Dutch translation for Graphs.</li>
+        </ul>
+      </description>
+    </release>
     <release version="1.5.2" date="2023-05-09">
       <description>
         <p>Changes:</p>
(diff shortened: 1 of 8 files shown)


Sjoerd1993:
We may also want to add the possibility to add credits for the translator in each language. It's fine for me for these languages, but it would be nice for potential contributors to have their name there in the "Translate by" section.

It's quite easy to add in the about dialog, but haven't found a neat way quickly to add this only for translated versions. Using a blank string works code-wise (it gets ignored), but isn't picked up by gettxt as blank string messageid is reserved.

Also, some of the lines from the plot styles are not picked up by gettext. This is an issue in the existing graphs.pot as well... Otherwise, it's quite a smooth process.

cmkohnen (approved):

Sjoerd1993:
Absolutely!I actually looked if there’s a default field for design there in the about window at one point but it doesn’t seem to exist. But perhaps we can look otherwise if we can use a more general field.But either way I agree such a credit doesn’t need to be bound to the language. That would also make that easier code-wise of course.Skickat från min iPhone9 maj 2023 kl. 17:48 skrev Christoph Matthias Kohnen ***@***.***>:﻿
Regarding crediting translators, I feel like crediting everyone regardless of the language they worked on in the about window might be suitable as well?
If possible we should credit Tobias Bernard for his Design input as well.

PR #260, opened 2023-05-09
WIP: style refactors

cmkohnen:
Tackling parts of https://github.com/Sjoerd1993/Graphs/issues/251

### DIFF
--- a/src/file_io.py
+++ b/src/file_io.py
@@ -1,11 +1,18 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+import logging
 import os
 import pickle
 import re
 from gettext import gettext as _
+from pathlib import Path
 from xml.dom import minidom
 
+from gi.repository import GLib
+
 from graphs import utilities
+
+from matplotlib import RcParams, cbook
+from matplotlib.style.core import STYLE_BLACKLIST
 
 import numpy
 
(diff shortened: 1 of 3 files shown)


cmkohnen:
Detailed changelists are for the most part inside the commits

src/plot_styles.py
```diff
@@ -1,46 +1,66 @@
...
 
 from cycler import cycler
 
-from gi.repository import Adw, GLib, Gtk
+from gi.repository import Adw, GLib, Gio, Gtk
 
 from graphs import file_io, graphs, misc, utilities
 
 
+def _styles_in_directory(directory):
```
Sjoerd1993, line 13: Why the _ in the function name itself? The function is used in get_system_styles and get_user_styles.

src/plot_styles.py
```diff
@@ -1,46 +1,66 @@
...
+    return styles
+
+
 def get_system_styles(self):
-    path = os.path.join(self.pkgdatadir, "styles")
-    return {Path(file).stem: os.path.join(path, file) for file
-            in os.listdir(path)
-            if os.path.isfile(os.path.join(path, file))}
+    return _styles_in_directory(
+        Gio.File.new_for_uri("resource:///se/sjoerd/Graphs/styles"))
```
Sjoerd1993, line 31: It is generally recommended to work with the uri actually (so this is good), I've been trying to get that to work for added data as well. This is because just using the path itself may not always work in edge-cases.

The main edge case being when using the nautilus google drive plugin, but as the Flatpak sandbox does not grant access to it, I gave up on this. Think granting that access may ruin the "Safe" label on the Software Center as well, which I would like to keep.

src/file_io.py
```diff
@@ -161,24 +170,79 @@ def get_data(self, import_settings):
...
     return style
 
 
-def write_style(path, style):
-    with open(path, "w", encoding="utf-8") as file:
-        file.write(f"# {style['name']}\n")
-        for key, value in style.items():
-            if key != "name":
-                file.write(f"{key}: {value}\n")
+WRITE_IGNORELIST = [
```
Sjoerd1993, line 219: I'm not 100% sure what this does, I assume it's ignoring these parameters right? 
We're not using these options right now (yet) afaik, so this makes sense.

src/file_io.py
```diff
@@ -1,12 +1,21 @@
...
 from gettext import gettext as _
+from pathlib import Path
 from xml.dom import minidom
 
+from gi.repository import GLib
+
 from graphs import utilities
 
+from matplotlib import RcParams, cbook
+from matplotlib.style.core import STYLE_BLACKLIST
```
Sjoerd1993, line 17: So according to the docs, STYLE_BLACKLIST contains the rcparams that are not updated.
Is this list empty by default? What does it usually contain?

src/utilities.py
```diff
@@ -226,10 +226,9 @@ def check_if_floats(array):
     return True
 
 
-def get_config_path() -> str:
-    if os.getenv("XDG_CONFIG_HOME"):
-        return os.path.join(os.getenv("XDG_CONFIG_HOME"), "Graphs")
-    return os.path.join(str(Path.home()), ".local", "share", "Graphs")
+def get_config_directory():
```
Sjoerd1993, line 229: Interesting with this implementation from GFile. Does this work the same for Flathub environments and for source builds?

While compiling from source is not really supported, it would be nice if it works either way.

Sjoerd1993 (approved):
My comments are generally more curiosities rather than requests for changes, so I've approved it either way.

Seems to be nice to use GFile with the uri instead of the direct path. At one point I tried to do this for the imported files as well (which still use the path). It's the recommended way to do this, and I was hoping it would allow us to add extensions to saved project files (doesn't seem to be the case). 

While my attempts kinda worked in the end actually, it didn't solve what I tried to solve (compatibility with Google Drive extension in Nautilus, and saving project files with an extension). 

But either way it may be good to use the uri there as well instead of the path itself, if only for consistency. 

Currently saving project files works quite well by the way with the pre-filled extension in the save dialog. So adding the extension automatically is not on top of my priority list at the moment. It's quite natural to keep the extension intact, double clicking on the name of the project file to change the template name only highlights the name itself. So during my workflow when working on my Thesis, I never accidentally forgot to add the extension.

cmkohnen:
> Seems to be nice to use GFile with the uri instead of the direct path. At one point I tried to do this for the imported files as well (which still use the path). It's the recommended way to do this, and I was hoping it would allow us to add extensions to saved project files (doesn't seem to be the case).

Using an URI is something we should do only for resource files, where we 100% know the location, where files are going to be. In other cases, we can always retrieve a `GFile` in some other method (e. g. when it gets returned by a `FileDialog`).

A GFile is not exactly a drop-in replacement for the usual process. Python works with [IO classes](https://docs.python.org/3/library/io.html), which contain useful functions like reading from and writing to files (the `open()` method returns such a class). When dealing with files we want to manually read, like styles or later import files, using the `load_bytes()` function of a `GFile` is pretty close to `file.read()`

When writing however we have to differentiate between replacing and creating. Either one giving an `OutputStream` which we can use to write to the file (a bit more complex but overall quite simple).

Some python functions like `json.dump()` like to work with before mentioned io classes. In this case we create a temporary buffer in memory which acts just like any file would in the previous implementation and shuffle the data from the buffer to an output stream using before mentioned methods (e. g. saving a JSON object)

> While my attempts kinda worked in the end actually, it didn't solve what I tried to solve (compatibility with Google Drive extension in Nautilus, and saving project files with an extension).
> 
> But either way it may be good to use the uri there as well instead of the path itself, if only for consistency.
> 
> Currently saving project files works quite well by the way with the pre-filled extension in the save dialog. So adding the extension automatically is not on top of my priority list at the moment. It's quite natural to keep the extension intact, double clicking on the name of the project file to change the template name only highlights the name itself. So during my workflow when working on my Thesis, I never accidentally forgot to add the extension.

Refactoring other functions to work with `GFile` which should allow for both file renaming and reading/saving to file systems like `gvfs` (which is probably used by Google Drive). This was planned for another pr.

src/plot_styles.py
cmkohnen: That is to indicate this function is for this file only. It is a shared element by the functions you mentioned and is just to avoid having duplicate code.

src/file_io.py
cmkohnen: Yes, we are now working with an `RcParams` which includes these by default. Writing them unchained to a file results in errors when trying to use these parameters again.

src/file_io.py
cmkohnen: STYLE_BLACKLIST is a set of parameters like backend, etc. which must not be changed by a style. These have nothing to do with visual appearance. They can be added to rcparams just fine however which is why the upstream file parser has this ignorelist which I'm just referencing here

src/utilities.py
cmkohnen: `GLib.get_user_config_dir()` does pretty much the same as `get_config_path()` used to do (excluding the subdir "Graphs") with the added benefit of working on macOS and windows as far as I can tell. This function simply creates a `GFile` for that path and returns the subdir "Graphs" of that file.

PR #261, opened 2023-05-10
Improve import

cmkohnen:
Use GFiles for importing data, should fixe some bugs and edge cases (like hopefully Google Drive)


Sjoerd1993 (approved):
Generally a cleaner solution either way.

Google Drive Nautilus plugin doesn't work in this build either. But either way it would require opening up the Flatpak Sandbox iirc. The main issue is that it's not really a mounted drive, it's just abstracted away to look like that. Making its behaviour a bit special. So it's not really a filesystem you're working with there.

PR #263, opened 2023-05-12
more refactors

cmkohnen:
- Small fixes
- Port opening and saving projects to GFile
- Remove confirm dialog in plot styles editor in favour of the actual trashcan

Exporting to two-column files will be a bit trickier. We'll have to create a temporary file which can be accessed via "normal" python methods, since numpy only accepts a filename or pathlike instead of a ioclass for files (the latter being the proper way to my knowledge). This file can than be moved to the location the user specified. This should still work with all locations we can get a GFile from. Will put that in the next pr however, just a heads up what is needed for that

### DIFF
--- a/src/plot_styles.py
+++ b/src/plot_styles.py
@@ -164,7 +164,7 @@ class PlotStylesWindow(Adw.Window):
         self.styles = []
         self.style = None
         self.reload_styles()
-        self.reset_button.connect("clicked", self.on_reset_button)
+        self.reset_button.connect("clicked", self.reset_styles)
         self.add_button.connect("clicked", self.add_data)
         self.back_button.connect("clicked", self.back)
         self.connect("close-request", self.on_close)
@@ -210,25 +210,6 @@ class PlotStylesWindow(Adw.Window):
         self.color_boxes = {}
 
         self.present()
-
-    def on_reset_button(self, _button):
-        heading = "Reset to defaults?"
-        body = "Are you sure you want to reset to the default styles?"
-        dialog = Adw.MessageDialog.new(self,
-                                       heading,
-                                       body)
-        dialog.add_response("cancel", _("Cancel"))
-        dialog.add_response("reset", _("Reset"))
-        dialog.set_close_response("cancel")
-        dialog.set_default_response("delete")
-        dialog.set_response_appearance("reset",
-                                       Adw.ResponseAppearance.DESTRUCTIVE)
-        dialog.connect("response", self.on_reset_button_press)
-        dialog.present()
-
-    def on_reset_button_press(self, _, response):
-        if response == "reset":
-            self.reset_styles(self.parent)
 
     def edit_style(self, _, style):
         self.style = get_style(self.parent, style)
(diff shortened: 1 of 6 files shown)


Sjoerd1993:
I've been playing around with this, and it mostly seems to work quite nicely.
Just one thing that doesn't work as intended at least, when I try to delete the style that is currently active, it fails to reload the list itself without that active one:

[Screencast from 2023-05-15 15-25-21.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/4d11b7e0-0abb-4f69-9b42-4fa5db6e3c0a)


> Remove confirm dialog in plot styles editor in favour of the actual trashcan

I did notice that these indeed get moved to the actual trashcan, which honestly is really cool. However, I still think that a confirmation dialog is still valuable. Partly also because there's currently no way for a user to import styles back (except for doing it by hand by moving the files manually). But I am open to keeping that without the remove dialog, I noticed that Nautilus doesn't give a warning about this either. Nautilus does give a toast with an undo button build in, that can maybe be nice in the future. Perhaps for now it could be good to add a toast with `"Stylename" moved to trash` or something like that.
Also for the "Reset to defaults" button, this would be good either way as this option overwrites existing configurations, so this cannot be restored by restoring the trash.

Finally, when troubleshooting the above behaviour a bit, I wanted to see what the old code does. Since we've added the localization support, these warning dialogs are broken. This is because we used _ as unused variables before adding localization, which also is the gettext translatable string character. So when we try to call gettext, it tries to call the _ variable. 

This means that the function  `def delete_style(self, _, style): `for instance needs to be changed to  `def delete_style(self, _button, style):`. I did not fix that in a PR right now, as this PR would remove those functions where this is a problem. So it's probably good to first decide how we will handle these before fixing it.

src/plot_styles.py
```diff
@@ -411,35 +391,20 @@ def save_style(self):
         file = directory.get_child_for_display_name(f"{style.name}.mplstyle")
         file_io.write_style(file, style)
 
-    def delete_color(self, _, color_box):
+    def delete_color(self, _button, color_box):
```
Sjoerd1993, line 394: Indeed these `_` arguments were the one I was talking about leading to issues in the current build. In general we should probably get rid of all of these _ arguments to avoid future problem.

Sjoerd1993 (changes requested):
I wanted to submit the following as separate PR, but it kinda clashes with this one, so I thought I'd just put it in here:

All the way down around line 554, it would be nice if the chooser there is populated. So it would be nice if we could change the populate_chooser line to:

```
        utilities.populate_chooser(self.plot_style_templates,
        	sorted(get_user_styles(parent).keys()), False)
```
        	
Around line 450, the `i += 1`  iterates for every style it checks. So if there's 25 styles, and a new style is created with the same name it will get a (26) prefix. Could you move the `i += 1`  line to after the `if new_style == style_1:` statement?

cmkohnen:
I agree, removing the dialogs entirely isn't the best of options. When resetting styles, every single one is actually sent to the trashcan, which I think we should change for them to be deleted instead with the current dialog.

In the future adding a toast with undo, when deleting a style is probably the way to go but for now reintroducing the dialogs is probably for the better.

Sjoerd1993 (approved):

PR #265, opened 2023-05-16
Rewrite data export.

cmkohnen:
Rewrite data export to work with GFile. Also move logic outside of file_io, since we already distinguish between single item and multiple in the caller function. This no longer uses `numpy.savetxt` since we always expect two-dimensional tuples for values thus reducing a lot of overhead intruduced by mentioned function.

Also refactor some logic.

### DIFF
--- a/src/file_io.py
+++ b/src/file_io.py
@@ -11,6 +11,7 @@ from xml.dom import minidom
 from gi.repository import GLib
 
 from graphs import utilities
+from graphs.item import Item
 
 from matplotlib import RcParams, cbook
 from matplotlib.style.core import STYLE_BLACKLIST
@@ -43,7 +44,7 @@ def read_project(file):
         project["version"]
 
 
-def save_file(file, xdata, ydata):
+def save_item(file, xdata, ydata):
     array = numpy.stack([xdata, ydata], axis=1)
     fmt = "\t".join(["%.12e"] * 2)
     if file.query_exists(None):
@@ -86,7 +87,7 @@ def import_from_xrdml(self, file, _import_settings):
 
     self.plot_settings.xlabel = f"{scan_axis} ({unit})"
     self.plot_settings.ylabel = _("Intensity (cps)")
-    return xdata, ydata
+    return Item(self, xdata, ydata)
 
 
 def import_from_xry(self, file, _import_settings):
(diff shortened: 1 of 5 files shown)


src/file_io.py
```diff
@@ -44,23 +43,17 @@ def read_project(file):
...
-        for item in self.datadict.values():
-            xdata = item.xdata
-            ydata = item.ydata
-            filename = item.name.replace("/", "")
-            array = numpy.stack([xdata, ydata], axis=1)
-            file_path = f"{path}/{filename}.txt"
-            if os.path.exists(file_path):
-                file_path = f"{path}/{filename} (copy).txt"
-            numpy.savetxt(str(file_path), array, delimiter="\t")
+def save_file(file, xdata, ydata):
```
Sjoerd1993: Much neater, nice improvment.

src/ui.py
```diff
@@ -50,34 +52,33 @@ def on_confirm_discard_response(_dialog, response, self):
...
+            for item in self.datadict.values():
+                file = directory.get_child_for_display_name(f"{item.name}.txt")
+                file_io.save_file(file, item.xdata, item.ydata)
         else:
-            path = dialog.save_finish(response).peek_path()
-        file_io.save_file(self, path)
-    except GLib.GError:
-        pass
+            file = dialog.save_finish(response)
+            # return value is not subscriptable, so we loop over one item
```
Sjoerd1993: Not sure which is better, but in principle you can convert the values object to a list with list(self.datadict.values), which is iterable.

The loop can then be replaced by a single item = list(self.datadict.values())[0]

cmkohnen: Ah, i trie `item = set(self.datadict.values())[0]` and still got the non-subscriptable error. This is probably a better solution, as the datadict *might* change between invoking the dialog and accepting it.

Sjoerd1993:
Biggest change in the latest commit seems to be to create and return the item on importing, instead of retreiving the xdata and ydata on import, and then adding that to an item in the gateway function. I'm inclined to agree with that design  :+1:

cmkohnen:
> Biggest change in the latest commit seems to be to create and return the item on importing, instead of retreiving the xdata and ydata on import, and then adding that to an item in the gateway function. I'm inclined to agree with that design +1

Yes, see my commit message. The idea was that maybe sometime a filetype would need to set more properties than just xdata and ydata (e. g. color), so having the item accessible there would remove a lot of refactoring down the line

Sjoerd1993 (approved):

PR #266, opened 2023-05-16
Slightly change adding items.

cmkohnen:
adding items now requires a list of items to add. This allows for deduplication of code, where we are handling multiple items. This also allows for multiple items to imported from one single file (if any filetype supports it in the future).


src/file_import.py
```diff
@@ -48,8 +46,12 @@ def _import_from_file(self, file, import_settings):
...
-        item.name = filename
-    return item
+    items = callback(self, file, import_settings)
+    if not import_settings.name:
+        i = 0
+        for item in items:
+            if i < 1:
+                item.name = filename
+            else:
+                item.name = f"{filename} ({i})"
```
Sjoerd1993, line 56: This will never reach the else condtionaly, as i is never changed upon iterations. i will always be zero.
Also, this logic is already handled in the add_data function.

Sjoerd1993 (changes requested):
Overall seems to be nice, and I like this change in the general architecture, but the (i) suffix for existing filenames in the new code does not really work. See comment. (This is also what triggered the other  PR I pushed when I investigated this a bit)

src/file_import.py
cmkohnen: You're absolutely right. My initial thinking was, if a file would extract multiple items, that they are always added, regardless of what the user configured while handling duplicates. A better solution would be to name the individual items something like "*filename* - Voltage" and "*filename* - Amperage" and handle that inside the respective function.

Removing this is probably for the better

PR #267, opened 2023-05-16
Defusedxml

cmkohnen:
Use [defusedxmls](https://pypi.org/project/defusedxml) minidom implementation. Also restructure python dependencies.


Sjoerd1993:
Bundling the python dependencies into one json is actually the preferred way by Flathub, as long as they're not automatically generated by `flatpak-pip-generator`. (Generating new dependencies would namely overwrite the ones that are put in by hand). The reason they were split were mostly historical reasons. Either way, I'm all for bundling them.

Regarding defusedxml, the main difference with the bundled minidom implementation is mostly security, right? Not sure if this is a big issue in a sandboxed environment that doesn't have access to the network. Meanwhile it does add an extra dependency. On the other hand, an end-user can always open the sandbox in Flatseal.  

I've mainly been looking at [this thread](https://github.com/openmc-dev/openmc/issues/1775), to read up a bit about it, and I'm tempted to err on the side of using it as it does seem to work fine as a drop in replacement.

Sjoerd1993 (approved):

PR #268, opened 2023-05-17
Rework duplication logic

Sjoerd1993:
Rework the logic a bit when new styles of items are created with an existing name.
In the style sheet there was a bug where instead of going from "name (1)" to "name (2)", it went from "name (1)" to "name(1) (2)" etc.. This now makes it go as intended.

The other change is in the add_data logic, where it's basically just a bit leaner going for one less round of iterations.

Furthermore, it fixes a bug in the "overwrite existing items" function, which didn't work because the dictionary is not allowed to change size or keys during iteration.


Sjoerd1993:
I see I based this PR on #266. Either way it's probably best to merge those others first.

cmkohnen (approved):
Looks clean. One thing I noticed but didn't want to change in fear of breaking your PR is that handle_duplicates is reassigned for every item that is added. This should be moved out of that loop. Also I'd use `ignored, delete = [], []` for neatness sake

src/graphs.py
```diff
@@ -19,51 +19,61 @@ def open_project(self, file):
...
+                                utilities.get_all_names(self):
+                            new_name = f"{item.name} ({i})"
+                            break
+                    item.name = new_name
+                elif handle_duplicates == "Ignore duplicates":
+                    ignored.append(item.name)
+                    continue
+                elif handle_duplicates == "Override existing items":
+                    delete.append(item_1.key)
+                    continue
```
cmkohnen, line 60: Just noticed this. When deleting the old item, this continue prevents the new item from being added. Instead of deleting stuff would 
```python
item.key = item_1.key
```
(without continue) work in this case?

PR #270, opened 2023-05-18
Rework label adjustments

cmkohnen:
As suggested in https://github.com/Sjoerd1993/Graphs/issues/251#issuecomment-1551692877, This reworks setting labels automatically. This drops the guess_headers config option in favour of that

### DIFF
--- a/data/config.json
+++ b/data/config.json
@@ -8,7 +8,6 @@
 "export_figure_dpi": "100",
 "export_figure_filetype": "svg",
 "export_figure_transparent": true,
-"guess_headers": true,
 "handle_duplicates": "Auto-rename duplicates",
 "hide_unselected": false,
 "import_column_x": 0,
--- a/data/ui/add_data_advanced.blp
+++ b/data/ui/add_data_advanced.blp
@@ -86,15 +86,6 @@ template AddAdvancedWindow : Adw.Window {
           }
         }
 
-        Adw.ActionRow {
-          title: _("Retrieve Label");
-          subtitle: _("Try and retrieve label from the header in the dataset");
-          activatable-widget: guess_headers;
-          Switch guess_headers {
-            valign: center;
-          }
-        }
-
         Button confirm_button {
           halign: center;
           margin-top: 24;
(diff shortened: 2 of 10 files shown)


Sjoerd1993:
This is very interesting and I've been thinking about this for a bit...

The basic idea is that if the labels are different when importing a second data set, then it's likely different data so then it makes sense to plot it on a different axis. Which I am inclined to agree with. From what I can tell, it also only changes the axis that has changed. Which I like. This would mean for example if you have "Time" on the X-axis, then import a "Voltage vs Time" graph, and then a "Current vs Time" graph, it will plot it on the same time axis, but plot the voltage/current separately. That is super neat, and is what convinced me to follow this proposed structure. (In my initial thought, I was thinking of keeping the old style of always defaulting to bottom left, but again I like the possibilities here). Another thing I really like is how it doesn't change axes when labels were not declared.

Regarding dropping the guess headers option itself. I can also agree with that. A blog post that has changed my thinking about this a bit is [this one](https://ometer.com/preferences.html). And in general, just slapping a toggle on every possible behaviours seems like "giving up" a bit, and is not really the GNOME way anyway. I can also recommend [this read](https://blogs.gnome.org/tbernard/2021/07/13/community-power-4/) from Tobias Bernard. So while I initially introduced the existence of that toggle, I think it's probably best to make it as smart as possible and drop the toggle as this PR suggests.

I also like putting the label parameters as an attribute of the items themselves. We could maybe look into making it a bit more intelligent in that when I change the axis of an item to an unoccupied axis, that it then also takes the labels to those axes as well. Which I feel can be in a different PR.

From what I've been playing around with so far, it all seems to work as intended. I don't have too much time to test a lot more, as I forgot my laptop charger so soon I'll be left with a Windows computer at work. But it all seems to work OK.

Sjoerd1993 (approved):

cmkohnen:
My main idea was like the situation you described. I couldn't think of a situation, where one would like to disable this feature, if it works correctly. If issues arise with it, I'd consider it a bug much like the first blog you linked.

To address a few issues with it atm (like carrying over when items position is changed) I was thinking of reworking how labels are figured out, but haven't come up with anything satisfying enough yet.

PR #271, opened 2023-05-18
Rework import

cmkohnen:
Based on https://github.com/Sjoerd1993/Graphs/pull/270

Import Settings is now created using self, the file and optional name
and import_mode. If no name or mode is provided the default logic wll be
used (filename and suffix-based identification).
Depending on the mode the params attribute will be populated. This is
queried from an extension to preferences.
Each file now gets their own ImportSettings instance to accomodate for
possible changes in the future.


Sjoerd1993 (approved):
Initial testing seems to be good, I can see how this improves the usability of the code base.

PR #272, opened 2023-05-18
Rework xry

cmkohnen:
Based on https://github.com/Sjoerd1993/Graphs/pull/271. Sorry if this keeps stacking, but I had time to work on this and needed the architecture in place for that.

After obtaining a larger sample of xry files, I noticed that the parser we used was only applicable for files with one measurement (item) in them. Also if custom markers inside the file were present the import would fail. This fixes these issues and works for all files I've got on my hand currently.


Sjoerd1993 (approved):
I don't have xry files myself, so I am just blindly assuming that that part works good. 

No worries about stacking on the older PR's, as I too see that sometimes you need to build on existing PR's that are not merged yet :)

cmkohnen:
I'll share one file, that I'm testing mainly. It has all aspects I encountered with this filetype:
https://kohnen-rodenberg.de/nextcloud/index.php/s/X2TWwkDfTixEynB

PR #280, opened 2023-06-01
Improve dutch translation

Sjoerd1993:
Based on PR #279 (now in main), with the latest improvements in localization, a bunch of older strings that were not detected as translatable now function properly. These adds those missing strings in the Dutch translation so that the entire application is translated.


po/nl.mo
cmkohnen, line 1: Probably got in here by mistake?

PR #282, opened 2023-06-01
Make PreferencePage titles translatable

sstendahl:
The PreferencePage titles were marked not translatable, leaving those categories still on the english "General" and "Plot" titles. As far as I could see, these were the last remaining english parts in the software itself.


Sjoerd1993:
Oh no, I messed up when resolving my conflicts and pushed this PR to main directly. That's not good. 

For full transparency, this commit contains all changes: https://github.com/Sjoerd1993/Graphs/commit/d2522e566809b35d3923474804af3115aefdea2a

The Turkish string for "General" is used earlier, and I am confident it is correct. For the "Plot" string, I checked what @sabriunal used in the other translations, and checked this against ChatGPT (as a language model, it's actually pretty good with context-aware language usage). It is not a good strategy for a full translation, but I "think" it should be OK in this case.

My apologies for pushing it directly :facepalm:

cmkohnen:
Mistakes can happen. Nothing to worry about.
The changes look good to me.
Since this pr now includes only a merge commit I'd drop it

Sjoerd1993:
Yeah there's no point in merging an empty PR, but I only noticed the mistake after opening the PR. Kept it up anyway to at least make the commit changes a bit more visible.

PR #283, opened 2023-06-01
Fix advanced import + headers

Sjoerd1993:
Advanced import did not work because it loaded the column values into the GUI, instead of getting it from the GUI.
Also the headers were broken, as it was looking for an object attribute while the params are stored as dict.


Sjoerd1993:
On a related note, I don't think `add_data_advanced.py` is ever called now at this point since last refactor. We maybe could remove it then.

cmkohnen:
Yes, I thought I deleted that file

PR #284, opened 2023-06-01
Make limits persistent.

cmkohnen:
Refactor code regarding axis limits.
Axis limits are now stored inside the plot settings, making them persistent when opening and saving projects.

### DIFF
--- a/src/actions.py
+++ b/src/actions.py
@@ -78,6 +78,7 @@ def redo_action(_action, _target, self):
 
 def restore_view_action(_action, _target, self):
     self.canvas.load_limits()
+    graphs.refresh(self)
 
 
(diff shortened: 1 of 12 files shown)


Sjoerd1993:
This is very good at least for my personal workflow. As I often create graphs, specifically to have them in presentations and such. Having the axis limits persistent makes this much more usuable, and has been somewhere on my list as well!

Sjoerd1993 (changes requested):
Apart from the "restore view" thing, I haven't found any bugs so far with regard to the refactored limit optimization. So if we fix that, I think it's good for merging

Sjoerd1993 (approved):
Works well on my side now!

PR #285, opened 2023-06-02
Reload less agressively.

cmkohnen:
Reload the graph only when styles changed. Instead use refresh.

### DIFF
--- a/src/item_box.py
+++ b/src/item_box.py
@@ -49,7 +49,7 @@ class ItemBox(Gtk.Box):
             self.parent.datadict, drop_target.key, value)
         ui.reload_item_menu(self.parent)
         clipboard.add(self.parent)
-        graphs.reload(self.parent)
+        graphs.refresh(self.parent)
 
     def on_dnd_prepare(self, drag_source, x, y):
         snapshot = Gtk.Snapshot.new()
--- a/src/rename.py
+++ b/src/rename.py
@@ -3,8 +3,6 @@ from gettext import gettext as _
 
 from gi.repository import Adw, Gtk
 
-from graphs import graphs
-
 
 @Gtk.Template(resource_path="/se/sjoerd/Graphs/ui/rename_window.ui")
 class RenameWindow(Adw.Window):
@@ -46,5 +44,5 @@ class RenameWindow(Adw.Window):
             parent.plot_settings.right_label = text
         if item == parent.canvas.title:
             parent.plot_settings.title = text
-        graphs.reload(parent)
+        parent.canvas.set_axis_properties()
         self.destroy()
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -4,7 +4,7 @@ from contextlib import nullcontext
 
 from gi.repository import Gtk
 
-from graphs import file_io, plot_styles, utilities
+from graphs import utilities
 from graphs.item import Item, TextItem
 from graphs.rename import RenameWindow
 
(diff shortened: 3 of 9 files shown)


src/rename.py
```diff
@@ -46,5 +44,5 @@ def on_accept(self, _widget, parent, item):
             parent.plot_settings.right_label = text
         if item == parent.canvas.title:
             parent.plot_settings.title = text
-        graphs.reload(parent)
+        parent.canvas.set_axis_properties()
```
Sjoerd1993, line 47: The actual titles are not renamed until the graph is completely reloaded. So at the moment changing the graph titles does not work in this PR.

Changing the title, then resizing the window works, the problem lies in the lack of reloading..

src/item_box.py
```diff
@@ -49,7 +49,7 @@ def on_dnd_drop(self, drop_target, value, _x, _y):
             self.parent.datadict, drop_target.key, value)
         ui.reload_item_menu(self.parent)
         clipboard.add(self.parent)
-        graphs.reload(self.parent)
+        graphs.refresh(self.parent)
```
Sjoerd1993, line 52: I'm not sure if this line is the exact culprit, but when a file is loaded that has a header, it does not actually load the new label directly. This is because the graph needs to be re-rendered before text items are changed. So you need to e.g. resize the window first before that effect is taken into place.

So at least when a title/axis label is changed, it should probably reload the graph completely.

Sjoerd1993 (changes requested):
See comments about changing of axis labels not working. Other than that it seems to work good.

Sjoerd1993:
On a side-note, right now the "Restore View" restores to the plot_settings values. Meaning that if you set certain axis limits, zoom in, and then press the "Restore View" button, it does not restore the view  to the full size but rather to whatever was set previously. See attached screencast.

Is this behaviour on purpose? I only really noticed it until now. If not intentional, I think we can keep it like that as it may actually be useful to not re-set the user defined scales all the time. Sometimes, I want to just quickly zoom in on a part of the graph, without ruining my carefully chosen limits for the plot that I want to use in e.g. a presentation. We can maybe then add a "Reset view" or something similar to directly access the "optimize_view" function in the view's side-menu. 

[Schermopname van 2023-06-02 13-30-40.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/e9f09e34-44b1-4a12-952c-e37f02ef7445)

cmkohnen:
The labels are now queried every time the graph is refreshed. Also the legend is now updated only once instead of for each item.

Regarding the "Restore View" action, I feel like we always had different ideas of what this actually should do.
My first understanding was that it should feel like undo/redo and navigating home would lead to the very first view (or any that has been set, when adding items). This is very much akin to what matplotlib does by default with the toolbar.
I felt like having the action coupled to the limits in plot settings would feel more intuitive since the user went the extra steps to manually specify them. 
However I feel like a better solution here would be to have a set of back and forth buttons coupled to the view. This would mean that we'd have to implement a clipboard so to say for the views as well and hook into matplotlibs implementation and save zooming and panning states. In this regard I'd drop the navigate home button.

Either case, I'd definetely implement a "Optimize Limits" or something similarly worded

Sjoerd1993:
Yeah my point was that I think it was a good change, although an "optimize limits" option would be nice as it can happen that  you "lose" a data set if you set your limits out of range. Also, it's nice to have the quick option if you just want to see the data.

My initial thinking with the "Restore View" button was simply to emulate the Home button on Matplotlib's implementation, except that the "Restore View" button is a bit more fine-tuned. It handles multiple axes better, and works better with log-scaling in certain situations. However, since we've added the functionality in Graphs to set the limits manually in Plot Settings, the behaviour started to diverge a bit, since suddenly the default "Restore View" behaviour overrides the manually set values. You are correct that this "new" implementation (where it restores to the manually set values) is much more in line with Matplotlib's implementation and generally the preferred behaviour (which is also why I originally said that I like this implementation). 

Regarding back and forth options. As far as I know, this is basically what the currently available "Next View" and "Previous View" options do (Alt+Z and Shift+alt+Z). Optimizations and further tweaking is always nice, but either way I just wanted to say that it's a nice improvement in behaviour :)

PR #288, opened 2023-06-07
Fix crash with chooser change

Sjoerd1993:
In the current build (as well as with the latest stable Flatpak build), the Edit Item window crashes when the name of an item is changed, and then a new item is loaded.

This bug is triggered somewhere on the repopulation on the chooser, but I have not been able to pin down the exact cause of this. As I'm pressed on time, I decided to put this current workaround (which is to use the old implementation) here as a WIP PR, but I think it's probably for the better if we try and fix the current repopulation function instead.

Not sure when I have time to look closely into that behaviour, but at the latest early next week (PhD defense is on Friday, that is my last big deadline for a while).

There are a few bugs that got introduced when we added localization support. Most of them are fixed in the latest builds, but this is one of them. I noticed another small issue with the dots in the menu not changing colours when loading new styles. Next week I will look closely exclusively at fixing bugs. Then we may can release a new build so those fixes make it into stable.

### DIFF
--- a/src/edit_item.py
+++ b/src/edit_item.py
@@ -49,8 +49,10 @@ class EditItemWindow(Adw.PreferencesWindow):
 
         # If item_selector no longer matches with name, repopulate it
         names = utilities.get_all_names(self.parent)
-        if set(names) != set(self.item_selector.untranslated_items):
-            utilities.populate_chooser(self.item_selector, names, False)
+        if set(names) != \
+                set(utilities.get_chooser_list(self.item_selector)):
+            utilities.repopulate_chooser_from_list(
+                self.item_selector, names)
             self.item_selector.set_selected(index)
             self.parent.datadict_clipboard = \
                 self.parent.datadict_clipboard[:-1]
--- a/src/edit_item.py
+++ b/src/edit_item.py
@@ -47,11 +47,9 @@ class EditItemWindow(Adw.PreferencesWindow):
         self.load_values()
 
         # If item_selector no longer matches with name, repopulate it
-        names = utilities.get_all_names(self.parent)
-        if set(names) != \
-                set(utilities.get_chooser_list(self.item_selector)):
-            utilities.repopulate_chooser_from_list(
-                self.item_selector, names)
+        names = utilities.get_all_names(self.props.application)
+        if set(names) != set(self.item_selector.untranslated_items):
+            utilities.populate_chooser(self.item_selector, names, False)
             self.item_selector.set_selected(index)
             self.props.application.datadict_clipboard = \
                 self.props.application.datadict_clipboard[:-1]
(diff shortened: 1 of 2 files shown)


src/edit_item.py
```diff
@@ -24,14 +24,14 @@ def __init__(self, application, item):
...
         names = utilities.get_all_names(self.props.application)
-        utilities.populate_chooser(self.item_selector, names)
-        self.item_selector.set_selected(names.index(self.item.name))
-
         utilities.populate_chooser(self.plot_x_position, misc.X_POSITIONS)
         utilities.populate_chooser(self.plot_y_position, misc.Y_POSITIONS)
         utilities.populate_chooser(self.linestyle, misc.LINESTYLES)
         utilities.populate_chooser(self.markers, sorted(misc.MARKERS.keys()))
         self.load_values()
+        utilities.populate_chooser(self.item_selector, names)
```
Sjoerd1993, line 32: When this chooser is populated (so when the new model is set), it triggers "Apply", where the item name is set to the entry. The reason behind this, is so that changes are applied when switching to a different item. However, as the values were not loaded yet, the entry is blank and the item name is thus reset to a blank value.

So at init, this populate_chooser needs to be run after load_values.

src/utilities.py
```diff
@@ -86,8 +86,14 @@ def set_chooser(chooser, choice):
 
 def populate_chooser(chooser, chooser_list, translate=True):
     """Fill the dropdown menu with the strings in a chooser_list"""
+    if chooser.get_model():
+        model = chooser.get_model()
+        for _item in model:
+            model.remove(0)
+    else:
```
Sjoerd1993, line 93: It somehow crashes when setting a new model to a chooser that already has an existing model. This was the main cause of the bug that this PR has been trying to solve. I haven't found a way to do this in a more pretty way, but this is quite simple nevertheless: If the model already exist, just empty it instead of creating a new one.

src/edit_item.py
```diff
@@ -94,7 +94,7 @@ def apply(self):
         clipboard.add(self.props.application)
         graphs.refresh(self.props.application)
         if set_limits:
-            plotting_tools.optimize_limits(self)
+            plotting_tools.optimize_limits(self.props.application)
```
Sjoerd1993, line 97: This was just missed when moving from the seperate parent to this build in properties.

Sjoerd1993:
This is ready now, see attached comments for reasoning behind each change.

src/edit_item.py
cmkohnen: Yes, I noticed that after changing signal handlers to be in blueprint they are already connected during init, which caused the window to not show up at all, was working on the same fix myself just now

cmkohnen (approved):

cmkohnen:
snap fails on outdated blueprint, merging this would fix that

PR #292, opened 2023-06-08
Use actual system style.

cmkohnen:
When querying for the system preferred style, return the shipped version. This also includes removing the checkmark for the used style if no custom style is set and ensures the style is always available and has the correct values.
Also rename the functions as we are no longer returning a path bur rather a GFile represantation.

I feel this needs a bit of reasoning: In my opinion following the system style should always look like it fits in regardless of what the user experiments or changes in the adwaita styles. If something should be changed, this is either an issue with our style or the user wants to deviate for a specific reason. In the latter case that would mean to set the style to custom, bearing in mind that this is no longer default.
This way we ensure that the default experience always looks like tested and on the screenshots even after version changes, which might include changes to the default styles.

### DIFF
--- a/src/graphs.py
+++ b/src/graphs.py
@@ -115,7 +115,7 @@ def check_open_data(self):
 def reload(self):
     """Completely reload the plot of the graph"""
     pyplot.rcParams.update(
-        file_io.parse_style(plot_styles.get_preferred_style_path(self)))
+        file_io.parse_style(plot_styles.get_preferred_style(self)))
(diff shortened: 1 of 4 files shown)


Sjoerd1993 (approved):
I absolutely agree with this behaviour, the system style should be exactly as described: the default style that blends in with the GNOME UI. 

Custom style sheets should be supported of course, but are exactly that: custom style sheets. This works really well, and is frankly quite intuitive as well.

PR #294, opened 2023-06-08
Snap fixes and added a workflow to test it

soumyaDghosh:


.github/workflows/snapcraft.yml
cmkohnen, line 1: Would it be possible to add this to the main workflow?
This would mean everything is neatly in one place, as we would want to trigger under the same circumstances

Sjoerd1993: I agree that a single main workflow would be cleaner.

Also the Snap workflow would need more concise name that looks more like the Flatpak ones maybe. 

Finally, which is more of an open question, should we build Snap for Arm64 here as well, like we do for the Flatpak?

soumyaDghosh: We may not even need nothing, as this repo will be connected in snapcraft and we can build there in all arches we can. I would let you guys think. We actually need no workflow to upload the snap to the store.

cmkohnen: Yes the workflows are generally for verifying that no breaking changes are introduced. This way they can be caught before they get merged into main. I assume external checks are only targeting the main branch (e. g. when stuff is already merged) and potential regressions would have to be fixed after (which can get messy). So the idea is that we do test every change not only against flatpak but snap as well to avoid any unnecessary breakage,

Sjoerd1993: Exactly. Strictly speaking we don't *need* any Flatpak workflow either, this is actually done (more elaborately even with a full proper build) before pushing a build to Flathub as well.

But the idea with the workflow is to catch unnecessary breaks right at the PR _before_ merging to main. So adding a Snap workflow definitely has my preference.

soumyaDghosh:
Hi.. I am recreting this PR, so, closing this for now

PR #295, opened 2023-06-08
Build dialogs dynamically.

cmkohnen:
Dialogs, especially Adw.MessageDialogs are now stored in dialogs.blp and can be build using a new function in ui.

This requires a version bump in the blueprint-compiler, requiring slight syntax adjustments

### DIFF
--- a/blueprint-compiler.json
+++ b/blueprint-compiler.json
@@ -5,8 +5,8 @@
     {
       "type": "git",
       "url": "https://gitlab.gnome.org/jwestman/blueprint-compiler",
-      "commit": "6ac798ea6ff17aec17c9be1064572eda2b67db6c",
-      "tag": "0.8.1"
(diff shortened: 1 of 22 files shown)


Sjoerd1993 (commented):
There's been some nice QoL changes lately from your side. (At least code-wise this makes our lives easier I mean)

Won't have time to properly give it a spin until (hopefully) tomorrow. But at first glance it looks good. Had to dig a bit into the blueprint documentation, but finally found information about the $-prefix. Which indicate the type callback is provided by the application code.

Any specific reason for using bind-property instead of bind? Is this related to the new Blueprint update?

cmkohnen:
After upgrading, the compiler threw a bunch of warnings, regarding that the syntax changed since 8.0. This includes having a $-character as well as renaming the bind keyword to bind-property to be more in line with xml markup I guess.

Sjoerd1993 (approved):

PR #297, opened 2023-06-10
Connect from blueprint

cmkohnen:
Based on https://github.com/Sjoerd1993/Graphs/pull/295

Move signal handlers as well as a few other things to blueprint.
Moving Signal handlers helps remove the need to reference a widget by
name in code in most cases. This makes the code a bit cleaner in this
regard.
In order to accomplish that, a few light refactors have to be done. This
way the code is more consistent.

### DIFF
--- a/blueprint-compiler.json
+++ b/blueprint-compiler.json
@@ -5,8 +5,8 @@
     {
       "type": "git",
       "url": "https://gitlab.gnome.org/jwestman/blueprint-compiler",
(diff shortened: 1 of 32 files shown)


Sjoerd1993 (approved):
This move indeed does add a bit of changes to the code, but I agree it's cleaner and makes everything more maintainable if it's consistently in the blueprint-space.

PR #299, opened 2023-06-11
Use inbuilt parent

cmkohnen:
Based on https://github.com/Sjoerd1993/Graphs/pull/297 to avoid conflicts (sorry for stacking again)

Gtk Windows have an application property, that can be set during init or
later using either `props.application` directly or the
`set_application()` method. This has a few benefits in the background
(e. g. the application keeping track of all associated windows).


Sjoerd1993 (approved):
In most cases it doesn't change that much (despite the line-by-line changelist), but overall this is probably more proper.

PR #303, opened 2023-06-12
Added `blueprint-compiler` `v0.8.1` and added snap badge in the `README.md`

### LINKED ISSUE #301: Snap package requires Blueprint 0.8.1
With the latest build we've moved to blueprint 0.8.1, which introduces some breaking changes in the syntax. The latest version in the Ubuntu repositories is still from 0.6.x, and therefore fails to build right now. I am unsure if they'll upgrade before Ubuntu 23.10. I don't see any PPA available either, so it seems the snap manifest needs to build blueprint from source.

I think that should be possible, but I don't really know how to do this using snapcraft. Maybe @soumyaDghosh has some ideas here? Basically, building blueprint-compiler needs to be build from the source like we do in the Flatpak [here](https://github.com/Sjoerd1993/Graphs/blob/main/blueprint-compiler.json). It's maybe good anyway to just set these (the Flatpak and the Snap) to build from the same commit anyway, as to make sure both versions are more or less identical.

soumyaDghosh:
Fixed #301

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
 <a href='https://flathub.org/apps/details/se.sjoerd.Graphs'><img width='180' alt='Download on Flathub' src='https://flathub.org/assets/badges/flathub-badge-en.svg'/></a>
-
+<p><a href="https://snapcraft.io/graphs">
+  <img width='180' alt="Get it from the Snap Store" src="https://snapcraft.io/static/images/badges/en/snap-store-black.svg" />
+</a>
+</p>
 # Graphs
 
 ![image](https://raw.githubusercontent.com/Sjoerd1993/Graphs/main/data/screenshots/sin_cos.png)
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -17,7 +17,16 @@ slots:
     name: se.sjoerd.Graphs
 
 parts:
+  blueprint-compiler:
+    source: https://gitlab.gnome.org/jwestman/blueprint-compiler.git
+    source-tag: 'v0.8.1'
+    plugin: meson
+    meson-parameters:
+      - --prefix=/usr
+    override-prime: |
+      echo 'Skip'
   graphs:
+    after: [ blueprint-compiler ]
     # See 'snapcraft plugins'
     plugin: meson
     source: https://github.com/Sjoerd1993/Graphs.git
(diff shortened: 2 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -17,16 +17,26 @@ slots:
...
 
 parts:
+  blueprint-compiler:
+    source: https://gitlab.gnome.org/jwestman/blueprint-compiler.git
+    source-tag: 'v0.8.1'
+    plugin: meson
+    meson-parameters:
+      - --prefix=/usr
+    override-prime: |
+      echo 'Skip'
```
Sjoerd1993, line 27: I was just curious what the override-prime does?
It seems like it just prints a "Skip" in this step, so what was the purpose of these two lines? (line 26 and line 27)

Sjoerd1993 (commented):
I was just curious what the override-prime step does in the blueprint compiler, but in principle it looks good.

snap/snapcraft.yaml
soumyaDghosh: Snap has four steps to build, 
1. Pull
2. Build
3. Stage
4. Prime

Pull: Pulls all the repos and listed repo packages and saves the repo packages in `$CRAFT_STAGE` folder
Build: Builds it
Stage: Stages the build from every part to the `$CRAFT_STAGE` path
Prime: Copies those files which are said to be primed, into the `$CRAFT_PRIME` folder

Now, here, I am saying the builder to not prime the files of `blueprint-compiler` part. So, I am overriding the prime, and just making an arbitrary command to run.

README.md
```diff
@@ -1,5 +1,8 @@
 <a href='https://flathub.org/apps/details/se.sjoerd.Graphs'><img width='180' alt='Download on Flathub' src='https://flathub.org/assets/badges/flathub-badge-en.svg'/></a>
-
+<p><a href="https://snapcraft.io/graphs">
+  <img width='180' alt="Get it from the Snap Store" src="https://snapcraft.io/static/images/badges/en/snap-store-black.svg" />
+</a>
+</p>
```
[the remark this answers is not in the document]
soumyaDghosh: Thank you.

snap/snapcraft.yaml
Sjoerd1993: Thank you for  the explanation, I am new to Snap packaging so I was just trying to understand what was happening. This helps!

Sjoerd1993 (approved):

snap/snapcraft.yaml
soumyaDghosh: Sure, Ask me anytime related to snaps. Thanks.

PR #305, opened 2023-06-12
Refactor export_figure.py

cmkohnen:
- Use inbuilt application property instead of setting parent
- Move accept logic. This way we don't have to pass around a lot of values.


Sjoerd1993 (approved):
Considering on_figure_save_response was specific to this logic either way, this is indeed a lot cleaner without this many arguments.

Especially once we expand some more functionalities that will pay off. Testing on my side also worked very well, no issues.

Issue #298, opened 2023-06-10
Dialogs don't follow HIG


cmkohnen:
most action dialogs (I'd categorize them as such) do not follow the HIG.
The relevant article [here](https://developer.gnome.org/hig/patterns/feedback/dialogs.html#action-dialogs) describes, how they should be layout.
Most notably the accept button in the top right corner and the discard/cancel button in the top left one.
The pattern found the most in graphs atm is the accept buton all the way at the bottom of each dialog and a regular close (x) button in the top right corner.

The solution would be to somehow hide the x of each window, move the button and add a simple cancel button to each.

Sjoerd1993:
I've been looking around a bit, and indeed there's a bunch of dialogs that I would say fall in this category and do not follow this guideline. Just quickly, this would at least include:

- Add data (advanced)
- Transform dialog
- Add data from equation
- Add new style

I won't say these dialogs are in a bad shape, but I do think we should try and adhere to follow the HIG, and therefore adapt them to this description.. The confirmation dialogs (when deleting a style for instance) should be OK, although some/most of them can be replaced by a Toast with an undo button at some point.

PR #306, opened 2023-06-12
redesign action dialogs

cmkohnen:
fix https://github.com/Sjoerd1993/Graphs/issues/298

### DIFF
--- a/data/ui/add_equation_window.blp
+++ b/data/ui/add_equation_window.blp
@@ -5,6 +5,7 @@ template $AddEquationWindow : Adw.Window {
   modal: true;
   default-width: 450;
   title: _("Add Equation");
+  focus-widget: confirm_button;
 
   ShortcutController {
     Shortcut {
@@ -24,7 +25,7 @@ template $AddEquationWindow : Adw.Window {
       }
 
       [end]
-      Button {
+      Button confirm_button {
         label: _("Add");
         styles ["suggested-action"]
         clicked => $on_accept();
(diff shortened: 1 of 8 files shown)


Sjoerd1993:
I think it's nice in general. Some broad things:

In the "Add data (advanced)" dialog, the "Cancel button" still triggers apply. This may be the case for the others as well.
[Screencast from 2023-06-12 12-22-51.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/c8424623-f2d1-443c-ae51-b16abffe98f4)

Add new style maybe looks a bit nicer with equal padding on the bottom as the top:
image

Now the "Rename" button is in the title bar, maybe we can remove the title and subtitle in the rename windows? It may be obvious from the big "Rename" button what it does. Removing those strings maybe looks cleaner.
image

That's all comments I had.

cmkohnen:
Regarding import due to cancelling, this was because I had the logic in place, that when the advanced import is dismissed, it should behave regularily, but with cancel this should no longer be the case.

Sjoerd1993:
I can see how that bug appears there indeed. All working great now.

I feel like we can finetune the margins everywhere some more, so that it is consistent between the dialogs. As well as the rename dialog could be about 100 px more narrow or so. But I think we can do that in a seperate PR, as the margins here are pretty much identical to what we had before this PR. And this PR solves exactly what it is targeting (issue #298).

Sjoerd1993 (approved):

PR #307, opened 2023-06-12
Add language options

Sjoerd1993:
Adds a chooser to switch languages, regardless of system default. Requires a restart (of the application itself), but works as intended, even from Gnome Builder.

### DIFF
--- a/data/config.json
+++ b/data/config.json
@@ -10,6 +10,7 @@
   "export_figure_transparent": true,
   "handle_duplicates": "Auto-rename duplicates",
   "hide_unselected": false,
+  "language": "System default",
   "override_style_change": true,
   "plot_custom_style": "adwaita",
   "plot_legend": true,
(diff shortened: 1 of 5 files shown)


cmkohnen:
Kind of a general thing: I personally feel, that (on release builds) this provides too much clutter in the settings (though be it just one more setting). If a user would want their app to be localized to a specific language, their system is probably already set to that language. I do not know of any GNOME app, that provides such functionality.

I could definitely see this as a "hidden" option, as in only in the config file or only exposed graphically, when in debug builds however.

cmkohnen:
> If one wants to set the language of a single application, this is possible using the environmental variables in Flatseal: [flatpak/flatpak#1504](https://github.com/flatpak/flatpak/issues/1504). (This worked OK for me, with `flatpak run --env=LC_ALL=nl_NL se.sjoerd.Graphs`. 

I wasn't aware of this easy override actually. Then an inbuilt option is actually kind of redundant.

PR #308, opened 2023-06-12
Add Swedish translation

Sjoerd1993:
Most relevant strings have been translated, but it's time to go home now so will continue on this later.
Also need to check if everything works fine with label sizes and everything.


cmkohnen (approved):
LGTM

PR #310, opened 2023-06-12
use inbuilt appid

cmkohnen:
Similar to the pr using the inbuilt application reference.

Also remove deprecated use of FLAGS_NONE


Sjoerd1993 (approved):
lgtm

PR #311, opened 2023-06-13
Refactor Config.

cmkohnen:
- Streamline open and save process
- Check import options individually
- reference only preferences
- rename all references from config to preferences for consistency

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -128,24 +128,22 @@ class GraphsApplication(Adw.Application):
         We raise the application"s main window, creating it if
         necessary.
         """
-        win = self.props.active_window
-        if not win:
-            win = GraphsWindow(application=self)
-        self.main_window = win
-        win.set_title(self.name)
+        self.main_window = self.props.active_window
+        if not self.main_window:
+            self.main_window = GraphsWindow(application=self)
+        self.main_window.set_title(self.name)
         if "(Development)" in self.name:
-            win.add_css_class("devel")
-        config = self.preferences.config
-        self.plot_settings = PlotSettings(config)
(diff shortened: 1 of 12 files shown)


src/main.py
```diff
@@ -128,24 +128,22 @@ def do_activate(self):
...
-        win.toast_overlay.set_child(self.canvas)
-        win.redo_button.set_sensitive(False)
-        win.undo_button.set_sensitive(False)
+        self.main_window.toast_overlay.set_child(self.canvas)
+        self.main_window.redo_button.set_sensitive(False)
+        self.main_window.undo_button.set_sensitive(False)
         ui.enable_data_dependent_buttons(self)
         self.set_mode(None, None, InteractionMode.PAN)
-        win.present()
+        self.main_window.present()
```
Sjoerd1993, line 146: I didn't really notice before that we assign self.main_window, and then use win anyway. Giving us three different variables all refering to the same object in memory (`self.main_window`, `self.props.active_window` and `win`). 

I'm pretty sure that it's functionally identical, as they all point to the same object in memory, it's not exactly clean. So good change.

src/preferences.py
```diff
@@ -7,50 +7,65 @@
 from graphs import file_io, graphs, misc, plot_styles, utilities
 
 
-class Preferences():
+class Preferences(dict):
```
Sjoerd1993, line 10: Took me a minute to find that Preferences now inherits from dict, which is why this all works. Clever solution.

Sjoerd1993 (approved):
Looks good, clever trick to let preferences inherit from `dict`, and store the config settings directly in self

PR #312, opened 2023-06-13
Fix edit item

Sjoerd1993:
Fixes two issues in the edit item window:

- When the window was closed, the changes were not applied
- In the latest build, when clicking on a gear icon it would always just open the edit window of the top item. Regardless of which one was open.

### DIFF
--- a/src/edit_item.py
+++ b/src/edit_item.py
@@ -30,9 +30,9 @@ class EditItemWindow(Adw.PreferencesWindow):
         utilities.populate_chooser(self.markers, sorted(misc.MARKERS.keys()))
         self.load_values()
         utilities.populate_chooser(self.item_selector, names)
-
-        self.item_selector.set_selected(names.index(self.item.name))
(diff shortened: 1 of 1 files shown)


src/edit_item.py
```diff
@@ -30,9 +30,9 @@ def __init__(self, application, item):
         utilities.populate_chooser(self.markers, sorted(misc.MARKERS.keys()))
         self.load_values()
         utilities.populate_chooser(self.item_selector, names)
-
-        self.item_selector.set_selected(names.index(self.item.name))
+        self.item_selector.set_selected(names.index(item.name))
         self.set_transient_for(self.props.application.main_window)
+        self.connect("close-request", self.on_close)
```
Sjoerd1993, line 35: I actually thought I had already removed this line... but turns out I didn't solve the underlying issue yet.

This works as a workaround, but at least at one point we should solve it in the blueprint itself. For some reason the close-request signal does not connect properly to this function from the blueprint.

PR #313, opened 2023-06-13
Fix deleting multiple items

Sjoerd1993:
In the latest build, only the top item gets deleted when pressing the delete button. The reason for this is that it loops through the dictionary itself in actions.py, changing the dictionary size upon iteration which is not allowed. This solves this by looping through the selected keys instead like in previous releases.

### DIFF
--- a/src/actions.py
+++ b/src/actions.py
@@ -117,12 +117,10 @@ def open_project_action(_action, _target, self):
 
 
 def delete_selected_action(_action, _target, self):
-    delete_list = []
-    for item in self.datadict.values():
-        if item.selected:
-            delete_list.append(item)
-    for item in delete_list:
-        graphs.delete_item(self, item.key, True)
+    selected_keys = [key for key, item in self.datadict.items()
+        if item.selected]
+    for key in selected_keys:
+        graphs.delete_item(self, key, True)
 
 
 def translate_x_action(_action, _target, self):
--- a/src/actions.py
+++ b/src/actions.py
@@ -118,7 +118,7 @@ def open_project_action(_action, _target, self):
 
 def delete_selected_action(_action, _target, self):
     selected_keys = [key for key, item in self.datadict.items()
-        if item.selected]
+                     if item.selected]
     for key in selected_keys:
         graphs.delete_item(self, key, True)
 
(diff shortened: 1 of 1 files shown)


cmkohnen (changes requested):
Ah, the reason, I reformatted this is to have on less iteration.
Using your approach we iterate twice.
However I found that just inserting a `.copy()` like the following works just as well and resorts to only one iteration:
`for item in self.datadict.copy().values()`

Sjoerd1993:
> Using your approach we iterate twice.

Yeah, once over the entire dictionary to retrieve the selected keys, and then once more over each selected key to delete them. Making a copy of the datadict works indeed without any issues, and keeps everything contained to a single iteration.

It's an interesting hack actually, because you are deleting the items in your copy. But because Python doesn't create an actual copy, but refers to the same object in memory when copying dictionaries, it deletes the original item as well. I honestly didn't expect that would work (as both the copy and the original should change size in iteration right?), but somehow it does.  

It's a fine solution for me. As it's not a deep copy, it won't really change memory consumption by making this copy either.

cmkohnen:
It behaves actually a bit different. The copied dict will stay the same size and should get picked up by the garbage collector when the function ends. Since items contain their own key, we can reference the actual item to be deleted, as the graphs.delete() function takes only a key

src/actions.py
```diff
@@ -117,11 +117,10 @@ def on_response(_dialog, response):
 
 
 def delete_selected_action(_action, _target, self):
-    for item in self.datadict.values():
+    for item in self.datadict.copy().values():
         if item.selected:
             graphs.delete_item(self, item.key, True)
 
-
```
cmkohnen, line 124: don't know why flake didn't pick up on this: two newlines required here

Sjoerd1993:
> the graphs.delete() function takes only a key

Somehow in my head I was thinking we were deleting the copy on each loop, taking away the original with it. The deletion is indeed done on the actual item using the key. Either way, it does create a new dictionary in memory (at least within the life-span of the function), meaning the original will remain its size anyway. 

I got things mixed up with regular assignments where both dictionaries refer to the same points in memory I've had some issues with regular shallow copies in the past, where I was probably thinking of. Either way, the point is that you are correct.

Thanks for the elaboration.

cmkohnen (approved):

PR #314, opened 2023-06-13
drop advanced import

cmkohnen:
Drop the advanced import option. Also use a MenuButton for adding again and split off import and equation. In the future adding text, and other items will get another section

I found that the keybind we used didn't work either way as it toggles between light and dark mode

### DIFF
--- a/data/ui/window.blp
+++ b/data/ui/window.blp
@@ -15,10 +15,10 @@ template $GraphsWindow : Adw.ApplicationWindow {
       }
       Box {
         visible: bind-property sidebar_flap.folded inverted;
-        Adw.SplitButton add_button {
+        MenuButton add_button {
           icon-name: "list-add-symbolic";
-          action-name: "app.add_data";
           tooltip-text: _("Add new data");
+          always-show-arrow: true;
           menu-model: add_data_menu;
         }
         Separator {styles ["spacer"]}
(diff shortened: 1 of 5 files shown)


Sjoerd1993:
> I found that the keybind we used didn't work either way as it toggles between light and dark mode

This is likely a default in your distribution, as I don't have such a shortcut in Fedora at least. The current shortcut works fine for me. Having said that, as discussed in #249, I'm down with removing the advanced import and unifying it to a single setting. So I think that's OK.

Might want to remove the advanced import from the actual menu itself, as currently it just shows a greyed out version.

image

> In the future adding text, and other items will get another section

Yeah, I've been thinking about where that would go. My initial thought was in the sidepanel, but in the "Add" section does make sense, as it literally "adds stuff" to the graph. Keeping things nice and consistent. ("Add" button adds stuff, and the sidebar changes stuff).

Sjoerd1993 (approved):
Ok, looking through the code this part was removed from the menu. Somehow it was kept in cache. Cleaning my build directory solved this....

Looks okay to me. I'm unsure if we need a horizontal line between "Add data from file" and "Add data from equation", but think it's fine. As per the discussion, I think streamlining the import workflow into one single entry for adding from files is the correct way forwards.

cmkohnen:
Wll, I'm running Fedora as well and Ctrl+Shift+N toggles light/dark and I haven't reconfigured anything. Might be an extension.

As you can see in the diffstat, I have changed the menu and it shows on my machine. Try changing the window.blp file (adding and deleting a space should suffice) this worked for me at least.

PR #317, opened 2023-06-13
Add optimize limits option

sstendahl:
Adds an "optimize limits" option, that resets the limits to the default settings. It's basically a fail-safe if a user has set limits and the data is way out of range, or if the set limits are somehow incorrect (which can happen sometimes).

The idea is to add this still in the 1.6.0 release


cmkohnen (approved):
LGTM

PR #320, opened 2023-06-13
Remove extra lines of code

soumyaDghosh:


Sjoerd1993 (approved):
I get this removed part is redundant because we don't use the WebKitgtk libraries right? 

Seems good.

soumyaDghosh:
Yup.. sure. The CI also shows it.

PR #324, opened 2023-06-14
Fix legend

Sjoerd1993:
Fixes an issue where the legend would not get hidden properly if the legend was turned off. In the current main branch, if you turn off the legend it does not disappear until the graph is actually reloaded (not refreshed). This fixes that issue.

### DIFF
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -215,10 +215,10 @@ class Canvas(FigureCanvas):
                     new_lines, labels,
                     loc=self.application.plot_settings.legend_position,
                     frameon=True, reverse=True)
-            else:
-                legend = self.top_right_axis.get_legend()
-                if legend is not None:
-                    legend.remove()
+        else:
+            legend = self.top_right_axis.get_legend()
+            if legend is not None:
+                legend.remove()
 
 
 class DummyToolbar(NavigationToolbar2):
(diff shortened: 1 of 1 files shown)


src/canvas.py
```diff
@@ -215,10 +215,10 @@ def set_legend(self):
                     new_lines, labels,
                     loc=self.application.plot_settings.legend_position,
                     frameon=True, reverse=True)
-            else:
-                legend = self.top_right_axis.get_legend()
-                if legend is not None:
-                    legend.remove()
+        else:
```
cmkohnen, line 218: This change reintroduces the bug, where deleting items won't make it disappear.
```suggestion
                return
        legend = self.top_right_axis.get_legend()
        if legend is not None:
            legend.remove()
```

src/canvas.py
```diff
@@ -215,10 +215,10 @@ def set_legend(self):
...
+            legend = self.top_right_axis.get_legend()
+            if legend is not None:
+                legend.remove()
```
cmkohnen, line 221: This change reintroduces the bug, where deleting items won't make it disappear.
```suggestion
                return
        legend = self.top_right_axis.get_legend()
        if legend is not None:
            legend.remove()
```

Sjoerd1993: Yes this suggest solution works great for me. Instead of relying on an explicit else, it just ends the function using a return when successfully executed. Which is OK.

Shortened it slightly by not explicitly assigning a separate "legend" variable.

cmkohnen (approved):

PR #329, opened 2023-06-14
Add filters to new file dialog

Sjoerd1993:
Adds filters to the "Add data from file" dialogs. Also reformats the `create_file_filters` function slightly, to take in a list of suffixes. This way multiple suffixes can be added to a single filetype.

### DIFF
--- a/src/export_figure.py
+++ b/src/export_figure.py
@@ -61,6 +61,6 @@ class ExportFigureWindow(Adw.Window):
         dialog = Gtk.FileDialog()
         dialog.set_initial_name(f"{filename}.{file_suffix}")
         dialog.set_accept_label(_("Export"))
-        dialog.set_filters(utilities.create_file_filters([(fmt, file_suffix)]))
+        dialog.set_filters(utilities.create_file_filters([(fmt, [file_suffix])]))
         dialog.save(application.main_window, None, on_response)
         self.destroy()
--- a/data/se.sjoerd.Graphs.appdata.xml.in.in
+++ b/data/se.sjoerd.Graphs.appdata.xml.in.in
@@ -50,6 +50,7 @@
     <release version="1.6.1" data="2023-06-15">
         <p>Minor patch with some bug fixes:</p>
         <ul>
+	  <li>Filters have been added for the "New Data" dialog.</li>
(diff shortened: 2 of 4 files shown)


Sjoerd1993:
image

src/export_figure.py
```diff
@@ -61,6 +61,7 @@ def on_response(dialog, response):
         dialog = Gtk.FileDialog()
         dialog.set_initial_name(f"{filename}.{file_suffix}")
         dialog.set_accept_label(_("Export"))
-        dialog.set_filters(utilities.create_file_filters([(fmt, file_suffix)]))
+        dialog.set_filters(utilities.create_file_filters(
+            [(fmt, [file_suffix])]))
```
cmkohnen, line 65: With the ability to add multiple formats I'd use all available ones in this case.
On line 45 we use just the first, but we can use just the first for the initial filename.

cmkohnen (approved):

Sjoerd1993:
Matplotlib savefig still takes a single file format at a time, but I did restructure this code a bit. Now technically the filter adds different suffixes, but the end-experience is the same.

There was another bug however that seems to have made it into version 1.6.0, which is pretty major. When it loads the figure dpi into the settings, it loads this as a string. Which leads to problems when opening preferences or exporting the figure. Making both these options (preferences and export figure) inaccessible in 1.6.0. My guess is that this somehow got in with the refactor of the settings last week. 

Given the severity of this issue, I'm tempted to rush a 1.6.1 patch as soon as today. With the other bug fixes, this would lead to a relatively stable base again.

cmkohnen:
Ah, I see. I thought these values were stored in the config as their respective type. Maybe we should change that starting with patch 1.6.1

Sjoerd1993:
> Ah, I see. I thought these values were stored in the config as their respective type. Maybe we should change that starting with patch 1.6.1

The reasoning for it being stores as string is that some other values are set by GtkEntry, as those can be any arbitrary float where a spinner makes less sense (e.g. axis limits and step size for equations), and there a string is needed as input. However for these spinners (used in figure dpi and clipboard length), there's no reason storing this as string actually. So changed those to regular strings. (Json [does not distinguish](https://json-schema.org/understanding-json-schema/reference/numeric.html) between floats and ints, hence it still being stored as general "number". But that's still less bug-prone than using strings)

PR #331, opened 2023-06-14
Prepare first 1.6.1 patch

Sjoerd1993:
See title, fixes some important bugs. Giving the preferences and export figure is broken, it's probably good to get this release out relatively quickly.


Sjoerd1993:
If there's any more issues that are known, we can wait a bit with pushing this release. But in principle, I think this is about ready as hotfix patch, fixing the most pressing issues.

cmkohnen:
Maybe checking for dependency updates, which we kinda glossed over for 1.6

Sjoerd1993:
Somehow it seems all dependencies are still up-to-date with the latest releases on Pip at least, so those seem OK.

cmkohnen:
> Somehow it seems all dependencies are still up-to-date with the latest releases on Pip at least, so those seem OK. I can of course wait with the release itself until later in the evening, just in case something pops up. I've spend most of today testing (hence the nature of most PR's today), and I think most major/obvious stuff is fixed now.

Major dependencies are up to date, but two dependencies of them recieved minor updates. Using snap they are already used I think

cmkohnen (approved):

PR #334, opened 2023-06-14
update dependencies

cmkohnen:


Sjoerd1993:
When I checked, I just ran the "major dependencies" (matplotlib, numpy and scipy) through `flatpak-pip-generator` and checked which versions were requested by pip. Those were all up-to-date to the current manifest. My guess is that contourpy and fonttools have had an update in the meantime, hence matplotlib/numpy/scipy not requesting these newer versions.

Either way, I'll check if these new versions cause any issues, otherwise I'm fine with using these updated ones.

Sjoerd1993 (approved):
lgtm

PR #338, opened 2023-06-15
Fix appdata changelist

Sjoerd1993:
There's a slight syntax error in the changelist, making the latest changelist not show up in GNOME Software (at least for the latest release). Fixing it would require tagging and pushing an entire new release and it doesn't really change the user experience, so that may not be worth to fix it in this version (the info is still there in whats new also), but it's good to fix it anyway so that it will show up in the history after next patch.

I guess this syntax error is why the changelist for 1.6.1 doesn't show up at least... I tried validating the xml file and couldn't detect any more errors.

### DIFF
--- a/data/se.sjoerd.Graphs.appdata.xml.in.in
+++ b/data/se.sjoerd.Graphs.appdata.xml.in.in
@@ -60,14 +60,13 @@
     </release>
     <release version="1.6.0" date="2023-06-13">
       <description>
-	<p>New in Graphs:
+	<p>New in Graphs:</p>
 	<ul>
 	  <li>Data can now be opened directly from project files.</li>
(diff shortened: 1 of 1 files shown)


cmkohnen:
I mean, we don't have to tag it or anything. We could just point it to this commit or the respective merge commit since nothing has changed otherwise

cmkohnen (approved):

Sjoerd1993:
If I recall correctly, the workflow in Flathub now requires a tag for the main branch. Could be mixing this up with the general recommendation to use a tag though. But hence the comment that "fixing" this in Gnome Software would essentially require a new release, which may not be really worth it considering how minor the issue is.

PR #339, opened 2023-06-15
Redesign navigation

cmkohnen:
Use two buttons for navigating the view back and forth.
This functionality piggybacks of matplotlibs implementation in the
toolbar, requiring very little logic from our side.
Limits in Plot Settings are now automatically updated, whenever an
action (zoom, pan, optimize) is performed. The history however does not
persist across a reload yet.

Also:
Upon init, the toolbar sets itself as the canvas.toolbar property, thus
making our canvas.dummy_toolbar redundant.

### DIFF
--- a/src/actions.py
+++ b/src/actions.py
@@ -80,7 +80,7 @@ def view_back_action(_action, _target, self):
         self.canvas.toolbar.back()
         self.canvas.apply_limits()
         # Fix weird view on back
-        self.canvas.draw()
+        graphs.refresh(self)
 
 
 def view_forward_action(_action, _target, self):


src/actions.py
```diff
@@ -75,17 +75,14 @@ def optimize_limits_action(_action, _target, self):
...
 
 
-def restore_view_action(_action, _target, self):
-    self.canvas.load_limits()
-    graphs.refresh(self)
-
-
 def view_back_action(_action, _target, self):
-    self.canvas.dummy_toolbar.back()
+    self.canvas.toolbar.back()
```
Sjoerd1993, line 79: Since DummyToolbar inherits from NavigationToolbar2, isn't this essentially the same thing?

src/actions.py
```diff
@@ -75,17 +75,14 @@ def optimize_limits_action(_action, _target, self):
...
+    self.canvas.apply_limits()
 
 
 def view_forward_action(_action, _target, self):
-    self.canvas.dummy_toolbar.forward()
+    self.canvas.toolbar.forward()
+    self.canvas.apply_limits()
```
Sjoerd1993, line 85: Is the apply_limits() really necessary here? When it sets the view back the previous one, the limits are in principle already set. Seems to work for me when removing this line as well, but I could be missing something or maybe some edge case.

src/canvas.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, application):
                      self.top_right_axis]:
             axis.get_xaxis().set_visible(False)
             axis.get_yaxis().set_visible(False)
-        self.dummy_toolbar = DummyToolbar(self)
+        DummyToolbar(self)
```
Sjoerd1993, line 46: We still create the dummy toolbar, but never assign it. Is this so that we can still use the overwritten rubberband functions?

src/canvas.py
```diff
@@ -244,6 +243,19 @@ def remove_rubberband(self):
         self.canvas._rubberband_rect = None
         self.canvas.queue_draw()
 
+    # Overwritten function - do not change name
+    def push_current(self):
+        super().push_current()
```
Sjoerd1993, line 248: The difference with the default push_current is basically that it now automatically applies the limits again. Why is this necessary?

src/canvas.py
```diff
@@ -244,6 +243,19 @@ def remove_rubberband(self):
...
+        self.canvas.apply_limits()
+
+    # Overwritten function - do not change name
+    def set_history_buttons(self):
+        self.canvas.application.main_window.view_back_button.set_sensitive(
+            self._nav_stack._pos > 0)
+        self.canvas.application.main_window.view_forward_button.set_sensitive(
+            self._nav_stack._pos < len(self._nav_stack._elements) - 1)
```
Sjoerd1993, line 256: This behaviour works well for me.

src/plot_settings.py
```diff
@@ -157,4 +157,5 @@ def on_close(self, *_args):
             ui.reload_item_menu(self.props.application)
         else:
             self.props.application.canvas.load_limits()
+            self.props.application.canvas.toolbar.push_current()
```
Sjoerd1993, line 160: Similar question here as before, what is the reasoning that a push_current is needed now here?

Sjoerd1993 (commented):
Code-wise my main points are just about elaboration, but that all seems to work well. 

Design-wise, playing around with it works quite well. I'm mainly wondering how much clutter it adds, and if it should be permanently there in the top bar or hidden in the view menu as with the current main branch. I'll try and see if I can get some guidance from the GNOME HIG as well. Just wanted to hear opinions about this.

Also, I was thinking maybe we want to couple the default "Home" behaviour in Matplotlib to the Eye in the view menu, but on the other hand that may be confusing to have both the "Optimize limits" and the "Home" buttons both available as they are quite similar. So this solution may work fine. Again, just throwing it out there for opinions about this.

If we do get rid of the restore view button, we should remove it from the keyboard shortcuts as well.

Sjoerd1993:
> I'm also getting a bug where the canvas has a tendency to now always stretch out completely on a view change, but haven't looked that closely into that. Maybe it already exists in the previous implementation as well.
> 
> Running optimize_limits fixes this again. [Screencast from 2023-06-16 10-58-51.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/5056a733-fd84-4149-8b41-b987a9c65834)

Adding a self.canvas.draw() to the back and forward action mostly fixes this. It still moves slightly at times, but it immediately snaps back to the correct position. Which is an improvement.

cmkohnen:
As many comments are related, I'll answer them as best as I can in a single answer:

The base NavigationToolbar2 automatically sets itself to the `canvas.toolbar` property on init. So the whole time we had both `canvas.toolbar` and our own `canvas.dummy_toolbar` as references to one and the same object. The only thing I did was to use the inbuilt reference and drop our own.

The reason why `push_limits()` is extended, is because actions like panning or zooming call it when the user releases the mouse button. The view gets updated and most importantly, pushed to the view stack (The one responsible for storing back and forth states). Since it is internally called when the view changed we have to set the limits in plot settings as well.
tldr: When panning and zooming plot limits get set

Also `push_limits()` gets called across our code now. The reason for that is, if we change the view via optimizing limits or manually setting them, it gets added to the view stack.
A slight flaw with these cases is the following flow:
- we apply limits to plot settings
- we load limits to the graph
- we apply limits from the graph to view stack and plot settings
We could just skip the first step in these cases and apply any new limits to the graph directly, but that would require us to implement a fail save everywhere when setting limits fails. I'll see if I can think of something here.

Regarding dropping the "home" action:
I have thought a bit about this and came to the conclusion, that it really doesn't fit right. In most cases, home would either lead back to the absolute initial view, or when an item was first added. If something was added afterwards that would mean the initial view might not be the desired one. Especially with the Restore View action, which is a better fit in my opinion.
I feel like the back and forth buttons behave rather snappy and using them in quick succession isn't that much slower as heading to the view menu and using optimize limits.

One issue that just came to my attention, is locking the action itself. The buttons are locked when the action isn't available. However the keyboard shortcut still works. This can lead to the bug you described.

Sjoerd1993:
Right, so we bascially need to push the limits in the view stack to get these back and forth states work properly. It makes more sense than to actively call these. 

This bug still persists in the latest build, even when just using the buttons. See the attached video. Drawing the canvas again on back/forward action would be an OK workaround if we can't really find the cause directly.

[Screencast from 2023-06-16 13-21-57 (trimmed).webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/ad6103f5-f055-4a71-91ad-86aa19954147)

Regarding the "Restore View" action. Originally (1.5.x and before), the button would simply be identical to "Optimize limits".  Having used this build a bit, it feels like going back and forth from the menu is simply not efficient enough. If you do some specific zooming, and then want to go back you'd have to go in and out of the menu several times. Having these buttons readily available is quite convenient. This is not really a problem for "optimize limits" which you generally just press once. I think the current design (this PR I mean) works quite well from an ergonomics perspective (in the sense of how often you need to dive into menus, and how accessible and discoverable is each action). So I think this works really well actually now I'm using it a bit more.

Sjoerd1993 (approved):
Now I think of it, this bug was actually present in the older implementation as well. So I'm also fine dealing with it in a separate PR as it's basically not related (or at least not caused by) this new part of the code.

cmkohnen:
The bug appears only when navigating back, not forth from my testing. I think this might actually an upstream bug. I'm fine however to append the fix here, as I have to deal with the shortcut menu as well.

Sjoerd1993:
It seems like this behaviour is only triggered on the "back" action, the forwards action always has the axes labels fixed on the same position. Code-wise I can't find any difference in our code between these actions, except for the call to the toolbar function being different obviously. Could be an upstream thing.

cmkohnen:
> **Edit:** I swear I didn't see you latest comment yet before writing the above.

They were a few seconds apart, with GitHubs interface (especially when checks are running) I don't blame you

src/actions.py
```diff
@@ -75,17 +75,18 @@ def optimize_limits_action(_action, _target, self):
...
-    graphs.refresh(self)
-
-
 def view_back_action(_action, _target, self):
-    self.canvas.dummy_toolbar.back()
+    if self.canvas.toolbar.back_enabled:
+        self.canvas.toolbar.back()
+        self.canvas.apply_limits()
+        # Fix weird view on back
+        self.canvas.queue_draw()
```
Sjoerd1993, line 83: It seems to improve, but it still sometimes gets stuck in shifted state. Using a regular `self.canvas.draw()` does not seem to have the issue.

Sjoerd1993 (changes requested):
Overall seems to be good now. Although I think it would work better with a regular `canvas.draw()` as I still seem to be able to trigger this issue quite easily. Even though it does seem to snap back sometimes (just not always).

src/actions.py
cmkohnen: Okay, looking at the documentation, queue_draw waits on the GUI, so when its called multiple times like draw_idle, it only renders once. However since the `back()` function calls draw_idle down the line, that is maybe the root problem here.

src/actions.py
```diff
@@ -79,6 +79,8 @@ def view_back_action(_action, _target, self):
     if self.canvas.toolbar.back_enabled:
         self.canvas.toolbar.back()
         self.canvas.apply_limits()
+        # Fix weird view on back
+        self.canvas.draw()
```
Sjoerd1993, line 83:
````suggestion
graphs.refresh(self)
```

Actually `self.canvas.draw()` also seems to get stuck. The only thing that actually seems consistent so far is a full refresh. Which still feels as snappy as without the fix. So I think that should work.

Sjoerd1993: Actually `self.canvas.draw()` also seems to get stuck. The only thing that actually seems consistent so far is a full refresh. Which still feels as snappy as without the fix. So I think that should work.

Sjoerd1993:
I was actually missing this feature already when I was just working with the stable release. The old "Next/Previous View" options did not really work nicely at all.

One weird use case I thought that was useful myself was also when the view gets reset due to an action being performed when you don't want it to. Now you can just simply click on the previous arrow to get back to the previous view.

 At one point we may want to look a bit more careful into which actions should trigger new limits. It's pretty useful to do this when translating/multiplying data for instance, as things quickly get out of frame, but maybe "Cut" doesn't need a limit reset. Especially now we have a separate button to optimize the limits.

PR #341, opened 2023-06-16
reorder preferences

cmkohnen:
Use a third group to display setting related to importing data.
If we extend configurable filetypes, this will result in less clutter.


Sjoerd1993 (approved):
I'm a bit confused as to why the source to graph-symbolic is slightly changed, but overall an improvement that makes it easier to find specific settings.

cmkohnen:
> I'm a bit confused as to why the source to graph-symbolic is slightly changed, but overall an improvement that makes it easier to find specific settings.

That was me generating a new one from icon library, not realising we already have it in place.

PR #342, opened 2023-06-16
Use functions for operations.

cmkohnen:
For the operations only invoked via their buttons, it is redundant to invoke a complete action only to have a few lines of code coupled to them.
This uses signal handlers to perform logic when the button is clicked.

### DIFF
--- a/data/ui/window.blp
+++ b/data/ui/window.blp
@@ -231,15 +231,16 @@ template $GraphsWindow : Adw.ApplicationWindow {
                       label: _("Shift");
                     }
                     tooltip-text: _("Shift all data vertically with respect to each other");
-                    action-name: "app.shift_vertically";
+                    clicked => $shift_vertically();
                   }
 
-                  Button normalize_button {
+                  Button {
                     margin-start: 5;
                     margin-end: 5;
                     margin-top: 5;
                     margin-bottom: 5;
                     hexpand: true;
+                    sensitive: bind shift_vertically_button.sensitive;
                     layout {
                       column: 1;
                       row: 0;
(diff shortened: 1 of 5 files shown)


data/ui/window.blp
```diff
@@ -231,15 +231,16 @@ template $GraphsWindow : Adw.ApplicationWindow {
...
                   }
 
-                  Button normalize_button {
+                  Button {
                     margin-start: 5;
                     margin-end: 5;
                     margin-top: 5;
                     margin-bottom: 5;
                     hexpand: true;
+                    sensitive: bind shift_vertically_button.sensitive;
```
Sjoerd1993, line 243: It'ss probably cleaner like this:

```suggestion
                    sensitive: bind-property item_list.visible;
```

src/ui.py
```diff
@@ -18,25 +18,7 @@ def enable_data_dependent_buttons(self):
...
-        win.normalize_button,
-        win.center_button,
-        win.derivative_button,
-        win.integral_button,
-        win.transform_button,
-        win.combine_button,
-    ]
-    for button in dependent_buttons:
-        button.set_sensitive(enabled)
+    self.main_window.shift_vertically_button.set_sensitive(enabled)
```
Sjoerd1993, line 21: If we bind the buttons to  item_list.visible, we can get rid of this entire function

Sjoerd1993 (changes requested):
Looks mostly OK, but it's more elegant to bind the action buttons to the item list instead of to a single button.

src/ui.py
cmkohnen: No, because when no item is selected, the buttons become insensitive too, whilst the item_list isn't invisible

Sjoerd1993: Yep, didn't consider the  other condition of having no item selected. My bad!

Connecting them to the shift_vertically function removes the necessity to assign them in `window.py.` Otherwise I don't have a strong preference to either solution, think this is fine.

Sjoerd1993 (approved):

PR #343, opened 2023-06-18
Remove redundant lines of codes

Sjoerd1993:
Because these parts of the axes always are turned off anyway (line 84 and down), it makes no sense turning them visible first. So these lines can be removed without losing anything


cmkohnen:
I took the liberty of removing the next lines as well, as these are already set invisible above as well.

Sjoerd1993:
Yes that entire part indeed is redundant.

I think the logic was to turn all used axes visible, and then turn one of the overlapping ones off. Which in the end is just the same as only turning the oned you need visible in the first place.

cmkohnen (approved):

PR #344, opened 2023-06-19
Fix overlapping ticks

Sjoerd1993:
When multiple axes are used, the ticks from each axes are overlapping. See screenshot:

image

This is never something you want in an actual graph. Basically, the change in this PR is that for a specific axis, opposite ticks are not plotted now when the opposite axis is already in use by a data set. This does mean that the condition for each axis needs to be checked separately with slightly different conditions, so I could not get it to work in a single loop. But it's readable enough. See screenshot for the same data after this PR:

image

### DIFF
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -148,7 +148,7 @@ class Canvas(FigureCanvas):
         self.top_right_axis.set_yscale(plot_settings.right_scale)
 
     def set_ticks(self):
-        """ Set the tick parameters for the axes in the plot."""
+        """Set the tick parameters for the axes in the plot"""
         bottom = pyplot.rcParams["xtick.bottom"]
         left = pyplot.rcParams["ytick.left"]
         top = pyplot.rcParams["xtick.top"]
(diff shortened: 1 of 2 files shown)


src/canvas.py
```diff
@@ -75,6 +75,7 @@ def on_draw_event(self, widget, ctx):
             self.figure.draw(self._renderer)
 
     def plot(self, item):
+        self.set_ticks()
```
cmkohnen, line 78: I'd move this in `graphs.refresh()`. That way it gets called just once per refresh not once per item, which should suffice?

Sjoerd1993:
> which means, that the bottom axis displays ticks on the top one if enabled IIRC.

That is on purpose, it's a relatively common use-cases to have ticks on the opposite sides as well. At least in scientific publications. See here a graphs that we used in a paper that we recently published (this particular figure was not made in Graphs):
image

src/canvas.py
Sjoerd1993: Yes that definitely suffices. My immediate thinking was that these ticks change when plotting new data, so then they should be refreshed. Putting it in the general refresh may trigger this function in some other cases where it's not needed. But the reverse is also true (putting it in plot also leads to some unnecessary calls).

Putting it in the refresh function may make more sense either way. As it's basically meant to refresh the axes properties.

cmkohnen:
Oh, I didn't think of that

src/canvas.py
```diff
@@ -148,18 +148,35 @@ def set_axis_properties(self):
...
+            tick_params = {
+                "bottom": bottom and
+                          ("bottom" in directions or not used_axes["bottom"]),
+                "left": left and
+                        ("left" in directions or not used_axes["left"]),
+                "top": top and
+                       ("top" in directions or not used_axes["top"]),
+                "right": right and
+                         ("right" in directions or not used_axes["right"]),
+            }
```
cmkohnen, line 178:
```suggestion
            tick_params = {
                "bottom": bottom
                and ("bottom" in directions or not used_axes["bottom"]),
                "left": left
                and ("left" in directions or not used_axes["left"]),
                "top": top
                and ("top" in directions or not used_axes["top"]),
                "right": right
                and ("right" in directions or not used_axes["right"]),
            }
```

cmkohnen (approved):
Aside from formatting issues, I really like this change

cmkohnen:
Since Github auto-collabses outdated comments: 
I've looked at flake8 documentation and both W503 and W504 are on the ignore-list we might put them on there too

Sjoerd1993:
Finally just installed flake8 locally, which simplifies the troubleshooting significantly. Always fun if it complains about a line break after the binary operator, and then about a line break before the binary operator. Not sure what PEP8 says, but  flake8 stops complaining if I put the line break after the bracket. Which seems readable enough to make sense to me.

PR #346, opened 2023-06-19
WIP: fix scaling in plot settings

Sjoerd1993:
Currently when the scaling of an item is set in the Plot Settings menu, the new scale is not automatically loaded, because it only refreshes the graph.

Even when triggering a reload of the graph however, the scale items in the view menu are not automatically set to the correct scaling either. This is where I needed to dig a bit into the documentation, as I couldn't immediately find out how to programmatically set a new value for these items.

However, I was also considering removing these scalings from the plot settings all together, as these scalings are also directly available in the menu itself, which you need to click anyway to get to that scaling. Opinions on that (or suggestions how to set the scaling in the view menu programmatically) are welcome of course.

### DIFF
--- a/data/ui/window.blp
+++ b/data/ui/window.blp
@@ -562,7 +562,7 @@ menu view_menu {
     item (_("Optimize Limits"), "app.optimize_limits")
   }
   section {
-    submenu {
+    submenu top_scale {
       label: _("Top X Scale");
       item {
         label: _("Logarithmic");
@@ -575,7 +575,7 @@ menu view_menu {
         target: "linear";
       }
     }
-    submenu {
+    submenu bottom_scale {
       label: _("Bottom X Scale");
       item {
         label: _("Logarithmic");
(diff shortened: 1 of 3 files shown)


Sjoerd1993:
So basically I've been trying to trigger the `change_top_xscale` for the top x-axis whenever the axis gets changed in plot settings. But I'm kinda running into a wall here. The function requires the action and target as argument, but those are never really stored anywhere. I could try to get the children of each menu item and get the target attributes and trigger the respective actions, but that will get really messy code-wise.

Do you maybe have any ideas how to trigger these actions/sync these states programmatically from the plot settings menu @cmkohnen? Removing these from this window is also an easy solution, given these options are available in the dropdown menu as well. But not 100% sure yet if that's the way forwards.

cmkohnen:
One solution is to keep the current structure, keep a reference to the actions, and activate the actions if a change is detected.

I would however refactor the menu a bit, for these changes to be instant. (Connecting to the notify::selected)

Sjoerd1993:
Some kind of refactor for the menu probably will be the cleanest solution. Either way, it won't build upon the changed code in this PR, and I'm closing this for now until I've got a slightly more proper solution.

PR #347, opened 2023-06-20
Support item transparency and text rotation

cmkohnen:
Add Support for setting an items transparency as well as rotation for texts. Not accessible via UI yet.
Also slight refactor to simplify handling of common parameters when drawing


Sjoerd1993 (approved):
lgtm

Sjoerd1993:
Not completely sure why the last flake8 workflow is failing, could just be a fluke.

PR #348, opened 2023-06-20
Support writing headers to a file.

cmkohnen:
When exporting write headers, if present. Also refactor file_io to shorten common operations.


Sjoerd1993 (approved):
Adding the headers to the exported were on my list either way. Seems to work fine as long as the headers are item properties at least.

PR #349, opened 2023-06-20
Limit workflow runs

cmkohnen:
Currently workflows are run on every push and every pr. This changes that to only run for pushes directly to main and any pr that is targeted to main. This eliminates duplicate workflows, when pushing to a branch, that is already a pr

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,10 @@
-on: [push, pull_request]
+on:
+  push:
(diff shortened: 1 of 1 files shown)


Sjoerd1993 (approved):
Sounds like a good idea either way, separate branches don't really _need_ to pass the workflows anyway before they are merged into main. No real use in pushing these for every commit for a testing ground branch.

PR #350, opened 2023-06-20
Add transperency to item options.

cmkohnen:
Add a transparncy slider to the edit item window. This slider is in percent from 0 - 100. Also update the grid transparency slider in the plot style editor to be in percent as well.

### DIFF
--- a/data/ui/edit_item.blp
+++ b/data/ui/edit_item.blp
@@ -38,6 +38,20 @@ template $EditItemWindow : Adw.PreferencesWindow {
         title: _("Y-Axis Position");
         subtitle: _("The position of the Y-axis");
       }
+
+      Adw.ActionRow {
+        title: _("Transparency");
+        subtitle: _("The transparency of the item");
+
+        Scale transparency {
+          draw-value: true;
+          width-request: 200;
+          adjustment: Adjustment {
+            lower: 0;
+            upper: 100;
+          };
+        }
+      }
     }
 
     Adw.PreferencesGroup item_group {
--- a/data/ui/edit_item.blp
+++ b/data/ui/edit_item.blp
@@ -38,20 +38,6 @@ template $EditItemWindow : Adw.PreferencesWindow {
         title: _("Y-Axis Position");
         subtitle: _("The position of the Y-axis");
       }
-
-      Adw.ActionRow {
-        title: _("Transparency");
-        subtitle: _("The transparency of the item");
-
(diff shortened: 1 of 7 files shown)


Sjoerd1993:
I think it's a good decision, but some points:

1 - From my experience with creative tools, I think it's more common to define this in terms of opacity, so with inverted X-scales. (Zero being fully transparant, 100 being not transparant at all)
2 - Having this in the menu maybe works well for specific items, but for regular data points I think it makes more sense to set this in the color button. This can be achieved quite easily with just a few line changes in the `item_box.py` code. The relevant part that includes all those changes then becomes:

```
    def choose_color(self, _):
        color = utilities.hex_to_rgba(self.item.color)
        dialog = Gtk.ColorDialog()
        dialog.set_with_alpha(True)
        dialog.choose_rgba(
            self.application.main_window, color, None,
            self.on_color_dialog_accept)

    def on_color_dialog_accept(self, dialog, result):
        try:
            color = dialog.choose_rgba_finish(result)
            if color is not None:
                self.item.color = utilities.rgba_to_hex(color).upper()
                self.item.transparency = 100-100*color.alpha
                self.update_color()
                clipboard.add(self.application)
                graphs.refresh(self.application)
        except GLib.GError:
            pass
```
Note that I still used the transparency from 0 for being non-transparent to 100 for being completely transparent. It's probably better to revert that.

3 - If we put the transparency there in the item menu (again, I think for data lines at least it may make more sense to use the built-in GTK options in the colordialog), I think it makes more sense to put it in _Line Style_ category rather than the _Data_ category.

cmkohnen:
I've changed it, so that the slider in the plot style menu has been renamed and for regular items the alpha of the chosen color is used. Since this is a value between 0 and 1, we dont have to change anything to pass on to matplotlib

Sjoerd1993:
Looking quite nice. I think we can keep the slider still in the item menu, but have it be part of the linestyle category instead of data, as it effects the linestyle directly.

Also at the moment the value in the slider does not sync up with the actual item attribute.

cmkohnen:
My intention was to remove the slider, as it doubles with the alpha then. Perhaps you still have the one with the slider cached?

Sjoerd1993 (approved):
You are absolutely correct, the old UI files were still cached. It's a common issue, in principle for every change in the UI I need to rebuild the project. It's a quick thing, but it's easy to miss.

PR #351, opened 2023-06-20
Minor transparency improvements

Sjoerd1993:
- Set the transparency in the color button as well
- Load the transparency correctly when choosing a new color. In the current main branch, when the color button is clicked, the actively selected colour is the non-transparent one. So pressing OK without doing any changes will make the item lose its transparency.

### DIFF
--- a/src/item_box.py
+++ b/src/item_box.py
@@ -56,11 +56,16 @@ class ItemBox(Gtk.Box):
 
     def update_color(self):
         self.provider.load_from_data(
-            f"button {{ color: {self.item.color}; }}", -1)
+            ("button {"
+             f" color: {self.item.color};"
+             f" opacity: {self.item.alpha};"
+             "}"),
+            -1)
 
     @Gtk.Template.Callback()
     def choose_color(self, _):
         color = utilities.hex_to_rgba(self.item.color)
+        color.alpha = self.item.alpha
         dialog = Gtk.ColorDialog()
         dialog.set_with_alpha(False)
         dialog.choose_rgba(


src/item_box.py
```diff
@@ -58,11 +58,16 @@ def on_dnd_prepare(self, drag_source, x, y):
 
     def update_color(self):
         self.provider.load_from_data(
-            f"button {{ color: {self.item.color}; }}", -1)
+            ("button {"
+             f" color: {self.item.color};"
+             f" opacity: {self.item.alpha};"
```
cmkohnen, line 63: Correct me if I'm wrong here, but is opacity not in contrast to alpha?

Sjoerd1993: There's a slightly different lingo used between Matpltolib and GTK.

Matplotlib uses the term `alpha` for the parameter that describes how transparent something is, where alpha = 1 means completely opaque and alpha = 0 means completely transparent. Meanwhile the GTK Color seems to work with "`opacity`" which describes exactly the same thing with the same limits. 

Strictly speaking I would agree with you though. To me, alpha descripes transparency while opacity describes how opaque something is. So alpha at 1.0 should describe a completely transparent object, but Matplotlib does it the other way around.

cmkohnen: Sorry my bad, I got things confused.
There are to my knowledge three different "frameworks" at play here:
- matplotlib
- GTK et. al.
- css

All three expect alpha to be a value between one and two, whilst 0 being completely transparent. CSS just uses the term opacity for that. I got it confused with transparency, which is the exact inverse.
Apologies.

cmkohnen (approved):

PR #353, opened 2023-06-21
Set view back/forward on undo/redo

Sjoerd1993:
Sets the view back or forwards on an undo and redo action, so that the view is also reset to the state before the action was performed.

### DIFF
--- a/src/clipboard.py
+++ b/src/clipboard.py
@@ -43,7 +43,7 @@ def undo(self):
     if self.clipboard_pos < -1:
         self.main_window.redo_button.set_sensitive(True)
     graphs.check_open_data(self)
-    actions.view_back_action(None,None,self)
+    actions.view_back_action(None, None, self)
     ui.reload_item_menu(self)
 
 
@@ -61,5 +61,5 @@ def redo(self):
     if self.clipboard_pos >= -1:
         self.main_window.redo_button.set_sensitive(False)
     graphs.check_open_data(self)
-    actions.view_forward_action(None,None,self)
+    actions.view_forward_action(None, None, self)
     ui.reload_item_menu(self)
--- a/src/clipboard.py
+++ b/src/clipboard.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import copy
 
-from graphs import graphs, ui
+from graphs import actions, graphs, ui
 
 
 def add(self):
@@ -43,6 +43,7 @@ def undo(self):
     if self.clipboard_pos < -1:
         self.main_window.redo_button.set_sensitive(True)
     graphs.check_open_data(self)
+    actions.view_back_action(None,None,self)
     ui.reload_item_menu(self)
 
 
@@ -60,4 +61,5 @@ def redo(self):
     if self.clipboard_pos >= -1:
         self.main_window.redo_button.set_sensitive(False)
     graphs.check_open_data(self)
+    actions.view_forward_action(None,None,self)
     ui.reload_item_menu(self)


Sjoerd1993:
You are correct, I was thinking in the context of actions, but indeed adding data or changing item properties also adds a state to the clipboard. Pushing the view by running `self.canvas.toolbar.push_current()` on each `clipboard_add`  largely solves this but leads to some other edge cases where the view can be lost. Will take a further look later, could be that it's triggered twice now in some cases in my code in these cases. Maybe some other solution would be needed.

The simple solution is to trigger `optimize_limits` on undo/redo, but that's not my favourite solution at all. I see the current behaviour as a bug, where the old limits are kept when doing ctrl+z see video. 
[Screencast from 2023-06-21 15-57-07.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/9c9dbaaf-b8d3-459b-8c68-0210445b2c33)
This becomes especially apparant/annoying when doing transformations where the axes change quite radically, such as fourier transforms. 

But thanks for your attention to detail here, as the current state of the PR indeed does introduce some new bugs with this behaviour, and should be looked at a bit more carefully.

Sjoerd1993:
I think you are right about the view stack not being suitable for the behaviour proposed in this PR. Unless we add a new state for each time the view stack is changed to the clipboard, but that is not really nice behaviour either and the clipboard can get full really quickly then when just panning and zooming.

Not resetting at all for all cases makes certain transformations "not work" on first sight, even with common use cases such as a translation or normalization. See video (normalize now works kind of OK because it's on log scale, but on linear scale it would just look like a flat line):

[Screencast from 2023-06-22 08-05-34.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/09a95d7d-c71a-4dac-93d1-0cfff7782a6f)

Having said that, I think at least the "cut" action should not trigger a reset of the limits. For the others that currently do reset the limits, I do think it's kinda motivated though as the behaviour in above screencast is not exactly intuitive and gives a broken feeling imo. But I am open to a discussion on this as well, as it would give consistent behaviour. Perhaps I can take a look what most graphing software does and what is generally seen as most intuitive.

- One option is to keep the actual limits stored in the clipboard states, and set those each time the next or previous state is triggered. This would require rewriting this a bit, and probably maybe saving the clipboard states as an own class with these properties. Potentially we should keep some of the current code then to keep backwards compatibility for a while with the old clipboard. This would in my view be the most intuitive behaviour from a user-facing point of view at least. Undo/redo puts the user back in the previous/next state. Including the view at that time. 

- Another easy option would be to optimize limits on each clipboard undo/redo. But that may not be super intuitive behaviour, and could be slightly annoying.

- Finally, another solution is to accept that the limits are kept like they are on undo/redo. With other words, keep the current behaviour, only performing the undo/redo on the actual items. I could argue that this is also what is done with creative tools for example. See video here below, where I just inserted random shapes in Inkscape, zoomed out, and then undid them with Ctrl+Z.  

[Screencast from 2023-06-22 08-10-47.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/5a2a4c66-b748-4317-be2a-bcc602d53de9)

Sjoerd1993:
Closing this for now. 
Leaning towards keeping the automatic rescaling, although we can probably remove it for the cut-function. 

Regarding the undo/redo behaviour, I'm tempted to keep it as-is, basically having it anologous to what creative tools do. At least that seems better than rescaling it on every undo/redo action. Saving the limits in the clipboard may get a bit ugly code-wise also.

cmkohnen:
I was about to say, that one possible solution, would be to keep a clipboard of views in sync with the actual clipboard. 
This would require us to implement own functionality of this feature, as matplotlibs one isn't suited for that task. However this could kill two birds with one stone, as this would make the view history survive a reload (and even be saved in projects).

Sjoerd1993:
Yeah, the way I've been thinking about it earlier is to create an own class for the clipboard  state basically. Which should contain all relevant states to get back to a specific position in the clipboard. Then the clipboard list that we use now could be filled with these items/states. Each time undo/redo is triggered, the state can then be loaded from there.

This should hook in reasonably well in the existing code where such a list already exit (except that it contains the `datadict` instead of the entire state on each position), with the same logic of switching states to the correct position with the undo/redo actions. I'll look into that at a later date when I've got some free time.  

Will try to see if I can get some relatively clean logic to make this backwards compatible with the older project files. Perhaps I'll designate a separate .py file for such backwards compatibility logic. But that's a thing I'll look into when I get there. Perhaps that won't be needed yet (it can probably be done in a single function). But in the future we may end up with some specific functions with the sole purpose of converting old project files to newer versions.

PR #355, opened 2023-07-15
Linguistic improvements

cmkohnen:
Apologies for the lack of contributions in the last couple of weeks. I've passed my final exams and had quite eventful weeks since then.

Follow HIG advice on text throughout the application. Use correct Typography.
Use "Figure" in favour of "Plot" or "Graph" (only in UI).

### DIFF
--- a/data/ui/window.blp
+++ b/data/ui/window.blp
@@ -17,7 +17,7 @@ template $GraphsWindow : Adw.ApplicationWindow {
         visible: bind-property sidebar_flap.folded inverted;
         MenuButton add_button {
           icon-name: "list-add-symbolic";
-          tooltip-text: _("Add new data");
+          tooltip-text: _("Add new Data");
           always-show-arrow: true;
           menu-model: add_data_menu;
         }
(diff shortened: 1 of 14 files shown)


data/ui/window.blp
```diff
@@ -525,18 +525,18 @@ template $GraphsWindow : Adw.ApplicationWindow {
...
-    item (_("Save Project"), "app.save_project")
-    item (_("Open Project"), "app.open_project")
+    item (_("Save Project…"), "app.save_project")
+    item (_("Open Project…"), "app.open_project")
   }
   section {
-    item (_("Export Data"), "app.export_data")
-    item (_("Export Figure"), "app.export_figure")
+    item (_("Export Data…"), "app.export_data")
+    item (_("Export Figure…"), "app.export_figure")
```
Sjoerd1993, line 533: I had to look up the use of ellipses in the Gnome HIG (the usage is correct here as far as I can tell), just for reference:

`Placeholders can be phrased in the same way as labels. Alternatively, if the text field performs an action, they can be phrased as an instruction, followed by ellipses. For example: “Search for a city…” or “Write a message…”. `
https://developer.gnome.org/hig/patterns/controls/text-fields.html#guidelines

Sjoerd1993 (approved):
My own activity last week has also been on a stand-still. Last weeks I've been switching jobs and have had a week of vacation as well. There's been some useful things already reading up into new material that I want to apply in the future in Graphs as well. Not the least, I will probably start writing unit tests in the future, which could go a long way with preventing regression bugs.

Either way, these are mostly very positive changes. The change in lingo towards the consistent use of graph in particular is a good call.

data/ui/window.blp
cmkohnen: I was basically reading up on the HIGs comments on Wording and Text style, where I found this. Although the reference I used was [this one](https://developer.gnome.org/hig/guidelines/writing-style.html#ellipses), which states, that for each label, etc. that requires further action (selecting a file, adjusting parameters) an ellipsis should be used. Also note that this is the U+2026 Unicode Character instead of three dots.

PR #356, opened 2023-07-16
Use better Error Handling.

cmkohnen:
Instead of catching every possible Error in the caller function, raise
an ImportError with a translatable message with each respective error.


Sjoerd1993 (approved):
Improving the error handling was actually on my list of things that should be done based on recent things I've been working on, but this is a good step forwards there indeed.

PR #357, opened 2023-07-29
Safer Number conversion

cmkohnen:
Instead of using either eval() or float() for converting a string to a float we define a custom function.
This makes use of a regular expression matching any number even respecting scientific notation.

### DIFF
--- a/src/calculation.py
+++ b/src/calculation.py
@@ -3,13 +3,13 @@ import numpy
 from numpy import *
 
 
-def create_dataset(x_start, x_stop, equation, step_size):
+def create_dataset(x_start: float, x_stop: float, equation, step_size: float):
     """
     Create all data set parameters that are required
     to create a new data object
     """
-    datapoints = int(abs(eval(x_start) - eval(x_stop)) / eval(step_size)) + 1
-    xdata = numpy.linspace(eval(x_start), eval(x_stop), datapoints)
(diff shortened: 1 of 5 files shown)


src/calculation.py
```diff
@@ -3,13 +3,13 @@
 from numpy import *
 
 
-def create_dataset(x_start, x_stop, equation, step_size):
+def create_dataset(x_start: float, x_stop: float, equation, step_size: float):
```
Sjoerd1993, line 6: Good with the type hinting, eventually we'd probably want to move to type hinting overall, as well as to give an expected return type. That can be done in a seperate PR, but I don't mind adding these types along thew way as we touch the code.

Sjoerd1993 (approved):
This has been on my list for sooo long, as using evals is a pretty hack solution and not really recommended. 

While Graphs is a pretty benign tool, eval does expose a lot of power to the user. And in theory, when someone installs Graphs from source on a production machine, a user can potentially just execute arbitrary python code. While the real life risk may not be too large (which is why I accepted my temporary solution with evals), it's just bad practice.

I've done some testing in the past, as you can basically whitelist eval functions to only execute specific bits, it never got to a clean solution. This works quite well. And the custom regex function also allows us to expand on this if we would want to accept sines for instance (which feels a bit redundant at the moment).

PR #358, opened 2023-07-31
Refactor clipboard

Sjoerd1993:
This refactors the clipboard into its own class. The main reasoning was that this allowed to save the associated limits for each state to the clipboard state. This way, when undoing a certain action, the axes reset also to the state that was present at the time before that action was performed.

See old behaviour here: https://github-production-user-asset-6210df.s3.amazonaws.com/68477016/247573650-9c9dbaaf-b8d3-459b-8c68-0210445b2c33.webm

And new behaviour in this PR here:
[Screencast from 2023-04-02 21-15-03.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/bbc035d5-ef6b-4310-9951-0f3858b89d8a)

Another thing in the behaviour that has changed is that when opening a project, it automatically reloads the limits that were set during the previous session. See video:

[Screencast from 2023-07-31 12-46-57.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/b9aca9cf-aa0b-4a56-a588-0d8e672b7347)

### DIFF
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -52,7 +52,7 @@ class Canvas(FigureCanvas):
                      self.top_right_axis]:
             axis.get_xaxis().set_visible(False)
             axis.get_yaxis().set_visible(False)
-        self.dummy_toolbar = DummyToolbar(self)
+        DummyToolbar(self)
         self.highlight = Highlight(self)
 
     # Temporarily overwritten function, see
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -136,7 +136,7 @@ class Canvas(FigureCanvas):
         plot_settings.max_right = max(self.right_axis.get_ylim())
 
     def get_limits(self):
-        limits = {
+        return {
             "min_bottom": min(self.axis.get_xlim()),
             "max_bottom": max(self.axis.get_xlim()),
             "min_top": min(self.top_left_axis.get_xlim()),
@@ -146,7 +146,6 @@ class Canvas(FigureCanvas):
             "min_right": min(self.right_axis.get_ylim()),
             "max_right": max(self.right_axis.get_ylim()),
         }
-        return limits
 
     def set_axis_properties(self):
         """Set the properties that are related to the axes."""
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -52,7 +52,7 @@ class Canvas(FigureCanvas):
                      self.top_right_axis]:
             axis.get_xaxis().set_visible(False)
             axis.get_yaxis().set_visible(False)
-        DummyToolbar(self)
+        self.dummy_toolbar = DummyToolbar(self)
         self.highlight = Highlight(self)
 
     # Temporarily overwritten function, see
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -135,6 +135,19 @@ class Canvas(FigureCanvas):
         plot_settings.min_right = min(self.right_axis.get_ylim())
         plot_settings.max_right = max(self.right_axis.get_ylim())
 
+    def get_limits(self):
+        limits = {
+            "min_bottom": min(self.axis.get_xlim()),
+            "max_bottom": max(self.axis.get_xlim()),
+            "min_top": min(self.top_left_axis.get_xlim()),
+            "max_top": max(self.top_left_axis.get_xlim()),
+            "min_left": min(self.axis.get_ylim()),
+            "max_left": max(self.axis.get_ylim()),
+            "min_right": min(self.right_axis.get_ylim()),
+            "max_right": max(self.right_axis.get_ylim()),
+        }
+        return limits
+
     def set_axis_properties(self):
         """Set the properties that are related to the axes."""
         plot_settings = self.application.plot_settings
--- a/src/clipboard.py
+++ b/src/clipboard.py
@@ -61,7 +61,7 @@ class Clipboard:
             self.application.main_window.redo_button.set_sensitive(True)
         graphs.check_open_data(self.application)
         ui.reload_item_menu(self.application)
-        self.application.canvas.dummy_toolbar.push_current()
+        self.application.canvas.toolbar.push_current()
 
     def redo(self):
         """
--- a/src/clipboard.py
+++ b/src/clipboard.py
@@ -61,6 +61,7 @@ class Clipboard:
             self.application.main_window.redo_button.set_sensitive(True)
         graphs.check_open_data(self.application)
         ui.reload_item_menu(self.application)
+        self.application.canvas.dummy_toolbar.push_current()
 
     def redo(self):
         """
@@ -79,6 +80,7 @@ class Clipboard:
             self.application.main_window.redo_button.set_sensitive(False)
         graphs.check_open_data(self.application)
         ui.reload_item_menu(self.application)
+        self.application.canvas.dummy_toolbar.push_current()
 
     def clear(self):
         """Clear the clipboard to the initial state"""
--- a/src/clipboard.py
+++ b/src/clipboard.py
@@ -81,9 +81,7 @@ class Clipboard:
         ui.reload_item_menu(self.application)
 
     def clear(self):
-        """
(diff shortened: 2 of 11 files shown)


cmkohnen:
I feel this is a good change.
However I noticed weird issues with the view back and forth actions on this behaviour.

If you undo an action that changes the view the view gets changed, but the buttons don't update. Performing the "view back" action then just sets the view to the one that might already be the current. 

A solution that would not only fix this bug, but fix the issue, that the view states don't persist on a reload (or even a save to project) would be to implement a similar view_clipboard instead of relying on matplotlib for that and storing a state of that clipboard for each action performed.

Sjoerd1993:
> If you undo an action that changes the view the view gets changed, but the buttons don't update. Performing the "view back" action then just sets the view to the one that might already be the current.

I can see what you mean here. It's not really trivial to save the actual states of the view stack itself onto the clipboard, nor am I sure if I can pickle it and save it in a project. But this behaviour itself seems to be solved by pushing to the view stack upon undo/redo actions. So far I haven't found ill-intended side effects with that solution.

cmkohnen:
> > If you undo an action that changes the view the view gets changed, but the buttons don't update. Performing the "view back" action then just sets the view to the one that might already be the current.
> 
> I can see what you mean here. It's not really trivial to save the actual states of the view stack itself onto the clipboard, nor am I sure if I can pickle it and save it in a project. But this behaviour itself seems to be solved by pushing to the view stack upon undo/redo actions. So far I haven't found unintended side effects with that solution.

I tried that a while ago. The view stack itself stores properties tied to the axes on the current canvas. Since new ones are created every time the figure gets reloaded, it isn't possible to just save the stack property (`toolbar._nav_stack`). This approach does work, but I'd like to try and implement an mpl-independent solution to storing view states, so that they might survive a reload/be saved to a project

src/canvas.py
```diff
@@ -135,6 +135,19 @@ def apply_limits(self):
...
+            "min_bottom": min(self.axis.get_xlim()),
+            "max_bottom": max(self.axis.get_xlim()),
+            "min_top": min(self.top_left_axis.get_xlim()),
+            "max_top": max(self.top_left_axis.get_xlim()),
+            "min_left": min(self.axis.get_ylim()),
+            "max_left": max(self.axis.get_ylim()),
+            "min_right": min(self.right_axis.get_ylim()),
+            "max_right": max(self.right_axis.get_ylim()),
+        }
+        return limits
```
cmkohnen, line 149:
```suggestion
        return {
            "min_bottom": min(self.axis.get_xlim()),
            "max_bottom": max(self.axis.get_xlim()),
            "min_top": min(self.top_left_axis.get_xlim()),
            "max_top": max(self.top_left_axis.get_xlim()),
            "min_left": min(self.axis.get_ylim()),
            "max_left": max(self.axis.get_ylim()),
            "min_right": min(self.right_axis.get_ylim()),
            "max_right": max(self.right_axis.get_ylim()),
        }
```

src/canvas.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, application):
                      self.top_right_axis]:
             axis.get_xaxis().set_visible(False)
             axis.get_yaxis().set_visible(False)
-        DummyToolbar(self)
+        self.dummy_toolbar = DummyToolbar(self)
```
cmkohnen, line 55:
```suggestion
        DummyToolbar(self)
```
This might be intransparent, but upon init, the Toolbar automatically sets itself as the toolbar attribute of the canvas provided, making a second assignment redundant

src/clipboard.py
```diff
@@ -4,60 +4,86 @@
...
+            self.application.datadict = \
+                copy.deepcopy(self.datadict_clipboard[self.clipboard_pos])
+
+        if abs(self.clipboard_pos) >= len(self.datadict_clipboard):
+            self.application.main_window.undo_button.set_sensitive(False)
+        if self.clipboard_pos < -1:
+            self.application.main_window.redo_button.set_sensitive(True)
+        graphs.check_open_data(self.application)
+        ui.reload_item_menu(self.application)
+        self.application.canvas.dummy_toolbar.push_current()
```
cmkohnen, line 64:
```suggestion
        self.application.canvas.toolbar.push_current()
```

src/clipboard.py
```diff
@@ -4,60 +4,86 @@
...
+                copy.deepcopy(self.datadict_clipboard[self.clipboard_pos])
+            self.application.canvas.set_limits(
+                self.limits_clipboard[self.clipboard_pos])
+            self.application.main_window.undo_button.set_sensitive(True)
+
+        if self.clipboard_pos >= -1:
+            self.application.main_window.redo_button.set_sensitive(False)
+        graphs.check_open_data(self.application)
+        ui.reload_item_menu(self.application)
+        self.application.canvas.dummy_toolbar.push_current()
```
cmkohnen, line 83:
```suggestion
        self.application.canvas.toolbar.push_current()
```

src/canvas.py
Sjoerd1993: Thanks, I was slightly unsure if the attribute would persist, but it's good without the explicit assignment then.

src/file_io.py
```diff
@@ -35,10 +39,11 @@ def save_project(file, plot_settings, datadict, datadict_clipboard,
 
 def read_project(file):
     project = pickle.loads(_read_file(file, None))
+    project_items = ["plot_settings", "data", "clipboard", "version"]
+    project.extend({key, None for key in project_items if key not in project})
```
cmkohnen, line 43:
```suggestion
    project.extend({key: None for key in project_items if key not in project})
```
Oops, my bad

Sjoerd1993: Yep, didn't notice it immediately either, had it already fixed in my local version.

cmkohnen (approved):

Sjoerd1993:
> I tried that a while ago. The view stack itself stores properties tied to the axes on the current canvas. Since new ones are created every time the figure gets reloaded, it isn't possible to just save the stack property (`toolbar._nav_stack`). This approach does work, but I'd like to try and implement an mpl-independent solution to storing view states, so that they might survive a reload/be saved to a project

You mean we'd rewrite the view-switcher itself entirely to it's own clipboard and use that instead?

That would likely be possible, and make things a bit more approachable for our ends. It would probably work best with it's own separate `view_clipboard`, although we can probably create a general clipboard class to reduce some repetition and inherit from there.

The main challenge I see, is that we'd like to push a new state into the clipboard each time the canvas is zoomed/panned/changed. I'm not sure if there's a signal that's emitted that catches all those situations, without adding a whole bunch of redundant states as well. Perhaps `draw_event` would get us pretty close though. It would maybe go somewhat beyond the scope of this PR, but it's definitely something to investigate.

cmkohnen:
> You mean we'd rewrite the view-switcher itself entirely to it's own clipboard and use that instead?
> 
> That would likely be possible, and make things a bit more approachable for our ends. It would probably work best with it's own separate `view_clipboard`, although we can probably create a general clipboard class to reduce some repetition and inherit from there.
> 
> The main challenge I see, is that we'd like to push a new state into the clipboard each time the canvas is zoomed/panned/changed. I'm not sure if there's a signal that's emitted that catches all those situations, without adding a whole bunch of redundant states as well. Perhaps `draw_event` would get us pretty close though. It would maybe go somewhat beyond the scope of this PR, but it's definitely something to investigate.

That was the idea, although I haven't thought about an implementation thoroughly. When panning/zooming, the `toolbar.push_current()` function is called, which is responsible for pushing the current limits to the view stack. We can simply override this function to hook into our own implementation.

Sjoerd1993:
> When panning/zooming, the `toolbar.push_current()` function is called,

That's fantastic. It's called every single time to push the canvas into the stack. Well, it doesn't do on undo/redo actions actually, but we can call them manually then as in this current PR. This makes the further implementation not all that complicated.

I'll probably take a look at that tomorrow morning in a separate PR.

PR #360, opened 2023-07-31
use ast for evaluating mathematical expressions

cmkohnen:


src/utilities.py
```diff
@@ -256,22 +257,20 @@ def check_item(self, item):
...
-        return number
-    pi_pattern = re.compile(
-        # match pi with optional factor (e. g. "2.4pi")
-        r"(?P<f>[+\-]?(?:0|[1-9]\d*)(?:\.\d*)?)?(?:pi|π)$")
-    pi_result = pi_pattern.search(string)
-    if pi_result is not None:
-        factor = pi_result.group("f")
-        return numpy.pi if factor is None else numpy.pi * float(factor)
-    raise ValueError(_("No valid number specified"))
+    return _eval(ast.parse(string, mode="eval").body)
```
Sjoerd1993, line 260: A quick and dirty way to still allow the usage of pi, as well as to call powers with a ^. (The vast majority will never use the bitwise ^, but instead will use the ^ symbol for powers).


```suggestion
    string = string.replace("pi", f"({float(numpy.pi)})")
    string = string.replace("^", "**")
    return _eval(ast.parse(string, mode="eval").body)
```

cmkohnen: Ah, using brackets works well. I thought of a solution replacing it with `"*" + numpy.pi`, which resulted in errors when no factor was given, this is neat

Sjoerd1993 (approved):

PR #361, opened 2023-07-31
use custom function for converting strings to float during import

cmkohnen:
Allows to express arithmetic operations in column files


Sjoerd1993 (approved):
It's probably rare to express a column like this, but why not. It makes importing even more robust, and doesn't really complicate the code.

Works well in my testing.

PR #363, opened 2023-08-01
Rework clipboard and view navigation

Sjoerd1993:
Reworks the clipboard as well as the view navigation. 

The clipboard is now one class, where two subclasses are used that inherit from the base class, one for the data and one for the view. This allows us to couple the view navigation to this custom clipboard instead of using the navigationtoolbar build-in. This also allows us to store the view clipboard in project files, and load the latest view upon loading in a less hacky way. 

The other thing that has been changed in behaviour, is that undo/redo now sets the views to whatever the status was on the previous/next action. Any suggestions on what exact behaviour would be preferred are always welcome :)

Also, there's some potential for further refactors into the clipboard classes eventually. There's still some duplication going on.

### DIFF
--- a/src/clipboard.py
+++ b/src/clipboard.py
@@ -39,7 +39,6 @@ class DataClipboard(BaseClipboard):
         Appends the latest state to the clipboard.
         """
         super().add(copy.deepcopy(self.application.datadict))
-
         # Keep clipboard length limited to preference values
         if len(self.clipboard) > \
                 int(self.application.preferences["clipboard_length"]) + 1:
@@ -63,7 +62,8 @@ class DataClipboard(BaseClipboard):
                 self.application.main_window.redo_button.set_sensitive(True)
             graphs.check_open_data(self.application)
             ui.reload_item_menu(self.application)
-            self.application.ViewClipboard.undo()
+            if self.application.ViewClipboard.view_changed:
+                self.application.ViewClipboard.undo()
 
     def redo(self):
 
@@ -82,8 +82,8 @@ class DataClipboard(BaseClipboard):
             self.application.main_window.redo_button.set_sensitive(False)
         graphs.check_open_data(self.application)
         ui.reload_item_menu(self.application)
-        self.application.ViewClipboard.redo()
-
+        if self.application.ViewClipboard.view_changed:
+            self.application.ViewClipboard.redo()
 
     def clear(self):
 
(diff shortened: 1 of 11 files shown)


Sjoerd1993:
Changed it a bit more so that the associated view-state (canvas limits) are stored on each datadict item. This way, when performing an undo/redo action, you are gauranteed to hit the canvas state that corresponds to the previous state in datadict. 

Initial commit broke apart when performing an action, then changing the view, then pressing undo. As it would just revert to the previous view when pressing undo.

src/clipboard.py
```diff
@@ -1,89 +1,145 @@
...
     def __init__(self, application):
         self.application = application
-        self.datadict_clipboard = [{}]
-        self.limits_clipboard = [self.application.canvas.get_limits()]
+        self.clipboard = []
+        self.clipboard_pos = -1
+
+    def add(self, new_state):
+        self.undo_button.set_sensitive(True)
+        # If a couple of redo"s were performed previously, it deletes the
```
cmkohnen, line 13:
```suggestion
        # If a couple of redo's were performed previously, it deletes the
```

src/clipboard.py
```diff
@@ -1,89 +1,145 @@
-# SPDX-License-Identifier: GPL-3.0-or-later
 import copy
-
 from graphs import graphs, ui
 
 
-class Clipboard:
+class BaseClipboard:
     def __init__(self, application):
```
cmkohnen, line 6: I feel like both undo and redo buttons should be passed along to the base class instead of being defined in each inheriting class

src/item_box.py
```diff
@@ -81,6 +82,7 @@ def on_color_dialog_accept(self, dialog, result):
                 self.item.alpha = color.alpha
                 self.update_color()
                 self.application.Clipboard.add()
+                self.application.ViewClipboard.add()
```
cmkohnen, line 85: Changing the color of an item doesn't change the view, so this might not be needed

cmkohnen (commented):
Toyed around with this for a bit. The same issue I've mentioned previously still prevails. I think a better solution would be storing the whole view_clipboard with each undo/redo state

Sjoerd1993:
> Toyed around with this for a bit. The same issue I've mentioned previously still prevails. I think a better solution would be storing the whole view_clipboard with each undo/redo state

That was a bug, should be fixed now. The view is stored for each datadict state right now (which would be the undo/redo state). Hence the DataClipboard clipboard attribute being a list of dictionaries containing both the datadict itself, and the view at the time of that specific state.

cmkohnen (approved):
I'll probably do a small refactor soon regarding logic in the toolbar base class, which is now redundant and only wastes memory/cpu power, but this functionality works fine in my testing

PR #366, opened 2023-08-01
use numexpr for calculating functions

cmkohnen:
Introduces a new dependency, which can evaluate mathematical expressions on arrays, getting rid of yet another use of eval.


Sjoerd1993 (approved):
Generally I tried to avoid adding new dependencies, but having straight evals should be avoided even harder. Works as expected in my testing

PR #367, opened 2023-08-01
use ellipsis

cmkohnen:


Sjoerd1993 (approved):
Given it's actually unicode, it's porbably for the better

PR #368, opened 2023-08-01
Rafactor.

cmkohnen:
No longer call toolbars inbuilt functions to store view states. Rely on own mode implementation inside toolbar logic.

fix linting.

tox.ini: allow W503

Use @property decorator for setting and getting canvas limits

### DIFF
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -11,7 +11,7 @@ from graphs.item import Item, TextItem
 from graphs.misc import InteractionMode
 from graphs.rename import RenameWindow
 
-from matplotlib import backend_tools as tools, cbook, pyplot
+from matplotlib import backend_tools as tools, pyplot
 from matplotlib.backend_bases import NavigationToolbar2
 from matplotlib.backends.backend_gtk4cairo import FigureCanvas
 from matplotlib.figure import Figure
@@ -38,7 +38,7 @@ class Canvas(FigureCanvas):
         color_rgba.alpha = 0.3
         self.rubberband_fill_color = utilities.rgba_to_tuple(color_rgba, True)
         super().__init__(self.figure)
-        self.set_limits({
+        self.limits = {
             "min_bottom": self.application.plot_settings.min_bottom,
             "max_bottom": self.application.plot_settings.max_bottom,
             "min_top": self.application.plot_settings.min_top,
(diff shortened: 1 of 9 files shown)


src/canvas.py
```diff
@@ -273,6 +276,37 @@ def set_legend(self):
...
+                self.canvas.set_cursor(tools.Cursors.SELECT_REGION)
+                self._last_cursor = tools.Cursors.SELECT_REGION
+            elif (mode == InteractionMode.PAN
+                  and self._last_cursor != tools.Cursors.MOVE):
+                self.canvas.set_cursor(tools.Cursors.MOVE)
+                self._last_cursor = tools.Cursors.MOVE
+        elif self._last_cursor != tools.Cursors.POINTER:
+            self.canvas.set_cursor(tools.Cursors.POINTER)
+            self._last_cursor = tools.Cursors.POINTER
+
```
Sjoerd1993, line 309: What do these two overwritten function change compared to their original counterparts?

src/canvas.py
```diff
@@ -287,20 +321,9 @@ def remove_rubberband(self):
 
     # Overwritten function - do not change name
     def push_current(self):
-        super().push_current()
         self.canvas.apply_limits()
```
Sjoerd1993, line 324: Yeah running the base class version is just a waste of resources at this point

tox.ini
```diff
@@ -1,7 +1,7 @@
 [flake8]
 count = True
 #default_ignorelist: E121,E123,E126,E226,E24,E704,W503,W504
-ignore = U101, SCS109, SCS101, SCS113
+ignore = U101, SCS109, SCS101, SCS113, W503
```
Sjoerd1993, line 4: Yeah we should either disable W503 or W504, as they are kinda in direct conflict with each other (they describe each others best practice as antipattern and vice versa)

https://www.flake8rules.com/rules/W503.html
https://www.flake8rules.com/rules/W504.html

Not sure if I have a very strong preference tbh

Sjoerd1993 (approved):

src/canvas.py
cmkohnen: The base implementation checks against the mode stored inside the toolbar itself. The difference here is, that we check against InteractionMode stored inside the main application, deduplicating logic

Sjoerd1993 (approved):

PR #369, opened 2023-08-01
Refactor clipboard

Sjoerd1993:
Refactors the new clipboard a bit, moves a bit more logic into the base class in a deduplication effort, and also fixes the `ui.set_clipboard_buttons` function which was not working properly (at least now when opening saved projects).

### DIFF
--- a/src/clipboard.py
+++ b/src/clipboard.py
@@ -21,9 +21,35 @@ class BaseClipboard:
         self.clipboard.append(new_state)
         self.redo_button.set_sensitive(False)
 
+    def undo(self):
+        if abs(self.clipboard_pos) < len(self.clipboard):
+            self.clipboard_pos -= 1
+            self.set_clipboard_state()
+
+        if abs(self.clipboard_pos) >= len(self.clipboard):
+            self.undo_button.set_sensitive(False)
+        if self.clipboard_pos < -1:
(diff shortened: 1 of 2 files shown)


src/clipboard.py
```diff
@@ -21,10 +21,36 @@ def add(self, new_state):
...
+        """
+        Redo an action, moves the clipboard position forwards by one and
+        changes the dataset to the state before the previous action was undone
+        """
+        if self.clipboard_pos < -1:
+            self.clipboard_pos += 1
+            self.set_clipboard_state()
+            self.undo_button.set_sensitive(True)
+
+        if self.clipboard_pos >= -1:
```
cmkohnen, line 44:
```suggestion
        else:
```

src/clipboard.py
```diff
@@ -21,10 +21,36 @@ def add(self, new_state):
         self.clipboard.append(new_state)
         self.redo_button.set_sensitive(False)
 
+    def undo(self):
+        if abs(self.clipboard_pos) < len(self.clipboard):
+            self.clipboard_pos -= 1
+            self.set_clipboard_state()
+
+        if abs(self.clipboard_pos) >= len(self.clipboard):
```
cmkohnen, line 29:
```suggestion
        else:
```

cmkohnen (approved):

PR #370, opened 2023-08-01
use numexpr for performing custom transformation

cmkohnen:


Sjoerd1993:
Codewise it looks fine, and a bit cleaner even than the current implementation. There's a few issues with the limitation. 
Right now, using numpy we can convert a parameter from degrees to radians using radians(X). 

So if I have some coordinates in degrees, and I want to take the sine of this. I can do sin(radians(X)). Which by the way is ugly and unintuitive for regular users, but it works quite well. The functionality itself is something I'd like to have though, and I'll try and think of a way to just convert sind(X) to do the same. So basically dsin(X) should evaluate to sin((pi*X)/180). Same for cosd(X).

The other feature that it breaks is the ability to do list comprehensions. This is something very niche and frankly undiscoverable, so I'm probably the only person who ever did that in Graphs. But I've used it for example to add the smallest non-zero element to an actual data-set as a form of log-smoothening: `Y = Y + min([value for value in Y_range if value != 0])`

However, this too is pretty much undiscoverable, and I'd like to think of a better, more intuitive, way to add in this functionality. I'll probably submit a PR on a later date to expose a filter to the user. Perhaps call it using something custom like `exclude(value, list)`.  So they can for example call something like: Y = Y + min(exclude(0,Y)) to add the minimum value that is not 0. Other custom commands can be defines as well that way in the future.

PR #372, opened 2023-08-02
Refactor plot_settings window

cmkohnen:
Make style change instantaneous and refactor the whole class slightly.

### DIFF
--- a/src/plot_settings.py
+++ b/src/plot_settings.py
@@ -33,12 +33,10 @@ class PlotSettingsWindow(Adw.PreferencesWindow):
     max_top = Gtk.Template.Child()
     no_data_message = Gtk.Template.Child()
 
-    @property
-    def plot_settings(self):
-        return self.props.application.plot_settings
-
     def __init__(self, application):
         super().__init__(application=application)
+        self.plot_settings = self.props.application.plot_settings
+
         self.plot_title.set_text(self.plot_settings.title)
         self.min_left.set_text(str(self.plot_settings.min_left))
         self.max_left.set_text(str(self.plot_settings.max_left))
--- a/src/plot_settings.py
+++ b/src/plot_settings.py
@@ -1,8 +1,8 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 from gi.repository import Adw, Gtk
 
-from graphs import (file_io, graphs, misc, plot_styles,
-                    plotting_tools, ui, utilities)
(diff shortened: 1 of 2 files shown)


src/plot_settings.py
```diff
@@ -143,24 +133,38 @@ def get_float(entry):
...
-            for item in self.props.application.datadict.values():
-                item.color = None
-            for item in self.props.application.datadict.values():
-                item.color = \
-                    plotting_tools.get_next_color(self.props.application)
+        graphs.refresh(self.props.application)
+        self.props.application.Clipboard.add()
+        self.props.application.ViewClipboard.add()
+
+    @Gtk.Template.Callback()
```
Sjoerd1993, line 140: I'm loving the change to do this with the callback, instead of trying to detect if the used style is different from the set style on closing. Apart from the user-experience (it's just nice to see it changing immediately),  it's so much cleaner code-wise as well.

src/plot_settings.py
```diff
@@ -33,45 +33,50 @@ class PlotSettingsWindow(Adw.PreferencesWindow):
     max_top = Gtk.Template.Child()
     no_data_message = Gtk.Template.Child()
 
+    @property
+    def plot_settings(self):
+        return self.props.application.plot_settings
```
Sjoerd1993, line 38: Any particular reason to define the `plot_settings` as property of the class like this, instead of retrieving it as an attribute using `self.props.application.plot_settings`?

Sjoerd1993: Yeah I was just wondering if it perhaps isn't cleaner to just call `self.plot_settings = self.props.application.plot_setting
` in the init instead of having it as a property using the decorator, in particular since it's not really a fixed/read-only property of the class itself but rather an external attribute of the application that we're modifying. 
On the other hand, in the end, the property is kinda still static here, in the sense that it's still the same reference to the same attribute. That is still read-only, so in that argument using @property maybe does make sense. But my initial instinct would be to just go for an attribute directly. 

Either would be better than calling `plot_settings = self.props.application.plot_settings` each time by the way, it's reused throughout the class often enough that it should be defined as instance of the class.

cmkohnen: It was just initial instinct to do it this way, but moving it to init works just was well, I don't believe there is a measurable difference in memory consumtion/cpu usage either way

Sjoerd1993 (approved):
You're likely right about it not making a difference performance wise by the way. In both cases self.plot_settings points to the same object in self.application, I don't think the property method changes this deeply.

PR #375, opened 2023-08-03
Refactor preferences

cmkohnen:
Dynamically populate ui config items and retrieve values if possible.
Makes adding and removing config items easier and reduces visual
clutter.


src/canvas.py
```diff
@@ -267,7 +267,7 @@ def set_legend(self):
             if labels:
                 self.top_right_axis.legend(
                     new_lines, labels,
-                    loc=self.application.plot_settings.legend_position,
+                    loc=self.application.plot_settings.legend_position.lower(),
```
Sjoerd1993, line 270: Subtle, but indeed cleaner to do this here in canvas.

src/preferences.py
```diff
@@ -7,6 +7,13 @@
 from graphs import file_io, graphs, misc, plot_styles, utilities
 
 
+MIGRATION_KEYS = [
+    # old -> new
+    ("handle_duplicates", "other_handle_duplicates"),
+    ("hide_unselected", "other_hide_unselected"),
+]
```
Sjoerd1993, line 14: This can be useful in the longer term. The current code handles outdated keys relatively well by simply using the template value, but this does keep user settings set at least as long as the migration key is defined.

src/graphs.py
```diff
@@ -152,8 +152,10 @@ def refresh(self):
...
     self.canvas.set_ticks()
     for item in reversed(self.datadict.values()):
-        if (item is not None) and not \
-                (self.preferences["hide_unselected"] and not item.selected):
-            self.canvas.plot(item)
+        if item is None \
+                or (self.preferences["other_hide_unselected"]
+                    and not item.selected):
+            continue
+        self.canvas.plot(item)
```
Sjoerd1993, line 159: I think these should eventually do the same. Don't really have a strong preference either way, they both work afaik and are readable to me.

src/preferences.py
```diff
@@ -136,82 +153,72 @@ def __init__(self, application):
...
 
     def load(self):
-        preferences = self.props.application.preferences
-        self.clipboard_length.set_value(int(preferences["clipboard_length"]))
-
-        columns_params = preferences["import_params"]["columns"]
+        for key, value in self.props.application.preferences.items():
+            if key in CONFIG_IGNORELIST:
+                continue
+            try:
```
Sjoerd1993, line 159: These loops are pretty nice, indeed removes a lot of clutter

Sjoerd1993 (approved):

src/canvas.py
cmkohnen: The reason I've moved it to here is, that previously it was in the huge block of code, that loads and saves the preferences to ui.
The alternative was making an exception there and handling this key manually, but I though this was cleaner. Config keys should work just fine, as "best" isn't an option anymore, but "Best" is still the first choice.

src/graphs.py
cmkohnen: Main reason here was that the config key was too long to fit anymore, so made multiple attempts in rewriting ending up with this one.

Sjoerd1993 (changes requested):

Sjoerd1993 (approved):

PR #376, opened 2023-08-03
Refactor preferences

cmkohnen:
Handle key migration and template checking in one step. Also migrate
import_params if it were to change.

### DIFF
--- a/src/preferences.py
+++ b/src/preferences.py
@@ -15,7 +15,8 @@ MIGRATION_KEYS = {
 
 
 def _validate(config: dict, template: dict):
-    """Validates a given dictionary against a template, such that they
+    """
+    Validates a given dictionary against a template, such that they
     remain compatible with updated versions of Graphs. If the key in the
     dictionary is not present in the template due to an update, the key will
     be updated using MIGRATION_KEYS.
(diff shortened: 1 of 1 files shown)


src/preferences.py
```diff
@@ -8,15 +8,17 @@
...
+    # new: old
+    "other_handle_duplicates": "handle_duplicates",
+    "other_hide_unselected": "hide_unselected",
 }
 
 
-def _check_against_template(config: dict, template: dict):
-    return {key: config[key] if key in config else value
-            for key, value in template.items()}
+def _validate(config: dict, template: dict):
```
Sjoerd1993, line 17:
```suggestion
def _validate(config: dict, template: dict):
    """Validates a given dictionary against a template, such that they
    remain compatible with updated versions of Graphs. If the key in the
    dictionary is not present in the template due to an update, the key will
    be updated using MIGRATION_KEYS.

    If the template or validated key does not match with the MIGRATION_KEYS,
    the template keys and their associated values will be used instead for
    the validated dictionary.

    Args:
        config: Dictionary to be validated
        template: Template dictionary to which the config is validated against
    Returns:
        dict: Validated dictionary
    """
```

I had to think quite a bit before I fully understood what is happening here, but it's a very compact and clever way to validate these templates. Nice job. I did write a docstring in order to clarify this a bit more, feel free to make some changes to my suggested docstring if you have a better description. On the longer term, I want to add docstrings to other functions anyway.

Sjoerd1993 (changes requested):

Sjoerd1993 (approved):

PR #377, opened 2023-08-03
Refactor plot style editor.

cmkohnen:
Apply new logic for editing plot styles.

### DIFF
--- a/src/plot_styles.py
+++ b/src/plot_styles.py
@@ -108,6 +108,37 @@ def get_style(self, stylename):
     return style
 
 
+STYLE_DICT = {
+    "linestyle": ["lines.linestyle"],
+    "linewidth": ["lines.linewidth"],
+    "markersize": ["lines.markersize"],
+    "tick_direction": ["xtick.direction", "ytick.direction"],
+    "minor_ticks": ["xtick.minor.visible", "ytick.minor.visible"],
+    "major_tick_width": ["xtick.major.width", "ytick.major.width"],
+    "minor_tick_width": ["xtick.minor.width", "ytick.minor.width"],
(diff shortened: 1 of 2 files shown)


src/ui.py
```diff
@@ -161,6 +161,10 @@ def load_values_from_dict(window, values, ignorelist=None):
                 widget.set_active(bool(value))
             elif isinstance(widget, Adw.ExpanderRow):
                 widget.set_enable_expansion(bool(value))
+            elif isinstance(widget, Gtk.Scale):
```
Sjoerd1993, line 164: Generally, a whole bunch of "if isintance" checks are not really a great practice, but in this case I think it makes sense. And again, a big improvement over settings all possible properties by hand, this really simplifies things a lot.

src/plot_styles.py
```diff
@@ -253,105 +256,51 @@ def load_style(self):
...
-        style["xtick.labelsize"] = font_size
-        style["ytick.labelsize"] = font_size
-        style["axes.titlesize"] = font_size
-        style["legend.fontsize"] = font_size
-        style["figure.titlesize"] = font_size
-        style["figure.labelsize"] = font_size
+        for key in ["font.size", "axes.labelsize", "xtick.labelsize",
+                    "ytick.labelsize", "axes.titlesize", "legend.fontsize",
+                    "figure.titlesize", "figure.labelsize"]:
+            self.style[key] = font_size
```
Sjoerd1993, line 280: Not really that different from the previous logic, but definitely more readable. Was told the other day something along the lines of "when you're declaring a whole load of similar stuff like that, you basically know you should be using some loop or comprehension". Which kinda applies to this.

Sjoerd1993 (approved):
Cleans the old style editor quite nicely.

PR #379, opened 2023-08-04
Use more GObject-Style code

cmkohnen:
- inline transient_for property
- use gproperty for some classes
- refactor canvas (use super for creating self.figure)
- drop useless span argument in Highlight constructor
- refactor AboutWindow

Using GObject based properties gives the added benefit of having some sort of typesafety and generifying object creation.

### DIFF
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -4,7 +4,7 @@ import time
 from contextlib import nullcontext
 from gettext import gettext as _
 
-from gi.repository import Gtk
+from gi.repository import Adw, GObject, Gtk
 
 from graphs import utilities
 from graphs.item import Item, TextItem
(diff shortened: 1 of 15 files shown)


src/clipboard.py
```diff
@@ -1,13 +1,21 @@
...
+
 from graphs import graphs, ui
 
 
-class BaseClipboard:
-    def __init__(self, application):
-        self.application = application
-        self.clipboard = []
-        self.clipboard_pos = -1
+class BaseClipboard(GObject.Object):
```
Sjoerd1993, line 8: Any reason to let the Clipboard specifically inherit from GObject?

src/canvas.py
```diff
@@ -349,8 +349,6 @@ def __init__(self, canvas, span=None):
             interactive=True,
             drag_from_anywhere=True,
         )
-        if span is not None:
```
Sjoerd1993, line 352: Pretty sure this was legacy code, indeed no longer used.

Sjoerd1993 (commented):
Hopefully I have the time tomorrow to take a closer look then, otherwise that may be on sunday.

src/clipboard.py
cmkohnen: Since we currently only ever pickle the `clipboard` and `clipboard_pos` attribute which are a dict and an int respecively that isn't an issue.

Sjoerd1993: I was just thinking that `clipboard = GObject.Property(type=object)` kinda changes the type. 

I'm perfectly fine merging this, as it's a good change overall. I was just wondering if there's any benefit to having the Clipboard class inheriting from GObject as well, as opposed to keeping it "Vanilla Python".

cmkohnen: Technically it kind of does. However during get/set pygobject automatically handles the conversion for us.

As for benefit, I was planning an upcomming pr making use of GObjects Signal functionality and maybe also a huge refactor involving items trying to reduce unnecessary refreshs (also making use of some of this functionality).
The other reason is consistency. GObject has this really nic feature, where you define all properties with default values and can just pass them along during init. This is already used by pretty much all classes that depend on it, but isn't by the custom ones we created.

PR #381, opened 2023-08-04
expand ExpanderRows automatically

cmkohnen:
Default behavior when toggling an ExpanderRow expands it.
This way, when the initial state is enabled, it gets automatically expanded as well.

### DIFF
--- a/src/ui.py
+++ b/src/ui.py
@@ -161,6 +161,7 @@ def load_values_from_dict(window, values, ignorelist=None):
                 widget.set_active(bool(value))
             elif isinstance(widget, Adw.ExpanderRow):
                 widget.set_enable_expansion(bool(value))
+                widget.set_expanded(bool(value))
             elif isinstance(widget, Gtk.Scale):
                 widget.set_value(value)
             elif isinstance(widget, Gtk.Button):


Sjoerd1993 (approved):
It works well, and makes sense to expose the drop down option when the setting is turned on. The only thing I'm wondering is if these expanderrows should be expanded by default, also when opening the window initially and the toggle is turned on

cmkohnen:
> The only thing I'm wondering is if these expanderrows should be expanded by default, also when opening the window initially and the toggle is turned on

That is the whole change of this pr (expanding when the user toggles is default Gtk behaviour).
If feel it might be inconsistent if I enable e.g. the legend in settings (again), the expanderrow extends, I reopen the window and it is no longer extended

Sjoerd1993:
Ah, you are correct this this is standard behaviour. Didn't even notice that. But your second line was exactly what I meant, in this PR for me, if an expander row is enabled already when opening a window it is not extended. It may be more consistent to have it extended in that case as well.
[Screencast from 2023-08-06 08-44-35 (trimmed).webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/d2f274d4-593c-423a-b30b-9ffcc398c4b9)

cmkohnen:
The plot settings aren't yet adapted to the new logic, I've introduced with plot styles and preferences, That is why it won't work there.

Sjoerd1993:
Yeah I see now, indeed the plot settings don't follow then new logic yet for setting up the UI. Apologies for not looking thoroughly here.

PR #383, opened 2023-08-05
UI changes

cmkohnen:
implement changes based on recent [Design suggestions](https://gitlab.gnome.org/Teams/Circle/-/issues/185).

- Reword some items
- Remove indifferent descriptions
- Move Translate & Multiply to the bottom
- Redesign Translate and Multiply actions

### DIFF
--- a/data/ui/add_style.blp
+++ b/data/ui/add_style.blp
@@ -42,12 +42,11 @@ template $AddStyleWindow : Adw.Window {
 	      margin-end: 26;
         Adw.ComboRow plot_style_templates {
           title: _("Template");
-          subtitle: _("Template style used as a base");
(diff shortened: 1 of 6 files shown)


Sjoerd1993:
Very nice, looking a lot more clean as well actually when going into the settings.
I think most of the subtitles in "Import Settings" in the Preferences window are redundant as well actually.

Sjoerd1993 (approved):

cmkohnen:
> I think most of the subtitles in "Import Settings" in the Preferences window are redundant as well actually.

Yes they absolutely are. More so are the settings themselves. I haven't touched them because I anticipated, we'd remove them.

PR #386, opened 2023-08-07
Remove clipboard length option

Sjoerd1993:
Removes the maximum clipboard length option, and defaults to a standard limit of 75 instead. It's extremely rare for a user to actually want to change the setting.

The limit of 75 was taken as a bit of a compromise of being long enough that it should never really be a big issue, but still limited so that the amount of data stored in memory is somewhat limited. I'm happy to discuss the exact value of this limit though.

### DIFF
--- a/data/config.json
+++ b/data/config.json
@@ -4,7 +4,7 @@
   "addequation_step_size": "0.01",
   "addequation_x_start": "0",
   "addequation_x_stop": "10",
-  "clipboard_length": 25,
+  "clipboard_length": 75,
   "export_figure_dpi": 100,
   "export_figure_filetype": "svg",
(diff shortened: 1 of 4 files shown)


Sjoerd1993:
Slight change from initial PR, the clipboard length is still retrieved from the config file instead of it being hard-coded, so that it easily can be changed in a "hidden way". But the setting is just not exposed to the user.

cmkohnen (changes requested):
I can't comment on the specific line, but if we keep it as a hidden option, it should be added to the config ignorelist. (l. 82 in preferences.py)

cmkohnen (approved):

PR #388, opened 2023-08-07
Draft: remove preferences

cmkohnen:


data/ui/import.blp
```diff
@@ -38,61 +38,70 @@ template $ImportWindow : Adw.Window {
...
+              wrap: true;
+              adjustment: Adjustment {
+                step-increment: 1;
+                upper: 999;
+              };
+            }
+          }
+        }
+
+        CheckButton save_values {
```
Sjoerd1993, line 102: We probably want to use a Switch here instead of a toggle.

Also, the label should be a bit more descriptive I think. Like "Remember the above options" or something similar

data/ui/preferences.blp
```diff
@@ -10,39 +10,6 @@ template $PreferencesWindow : Adw.PreferencesWindow {
     icon-name: "applications-system-symbolic";
     title: _("General");
```
Sjoerd1993, line 12: The top page of the preference window does look slightly empty now. But I have not a super great suggestion to make that much better. I guess we can keep that as-is for now.

Sjoerd1993 (changes requested):
Mostly the "Remember button" needs some minor touch ups. It'd probably be better with a Switch, which would line up more with the GNOME HIG as well: https://developer.gnome.org/hig/patterns/controls/switches.html

And a slightly more verbose label would be prefered as well. Technically, I haven't noticed any bugs in my testing associated with this PR.

data/ui/import.blp
cmkohnen: I haven't given much thought to the actual UI, as I wanted to get the technical side correctly first.

I don't feel a switch would be the better choice here, as having a checkbox helps distinguishing it from actual parameters.

I'd change the UI side in a separate pr, which tackles a few other issues as well.

data/ui/preferences.blp
cmkohnen: My thought was to have this page be the second and move the action preference. But thats more suited for another pr imo.

Sjoerd1993: Yeah I was thinking this menu now needs a bit of a redesign anyway, but agreed that this should be done in a seperate PR.

data/ui/import.blp
Sjoerd1993: I've done quite some testing on the technical side of this PR, and no issues there. Seems to work exactly as intended.

Regarding the toggle/checkbox, my personal preference definitely goes to a switch here. A toggle quite clearly symbolizes the "on/off" state, and it's in my view just as distinct as a checkbox (neither of these are part of the other settings either). 

Just for a third opinion, I asked for advice in the GNOME Design Matrix channel, presenting the following four options.

image

PR #391, opened 2023-08-08
Add set to default options in import/export dialogs

Sjoerd1993:
Instead of having a "Remember these settings" option, this just skips that option but instead uses a "Set to defaults" button instead. Also adds this option to the "Export figure" and "Add Equation" dialogs.


cmkohnen:
I don't know if this is the pattern we want to go with. Having this in every single dialog feels weird.
My initial thought would be to place a single reset button in the actual preferences.

Sjoerd1993:
> I don't know if this is the pattern we want to go with. Having this in every single dialog feels weird. My initial thought would be to place a single reset button in the actual preferences.

A single reset button in the actual preferences would work. It does make it less discoverable though, and a bit more inconvenient to get back when things are out of whack. I do agree that it is cleaner though, especially visually. 

I've put the suggestion in the mentioned Matrix as well. Not that I'm blindly steering on the opinions of the people over there, but there's quite some people there with a good sense of user design. In particular in the context of the GNOME HIG.

cmkohnen:
Okay, I'm down with that.

However I don't feel like both export figure and add equation do need such a reset. As Export figure literally has only 3 options, which can't really break either and the add equation is quite intuitive.

cmkohnen (changes requested):

Sjoerd1993:
> However I don't feel like both export figure and add equation do need such a reset. As Export figure literally has only 3 options, which can't really break either and the add equation is quite intuitive.

I think that's sensible to not neccesarily have these options in those dialogs. Generally in `Export Figure` you just want to change one or two settings anyway, which is literally fewer clicks than "Reset to defaults" options at the moment.

Also, the main concern is with importing data, where settings are a bit more brittle. It's pretty easy to set it such that the functionality itself does not work, and then a user may be a bit lost.

data/config.json
```diff
@@ -6,7 +6,7 @@
   "addequation_x_stop": "10",
   "clipboard_length": 75,
   "export_figure_dpi": 100,
-  "export_figure_filetype": "svg",
+  "export_figure_filetype": "png",
```
cmkohnen, line 9: Just a question: any reason we set png as the now preferred default?

cmkohnen (approved):

PR #395, opened 2023-08-09
Ui Changes.

cmkohnen:
- Redesign preferences
- Remove redundant subtitles
- Linguistic/Contextual improvements

### DIFF
--- a/data/ui/preferences.blp
+++ b/data/ui/preferences.blp
@@ -11,7 +11,7 @@ template $PreferencesWindow : Adw.PreferencesWindow {
       title: _("General");
 
       Adw.ComboRow general_center {
-        title: _("Center Data Behaviour");
+        title: _("Center Action Behaviour");
         model: StringList{
           strings [_("Center at maximum Y value"), _("Center at middle coordinate")]
         };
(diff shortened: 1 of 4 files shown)


data/ui/preferences.blp
```diff
@@ -136,20 +122,19 @@ template $PreferencesWindow : Adw.PreferencesWindow {
               _("Lower center"), _("Upper center"), _("Center"),
             ]
           };
+        }
       }
-    }
 
       Adw.ExpanderRow figure_use_custom_style {
         title: _("Use Custom Style");
```
Sjoerd1993, line 129: It should be made clear somehow that this is to set the default style, upon launch of a new session. The style of the active window is set in Figure settings.

Perhaps "Use Custom Style by Default", or "Use Custom Style upon Launch", something like that.

Sjoerd1993 (changes requested):
Overall good changes, I've just got the single comment about that with "Custom Style" in preferences it should be made clear that this sets the defaults, not the current active canvas.

data/ui/preferences.blp
cmkohnen: Yes, absolutely

Sjoerd1993 (approved):
I like the new preference window, it's absolutely more clear and easy to find stuff.

PR #396, opened 2023-08-10
New project file format

cmkohnen:
Closes https://github.com/Sjoerd1993/Graphs/issues/394
Use new Format for storing project files.
Refactor huge portions of the codebase to adapt new format.
Rename some item and figure_settings properties.

Old project files can still be opened. (Ones created in 1.6.1 or earlier. Clipboards are lost however (like the current implementation))

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -76,11 +76,11 @@ blueprints = custom_target('blueprints',
 resources = gnome.compile_resources(
   application_id,
   application_id + '.gresource.xml',
-  dependencies: blueprints,
-  gresource_bundle: true,
-  source_dir: meson.current_build_dir(),
-  install: true,
-  install_dir: pkgdatadir,
+    dependencies: blueprints,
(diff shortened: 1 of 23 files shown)


Sjoerd1993 (approved):
Works well, apart from the safety issue of pickling, which may or may not be a big problem, another advantage is that this data type is able to be read by any text editor. Furthermore, I personally think it's slightly cleaner to work with these json-files code-wise.

When I initially went for pickle, my main reasoning for choosing it in favour of Json (which I considered as well), was the ability to store Python objects directly. Which I thought could simplify things a bit. Just pickle it, and then you can just quickly import everything upon loading without helper functions. However, it isn't able to store custom classes (which I did not know at the time), and we had to store these as dicts either way.  In the end, I feel this is the cleaner way to do this.

I found some bugs, which I fixed during testing. As it's a bit scattered over the code, I'll just put those in a seperate PR.

PR #398, opened 2023-08-11
Refactor

cmkohnen:
- refactor edit item window
- use GObject bindings (allow to change canvas properties without refresh)
- implement `application.get_settings()`
- rename plot_settings to figure_settings


Sjoerd1993 (approved):
Looks good, didn't encounter new bugs in testing either.

I know I'm a bit late here, have been trying to get unit tests to work but can't figure out to integrate them into the workflow so far. Should be possible to get them to run with the Flatpak runner when the tests are defined in meson, but haven't had too much success so far. (Although I've been trying with the regular runner so far, not the flatpak one)

PR #399, opened 2023-08-11
Add initial unit tests

Sjoerd1993:


Sjoerd1993:
This should now (finally) compile and run the tests properly. Still need to see if I also can get the appstream and desktop files test back in, which I've commented out now. Also need to fix the linting, and clean up in general. But I'm finally getting some confidence that it'll work at all.

Sjoerd1993:
This now seems to be working properly. It validates the desktop, metadata and schema file, and runs the Python unit tests that are located in the `tests` folder, all during the flatpak building stage. If any of these fail, the Flatpak CI check fails. The tests are also integrated in GNOME builder, so you can ran them from GNOME Builder as well:

image
 
At the moment these tests are just limited to check if the operations are working at all, it doesn't test the actual GUI bindings for instance. Additional tests to be added later on. 

This does add the pytest dependency, but this step may not be necessary for actual Flathub deployment itself.

cmkohnen (changes requested):
There are a lot of pycache files here, I guess these got in by mistake

Sjoerd1993:
> There are a lot of pycache files here, I guess these got in by mistake

Thanks for noticing, they're not part of my gitignore and completely missed them probably with the latest push.

python3-pytest.json
cmkohnen, line 1: I'd compact this into one simple step, instead of three submodules

data/meson.build
```diff
@@ -14,9 +14,15 @@ desktop_file = i18n.merge_file(
   install_dir: join_paths(datadir, 'applications')
 )
 
-desktop_utils = find_program('desktop-file-validate', required: false)
-if desktop_utils.found()
-  test('Validate desktop file', desktop_utils, args: [desktop_file])
+desktop_file_validate = find_program('desktop-file-validate', required: false)
+if desktop_file_validate.found()
+  test(
+    'validate-desktop',
```
cmkohnen, line 20: In my understanding this is the display name, this should be the one it was before imo

cmkohnen (changes requested):
I get warnings about GIntrospect libraries not being "initialized" first (Gtk, Gdk and Graps)

data/meson.build
Sjoerd1993: I think this should work. The previous tests for the desktop file and the schema were not working at all, they actually don't in the default GNOME Builder Python template. 

The solution I've used here, I've adapted from GNOME Sound Recorder, see [there meson.build file in data](https://gitlab.gnome.org/World/vocalis/-/blob/master/data/meson.build?ref_type=heads).

python3-pytest.json
Sjoerd1993: Yeah, I was actually hoping to get it working with a single wheel but that doesn't build for some reason. 

The manifest for this pytest was automatically generated using `flatpak-pip-generator`.  By default it fails to build with just the `pytest` install, as the dependencies `exceptiongroup` and `tomli` are not pulled in for some reason. So I added those to the requirements.txt when generating the manifest. This is the reason this is done in three steps. I can put this all in one single step. But, that would also mean that it should be manually be edited when updating these dependencies. Which is not a big issue by the way. So I'm fine with either.

Sjoerd1993:
> I get warnings about GIntrospect libraries not being "initialized" first (Gtk, Gdk and Graps)

The only warnings I'm getting locally, are about Gtk and Gdk needing to have their version specified. I was thinking to see if I can address that perhaps in a later PR. However, this is not really compatible with the linter, that really doesn't want any code before the imports of these modules. So then we should turn off those rules in flake8. So I was also fine with keeping those warnings for what they are, as the correct version of these modules are already specified in the meson build file as well.

See warning:
```
 ../../../app/lib/python3.10/site-packages/graphs/utilities.py:7
  /app/lib/python3.10/site-packages/graphs/utilities.py:7: PyGIWarning: Gdk was imported without specifying a version first. Use gi.require_version('Gdk', '4.0') before import to ensure that the right version gets loaded.
    from gi.repository import GLib, Gdk, Gio, Gtk

../../../app/lib/python3.10/site-packages/graphs/utilities.py:7
  /app/lib/python3.10/site-packages/graphs/utilities.py:7: PyGIWarning: Gtk was imported without specifying a version first. Use gi.require_version('Gtk', '4.0') before import to ensure that the right version gets loaded.
    from gi.repository import GLib, Gdk, Gio, Gtk

../../../app/lib/python3.10/site-packages/graphs/item.py:4
  /app/lib/python3.10/site-packages/graphs/item.py:4: PyGIWarning: Graphs was imported without specifying a version first. Use gi.require_version('Graphs', '1') before import to ensure that the right version gets loaded.
    from gi.repository import GObject, Graphs

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html

```

python3-pytest.json
Sjoerd1993: I've cleaned the module up now anyway, as it is nicer that way and in line with the other manifests. The reason for having this manifest separate from the main python dependencies file, is that it may not be necessary in the Flathub repository for instance.  So that makes it easier to just exclude it if we don't run these tests.

data/meson.build
cmkohnen: I meant just the display name (what you see in GNOME Builder in Unit tests). Changing the display name to something more out of the english language would be the ideal here. A quick check shows no breakage there. See l. 59

python3-pytest.json
cmkohnen: Yes. I'm absolutely fine with introducing a new module here as it wouldn't be necessary for a release build over on flathub for example. My point was just to have it as one to improve build times

cmkohnen:
Regarding the warnings, these are what I meant. Perhaps adding the necessary steps in __init__.py should solve the issue

Sjoerd1993:
> Regarding the warnings, these are what I meant. Perhaps adding the necessary steps in **init**.py should solve the issue

I'll be honest, I didn't think this would actually work as I assumed it would check on a per-file basis. But this suggestion works perfectly, we can just specify these versions in the testing init file. And since there's no modules to be imported after, there's no issues with the linter either. 

Specifying this in the main code, will give a structure like this:

```
import gi

gi.require_version('Gtk', '4.0')
gi.require_version('Adw', '1')

from gi.repository import Gtk, Gio, Adw
```

Which is exactly what the GNOME Builder template is doing as well. But as we have these `gi.require` statements in between the imports, flake8 starts to complain. Now we don't need to be dogmatic about such rules, I would be OK with ignoring it in flake8 if that would give as better safety, but as the specific versions are already set in `meson`, it feels a bit redundant. So this solution of putting them in the test init is a good one imo.

data/meson.build
Sjoerd1993: Ah, right. Yeah you are absolutely right, I agree here. The issue was likely simply that I was looking elsewhere how to solve the issue I was facing, and went a bit too far when adopting their solution.

cmkohnen:
Two things now:
1 - see pr I made
2 - Is it possible to stop a pycache generating in the source directory?

Sjoerd1993:
> Is it possible to stop a pycache generating in the source directory?

Pycache is generated when importing modules, for me this only seems to generate within the testing source directory. You can apparantly turn it off with `sys.dont_write_bytecode = True`, but it needs the cache to actually access the imported modules. Turning it off gives me:

```
Projects/Graphs/tests/__init__.py:9: in <module>
    gi.require_version("Graphs", "1")
/usr/lib/python3.10/site-packages/gi/__init__.py:126: in require_version
    raise ValueError('Namespace %s not available' % namespace)
E   ValueError: Namespace Graphs not available
=========================== short test summary info ============================
ERROR Projects/Graphs/tests/test_operations.py - ValueError: Namespace Graphs not available
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 0.30s ===
```

A possibility would maybe be to simply remove the pycache folder "forcefully" after the last unit test, but that's not really conventional with any unit testing nor really helpful probably. In my case, I just added */__pycache__ to `.gitignore`. The pytest's own cache (which is in the project root as hidden directory) should be added to gitignore automatically.

cmkohnen:
Hmm, maybe copying the test suite to the build directory might help, this way the pycache gets generated but inside directory that doesn't affect git regardless

cmkohnen:
> Hmm, maybe copying the test suite to the build directory might help, this way the pycache gets generated but inside directory that doesn't affect git regardless

But honestly, thats more of a build system configuration instead of actual unit test, so perhaps for another pr

cmkohnen (approved):

Sjoerd1993:
> > Hmm, maybe copying the test suite to the build directory might help, this way the pycache gets generated but inside directory that doesn't affect git regardless
> 
> But honestly, thats more of a build system configuration instead of actual unit test, so perhaps for another pr

That's actually really doable and a relatively minor change, just copy the folder to a subdirectory in the build directory, and run pytest on that. See latest commit.

PR #400, opened 2023-08-13
Refactor

cmkohnen:
- Use Data class for managing data
- Remove graphs.py, calculation.py and move logic to more suitable places
- Make things like reloading item menu happen automatically
- plot items automatically upon change
- Optimize plotting items (reduce number of loops and lookups)
- Use integers for some item attributes (parallel to settings, this allows for some faster lookups and other tricks during plotting)
- Implement `ArtistWrapper` to save unnecessary "refreshs"
- Make Selection persistent
- Remove RenameWindow


cmkohnen:
Oh, apparently I didn't mark this as draft, anyway this is ready now (also number 400)

src/canvas.py
```diff
@@ -44,29 +44,28 @@ def __init__(self, application):
...
+        self._legend = True
+        self._legend_position = LEGEND_POSITIONS[0]
+        self._handles = []
 
         for prop in dir(self.props.application.props.figure_settings.props):
-            if prop not in property_ignorelist:
+            if prop not in ["use_custom_style", "custom_style"]:
                 self.props.application.props.figure_settings.bind_property(
-                    prop, self, prop, 2,
+                    prop, self, prop, 1 | 2,
```
Sjoerd1993, line 64: I know this was already in the old code, but I tried to look a bit into the bind_property function from GObject, where this seems to come from.  Binding seems to be

`  g_object_bind_property (action, "active", widget, "sensitive", 0);  `

But the docs didn't make it very obvious for me what each argument does. Mainly what does the 1 | 2 pipe do for instance? 

Not relevant for the review, just wanted to know for my own understanding.

src/canvas.py
```diff
@@ -293,7 +313,7 @@ def right_scale(self, scale: int):
 
     @GObject.Property(type=float)
     def min_bottom(self):
-        return min(self.axis.get_xlim())
+        return self.axis.get_xlim()[0]
```
Sjoerd1993, line 316: In principle the first element should always be equal to the minimum value, so both sides are equivalent. In my view using `min` or `max` of the limits is slightly more readable though. Any specific reason for changing these?

src/canvas.py
```diff
@@ -419,32 +439,30 @@ def remove_rubberband(self):
...
-            "min_left": self.canvas.props.min_left,
-            "max_left": self.canvas.props.max_left,
-            "min_top": self.canvas.props.min_top,
-            "max_top": self.canvas.props.max_top,
-            "min_right": self.canvas.props.min_right,
-            "max_right": self.canvas.props.max_right,
-        })
+        for direction in ["bottom", "left", "top", "right"]:
+            self.canvas.notify(f"min-{direction}")
+            self.canvas.notify(f"max-{direction}")
```
Sjoerd1993, line 444: This helps reduce the clutter a bit :)

src/graphs.py
```diff
@@ -1,128 +0,0 @@
...
-        plotting_tools.hide_unused_axes(self, self.canvas)
-    self.canvas.set_ticks()
-    hide_unselected = \
-        self.settings.get_child("general").get_boolean("hide-unselected")
-    for item in reversed(self.datadict.values()):
-        if item is None or (hide_unselected and not item.selected):
-            continue
-        self.canvas.plot(item)
-    self.canvas.set_legend()
-    self.canvas.draw()
```
Sjoerd1993, line 128: It's probably good to see this file go, as with many cases it made sortof sense in very early development to have a central general purpose function, but as the project grows it becomes a weird mismatch of "other/central functions".

src/operations.py
```diff
@@ -157,16 +157,16 @@ def center(_item, xdata, ydata, center_maximum):
     return new_xdata, ydata, True, False
 
 
-def shift_vertically(item, xdata, ydata, left_scale, right_scale, datadict):
+def shift_vertically(item, xdata, ydata, left_scale, right_scale, items):
```
Sjoerd1993, line 160: In the main class, we now changed the attribute previously known as `datadict` to `data`, in this function it's called `items`. 

I'm happy with a rename, as `datadict` is maybe not ideal (the idea was that it was a dictionary that contains all the data), but perhaps we should stick to a more consistent naming throughout.

Sjoerd1993 (commented):
Apologies for the late feedback, see comments I made which are mostly just more minor questions. Overall the direction of this PR is one I like.

src/canvas.py
cmkohnen: My main reason was performance gain, as we don't have to compare the two values to see which is the smaller. 
According to [matplotlib documentation](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.set_xlim.html), they represent left and right (or bottom and top for y), which matches our usage as upper/lower

src/operations.py
cmkohnen: items in this case is a list of all items. Before this function would just query all items from the datadict for each item this operation was performed on. This refactor moves this query out to the caller function (window.py/shift_vertically)

cmkohnen:
> This also breaks item renaming from the item menu. At least when you then switch to a different item, see attached screencast.

Fixed, we now use a different logic (not repopulating the chooser in case names don't match, but always update the name, when a new item is selected in place using `GtkStringList.splice()`)

Sjoerd1993 (approved):

PR #401, opened 2023-08-15
Refresh less aggressively

cmkohnen:
Continuation of https://github.com/Sjoerd1993/Graphs/pull/400

Implement ArtistWrapper, a GObject based wrapper that converts between
GObject properties and matplotlib artist's attributes.
Automatically update, whenever an item property changes, reducing
overhead, as only that specific property and the legend gets update and
a draw triggered instead of recreating every item's artist every time.


Sjoerd1993:
Just a quick note, apologies for the slow review here. I'm having a week off, and am not spending as much time in front of my computer screen so it's a bit slow here. Will try to go through both PR's tomorrow if I have the time. 

Until then, I can at least note that the Graph now is not refreshed when the axes is changed. So if an item is set to the top axis for instance, it doesn't actually set it until a refresh is triggered. (Which is slightly more difficult now to trigger by hand, which is nice. As it really should only refresh if not otherwise possible, so overall this is looking good)

cmkohnen:
> Just a quick note, apologies for the slow review here. I'm having a week off, and am not spending as much time in front of my computer screen so it's a bit slow here. Will try to go through both PR's tomorrow if I have the time.

No worries

cmkohnen:
See #400

PR #403, opened 2023-08-24
small fixes

cmkohnen:
Small fixes across the codebase from my testing and pylint analysis

### DIFF
--- a/src/file_io.py
+++ b/src/file_io.py
@@ -18,18 +18,18 @@ from matplotlib.style.core import STYLE_BLACKLIST
 import numpy
 
 
-def save_item(file, item):
+def save_item(file, item_):
     delimiter = "\t"
     fmt = delimiter.join(["%.12e"] * 2)
     stream = _get_write_stream(file)
-    if item.xlabel != "" and item.ylabel != "":
-        _write_string(stream, item.xlabel + delimiter + item.ylabel + "\n")
-    for row in numpy.stack([item.xdata, item.ydata], axis=1):
+    if item_.xlabel != "" and item_.ylabel != "":
+        _write_string(stream, item_.xlabel + delimiter + item_.ylabel + "\n")
+    for row in numpy.stack([item_.xdata, item_.ydata], axis=1):
         _write_string(stream, fmt % tuple(row) + "\n")
     stream.close()
 
 
-def import_from_project(self, import_settings):
+def import_from_project(_self, import_settings):
     return [item.new_from_dict(dictionary)
             for dictionary in parse_json(import_settings.file)["data"]]
 
@@ -103,7 +103,7 @@ def import_from_xry(self, import_settings):
 
 
 def import_from_columns(self, import_settings):
-    item = Item.new(self, name=import_settings.name)
+    item_ = Item.new(self, name=import_settings.name)
     columns_params = self.settings.get_child(
         "import-params").get_child("columns")
     column_x = columns_params.get_int("column-x")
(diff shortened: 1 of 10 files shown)


src/file_io.py
```diff
@@ -18,18 +18,18 @@
 import numpy
 
 
-def save_item(file, item):
+def save_item(file, item_):
```
Sjoerd1993, line 21: Just as a general point, the trailing underscore is to distinguish it from item methods in e.g. dictionaries?

src/operations.py
```diff
@@ -189,9 +189,9 @@ def shift_vertically(item, xdata, ydata, left_scale, right_scale, items):
         ymin = min(x for x in previous_ydata if x != 0)
         ymax = max(x for x in previous_ydata if x != 0)
         if item.yposition == "left":
-            linear = (left_scale == "linear")
+            linear = left_scale == "linear"
```
Sjoerd1993, line 192: Any specific reason for getting rid of the brackets? Personally I feel the brackets make it a bit more readable, as it distinguishes the comparison  (==) from the assignment (=) a bit visually.

src/utilities.py
```diff
@@ -214,3 +206,4 @@ def get_next_color(items):
     for color in color_cycle:
         if color not in used_colors:
             return color
+    return "000000"
```
Sjoerd1993, line 209: In principle it should never reach this last return value. If all colors in the cycle are full, it should have emptied the color_cycle again.

But I don't mind this return value as a potential safeguard anyway in unforeseen edge cases.

src/operations.py
```diff
@@ -189,9 +189,9 @@ def shift_vertically(item, xdata, ydata, left_scale, right_scale, items):
...
         if item.yposition == "right":
```
Sjoerd1993, line 193: These are saved as enums now, with integer values:

```
        if item.yposition == 0:
            linear = left_scale == 0
        if item.yposition == 1:
            linear = right_scale == 0
```

src/operations.py
```diff
@@ -189,9 +189,9 @@ def shift_vertically(item, xdata, ydata, left_scale, right_scale, items):
         ymin = min(x for x in previous_ydata if x != 0)
```
Sjoerd1993, line 189: If item_ with the trailing underscore is used in the other methods, maybe it makes sense to do this here as well. Unless there's a reason for changing to that elsewhere that I missed.

Sjoerd1993 (changes requested):

Sjoerd1993:
There's also something out of wack with the plots when doing operations on multiple data sets. Basically it only plots the latest data set. Seems to be in the latest main commit as well, so probably a regression from last refactor. 
Will probably take a look into that tomorrow.

[Uploading Screencast from 2023-08-24 20-59-23.webm…]()

src/file_io.py
cmkohnen: A trailing underscore is a python convention to distinguish between definitions in a higher scope.

Here, item is already defined, hence appending it.

cmkohnen:
> There's also something out of wack with the plots when doing operations on multiple data sets. Basically it only plots the latest data set. Seems to be in the latest main commit as well, so probably a regression from last refactor. Will probably take a look into that tomorrow.
> 
> Uploading Screencast from 2023-08-24 20-59-23.webm…

Fixed the issue, only the last item emitted a notify signal for x and y data

Sjoerd1993 (approved):

PR #404, opened 2023-08-25
Fix shift vertically

sstendahl:
Very minor fix.
Fix the indentation of shift vertically, so that it shifts the correct data. Current main branch also leaves an error.

### DIFF
--- a/src/operations.py
+++ b/src/operations.py
@@ -199,6 +199,7 @@ def shift_vertically(item, xdata, ydata, left_scale, right_scale, items):
             else:
                 new_ydata = [value * 10 ** shift_value_log for value in ydata]
             return xdata, new_ydata, False, False
+    return xdata, ydata, False, False
 
 
 def cut_selected(_item, _xdata, _ydata):


src/operations.py
```diff
@@ -198,7 +198,7 @@ def shift_vertically(item, xdata, ydata, left_scale, right_scale, items):
                 new_ydata = [value + shift_value_linear for value in ydata]
             else:
                 new_ydata = [value * 10 ** shift_value_log for value in ydata]
-        return xdata, new_ydata, False, False
+            return xdata, new_ydata, False, False
```
cmkohnen, line 201:
```suggestion
            return xdata, new_ydata, False, False
        return xdata, ydata, False, False
```
Oh this was my bad for not testing correctly. It was meant as a failsafe, if for some absurd reason the logic didn't work.
Something like this should then work however

Sjoerd1993: Needs one indentation less, which I added in the latest commit.

PR #406, opened 2023-08-25
Fix clipboard

Sjoerd1993:
This took way too much time for such a simple fix, but in the current branch the clipboard is broken. The reason turned out to be that the data in the clipboard is saved as a list, where the items are all pointing to the same item in memory. So any changes to the data affect the entire clipboard.


cmkohnen (approved):
Huh, I never noticed any issues during my testing.

PR #407, opened 2023-08-25
only reload when edited style is in use

cmkohnen:
Also fix issues regarding style change


Sjoerd1993 (approved):
Simple change that makes sense to do.

PR #409, opened 2023-08-25
ui tweaks

cmkohnen:
Make Dialogs consistent.
Also fix adding a style

### DIFF
--- a/src/export_figure.py
+++ b/src/export_figure.py
@@ -20,7 +20,8 @@ class ExportFigureWindow(Adw.Window):
     def __init__(self, application):
         super().__init__(
             application=application, transient_for=application.main_window,
-            file_formats=application.canvas.get_supported_filetypes_grouped(),
+            file_formats=application.main_window.toast_overlay.get_child(
+            ).get_supported_filetypes_grouped(),
         )
 
         self.file_format.set_model(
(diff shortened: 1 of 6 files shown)


src/export_figure.py
```diff
@@ -20,7 +20,8 @@ class ExportFigureWindow(Adw.Window):
     def __init__(self, application):
         super().__init__(
             application=application, transient_for=application.main_window,
-            file_formats=application.canvas.get_supported_filetypes_grouped(),
+            file_formats=application.main_window.toast_overlay.get_child(
+            ).get_supported_filetypes_grouped(),
```
Sjoerd1993, line 24: Why this change? `get_supported_filetypes_grouped()` is a method from Matplotlibs canvas. It doesn't seem obvious why one would call it from the child of the toast overlay in the main window. (Which happens to be the same canvas)

Sjoerd1993 (commented):
Just wondering about the comment I placed. Indeed does fix the broken add style.

src/export_figure.py
cmkohnen: We no longer store a reference to the actual canvas, as this is probably (one of) the only cases, where we would need to access it.
This is why it broke before (as I didn't catch it during search).

Sjoerd1993: Ah gotcha, I didn't catch it either as I didn't think about the actual reference no longer being stored. That makes sense then.

Sjoerd1993 (approved):

PR #410, opened 2023-08-25
Suggestion: Redesign operations

cmkohnen:
Bildschirmfoto vom 2023-08-25 17-38-21

Open for discussion if we wanna keep it the way it is or change it.
The benefits of this are in my opinion, that buttons always stay relative to each other.
On small resolutions you'd always have to scroll but given the default resolution all elements (except translation) are visible.
On a 100% scaled 1080p monitor all items are visible without the need for a scrollbar.

### DIFF
--- a/data/ui/window.blp
+++ b/data/ui/window.blp
@@ -204,23 +204,21 @@ template $GraphsWindow : Adw.ApplicationWindow {
         ScrolledWindow {
           hscrollbar-policy: never;
           Viewport {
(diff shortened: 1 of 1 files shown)


Sjoerd1993:
I think I'll approve this as it adresses one of the points of the GNOME Circle Feedback quite directly, and further tweaks can be done in seperate PR's.

Sjoerd1993 (approved):

PR #411, opened 2023-08-25
language maintenance

cmkohnen:
Update POTFILES and update .po-files


Sjoerd1993 (approved):
I have set up Graphs at weblate, which will make translation contributions _a lot_ easier for contributors (especially for ones that are less comfortable with coding and Github). We should be elligible for Libre hosting, which I have just requested. One of the requirements is that we will put in the Readme that we use Weblate for translation, but that's OK by me, and probably convenient anyway as that makes it discoverable for people wanting to contribute with localization.

https://hosted.weblate.org/projects/graphs/

cmkohnen:
I do believe it is one of the requirements to have hosted translation capabilities for GNOME Circle as well.
I haven't really worked with weblate, but know that a lot of big projects use it, so having translations more accessible is really great.

PR #413, opened 2023-08-26
experimental new clipboard

cmkohnen:
Still WIP, swapping items doesn't work yet, all other changes work reasonably well

### DIFF
--- a/src/clipboard.py
+++ b/src/clipboard.py
@@ -2,7 +2,7 @@ import copy
 
 from gi.repository import Adw, GObject
 
-from graphs import ui
+from graphs import ui, item
 
 import numpy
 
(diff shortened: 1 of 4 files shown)


Sjoerd1993:
Works quite well, but sometimes the view seems to get out of sync somehow on the redo action, it's fine on undo:

[Screencast from 2023-08-27 13-09-14.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/a0c5c074-518d-4df8-9e72-7be53becc744)

Sjoerd1993:
If it helps, I noticed it's not the view that is out of sync. The limits are set correctly to what Redo should do. Instead, it's the data itself that is out of sync with one single step when pressing redo. Haven't been able to look into that code-wise in too much detail unfortunately.

cmkohnen:
The issue isn't that limits are out of place, sometimes item's simply vanish

Sjoerd1993:
> The issue isn't that limits are out of place, sometimes item's simply vanish

From what I've checked, the limits are set correctly on redo, but the item values are out-of-sync if that makes sense. So when it performs a redo action, the limits of the "next" item are usually obtained, but it does use the item-values of one state/position after.

I haven't had the time to properly investigate, but everything is correct in the undo function so I think the bug lies somewhere in the redo logic.

cmkohnen:
The culprit was references. When obtaining information on which property changed, a reference was kept. Using deepcopy solved the issue

PR #415, opened 2023-08-28
migrate clipboard

cmkohnen:
Add automatic clipboard migration.
Since we didn't store associated limits before, these are calculated anew. However since we can't know the scale, padding is omitted.

### DIFF
--- a/src/migrate.py
+++ b/src/migrate.py
@@ -3,7 +3,7 @@ import contextlib
 import pickle
 import sys
 
-from graphs import file_io, utilities
+from graphs import file_io, misc, utilities
 
 
 CONFIG_MIGRATION_TABLE = {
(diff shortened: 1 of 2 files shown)


Sjoerd1993 (approved):
Seems to work in my quick tests.

Sjoerd1993:
I actually hadn't written a migration function for the first rewrite we did here. So before today, the clipboard was just lost when opening an old project. I did mean to write a migration function before we release by the way, just hadn't gotten around to it yet. Probably good that I didn't do that directly, as it'd probably be redundant after the second clipboard redesign with the transactional clipboard.

> Since we didn't store associated limits before, these are calculated anew. However since we can't know the scale, padding is omitted.

That's perfectly acceptable, and relatively close to what the old implementation (the one currently on Flathub stable) does. Which is just recalculate the limits on the fly (with padding).

PR #417, opened 2023-08-28
fix limits in figure settings

cmkohnen:


Sjoerd1993 (approved):
My guess is that this broke when changes became instantanious, this seems to work.

PR #418, opened 2023-08-28
Limit item name length

Sjoerd1993:
Somewhere along the way the item labels lost their limits. This makes the side-panel expand to unreasonable widths with very long filenames. This fixes the label length again

### DIFF
--- a/src/item_box.py
+++ b/src/item_box.py
@@ -6,6 +6,7 @@ from gi.repository import Adw, GLib, GObject, Gdk, Gtk
 
 from graphs.edit_item import EditItemWindow
 from graphs.item import ItemBase
+from graphs import utilities
 
 
 @Gtk.Template(resource_path="/se/sjoerd/Graphs/ui/item_box.ui")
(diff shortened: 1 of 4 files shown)


cmkohnen (changes requested):
This only fixes name length for Items (not TextItems).

A proper place for this should be to implement this property in item_box and use `utilities.shorten_label`.
This would also allow for longer names to persist in theory. E.g. when importing from some files and exporting again

src/item_box.py
```diff
@@ -98,5 +99,9 @@ def delete(self, _button):
         )
 
     @Gtk.Template.Callback()
-    def edit(self, _):
+    def edit(self, _button):
         EditItemWindow(self.props.application, self.props.item)
+
+    @Gtk.Template.Callback()
+    def on_label_change(self, _itembox, _gobject):
+        self.label.set_text(utilities.shorten_label(self.item.name))
```
cmkohnen: I'm not sure this logic is the way to go.

First, when an item's name is updated, the label instantly gets updated to the same value (potentially stretching the sidebar in the process).
Then the signal listener queries the items property and sets a shortened version, reducing the width again.

Using an intermediary function (e. g. like in #420), this issue can be avoided, by shortening the label whilst it gets copied from the item.

src/item_box.py
```diff
@@ -4,10 +4,10 @@
 
 from gi.repository import Adw, GLib, GObject, Gdk, Gtk
 
+from graphs import utilities
 from graphs.edit_item import EditItemWindow
 from graphs.item import ItemBase
```
cmkohnen, line 10:
```suggestion


```
Not sure, why flake doesn't flag this but two newlines are required

cmkohnen (approved):

PR #424, opened 2023-08-29
use getters and setters

cmkohnen:
Use getters and setters for most commonly used properties.
This is more in line with GObject based libraries, which have explicit getter and setter methods.
This also allows to make the code a bit more readable and in case we need them inject safeguards into those methods.

In cases, where have function-defined properties, this can in some cases improve performance, as annotators have a slight overhead compared to regular functions.

### DIFF
--- a/src/actions.py
+++ b/src/actions.py
@@ -11,12 +11,12 @@ from graphs.styles import StylesWindow
 
 
 def toggle_sidebar(_action, _shortcut, self):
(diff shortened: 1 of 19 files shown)


Sjoerd1993:
Not sure how much performance differs in real life, but definitely more readable either way.

Sjoerd1993 (approved):

PR #428, opened 2023-08-29
fix help overlay

### LINKED ISSUE #427: Keybord Shortcuts dialog does not show some shortcuts
Hi, Keyboard Shorcuts dialog (aka Help Overlay) seem broken.

Ekran Görüntüsü - 2023-08-29 14-01-32

It try to read accellerators from [action names](https://github.com/Sjoerd1993/Graphs/blob/main/data/ui/help_overlay.blp#L20) but there is no defination for shorcuts in [main.py](https://github.com/Sjoerd1993/Graphs/blob/main/src/main.py#L22-L28)

There are several triggers in the [Window.blp](https://github.com/Sjoerd1993/Graphs/blob/main/data/ui/window.blp#L7-L79)

cmkohnen:
fix #427

### DIFF
--- a/data/ui/help_overlay.blp
+++ b/data/ui/help_overlay.blp
@@ -17,22 +17,22 @@ ShortcutsWindow help_overlay {
 
       ShortcutsShortcut {
         title: C_("shortcut window", "Open Preferences");
-        action-name: "app.preferences";
+        accelerator: "<primary>p";
       }
 
       ShortcutsShortcut {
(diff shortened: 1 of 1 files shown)


Sjoerd1993:
It doesn't automatically update anymore (previous implementation retrieved the shortkeys automatically), but that's not a big issue by itself. Just a thing to consider if we were to change shortcuts.

PR #429, opened 2023-08-29
Fix edit item window

Sjoerd1993:
Function was not called properly, making it impossible to load add item window


cmkohnen (approved):
Probably one of the several instances Builder crashed on me

PR #431, opened 2023-08-29
Radians scale

cmkohnen:
Add Radians Scale as a third option
Bildschirmfoto vom 2023-08-29 18-41-14

Also implement bas structure for future custom scales

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -10,7 +10,7 @@ from gettext import gettext as _
 
 from gi.repository import Adw, GLib, GObject, Gio
 
-from graphs import actions, migrate, scales, ui
+from graphs import actions, migrate, ui
 from graphs.clipboard import DataClipboard, ViewClipboard
 from graphs.data import Data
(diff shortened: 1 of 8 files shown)


src/main.py
```diff
@@ -10,7 +10,7 @@
 
 from gi.repository import Adw, GLib, GObject, Gio
 
-from graphs import actions, migrate, ui
+from graphs import actions, migrate, scales, ui
```
cmkohnen, line 13:
```suggestion
from graphs import actions, migrate, ui
```

Sjoerd1993:
This is great, I've got a PR ready based on this that adds square root scaling as well. On the longer term, I'll probably add a context-dependend one for q-spacing. Which should then pop up when importing .xrdml files at least. So it's possible then to switch between reciprocal space and angular space. But that's all lower down on the priority list.

Sjoerd1993 (approved):

PR #432, opened 2023-08-29
Stop canvas from stealing focus

Sjoerd1993:
Currently, when panning/zooming on the graph, the canvas itself steals focus. This means that shortcuts such as ctrl+z or ctrl+N and such no longer work.

There's no real reason to have a the focus on the canvas. This change makes it such that shortcuts keep working.

### DIFF
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -92,6 +92,7 @@ class Canvas(FigureCanvas):
         """
         GObject.Object.__init__(self, application=application)
         super().__init__()
+        self.set_can_focus(False)
         self.figure.set_tight_layout(True)
         self.mpl_connect("pick_event", self._on_pick)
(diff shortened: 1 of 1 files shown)


src/canvas.py
```diff
@@ -92,6 +92,7 @@ def __init__(self, application):
         """
         GObject.Object.__init__(self, application=application)
```
cmkohnen, line 93:
```suggestion
        GObject.Object.__init__(self, application=application, can_focus=False)
```

src/canvas.py
```diff
@@ -92,6 +92,7 @@ def __init__(self, application):
...
         super().__init__()
+        self.set_can_focus(False)
```
cmkohnen, line 95:
```suggestion
```

PR #433, opened 2023-08-29
Add square root scaling

Sjoerd1993:
Adds square root scaling as fourth option. I don't have any immediate plans of adding more options here (inverted (1/x) would be an option though). But as I mentioned, I am considering adding a context-depended option for .xrdml files so it will be possible to easily toggle between angular coordinates and reciprocal space. (Something which isn't even possible with the official DataViewer software from Panalytical, which frankly is pretty bad anyway)

### DIFF
--- a/src/scales.py
+++ b/src/scales.py
@@ -43,9 +43,6 @@ class SquareRootScale(scale.ScaleBase):
     """Class for generating custom square root scale."""
     name = "squareroot"
 
-    def __init__(self, axis):
-        scale.ScaleBase.__init__(self, axis)
-
     def set_default_locators_and_formatters(self, axis):
         axis.set_major_locator(ticker.AutoLocator())
         axis.set_major_formatter(ticker.ScalarFormatter())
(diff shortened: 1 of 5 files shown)


src/scales.py
```diff
@@ -39,4 +39,49 @@ def set_default_locators_and_formatters(self, axis):
         axis.set_major_locator(ticker.MultipleLocator(base=numpy.pi))
 
 
+class SquareRootScale(scale.ScaleBase):
+    """Class for generating custom square root scale."""
+    name = "squareroot"
+
+    def __init__(self, axis):
+        scale.ScaleBase.__init__(self, axis)
+
```
cmkohnen, line 48: Is this needed? Simply removing this should result in the same init. Unless I'm missing something here.

cmkohnen (commented):
This would have been my next step as well.

This implementation looks very clean. Much better than my initial test.

See comment for one question

src/scales.py
Sjoerd1993: I've added this as this is what all examples I found online were doing, see [this for example](https://stackoverflow.com/questions/64187145/trying-to-make-a-cubed-scalebase-class-in-matplotlib-but-i-cant-get-it-working), but honestly critically looking at it I don't see either that it actually does anything other than the standard init. It's completely repetitive, removing it doesn't break anything either.

Also looking at [official documentation](https://matplotlib.org/stable/gallery/scales/custom_scale.html) gives me the impression this init is just redundant. So I am removing this.

Sjoerd1993:
Mainly debating whether an "Inverse" scale would be worth it to add. I think it can be useful to add, but at the same time we have to strike a balance between useful scales being available and having too many options out there.

cmkohnen (approved):

PR #435, opened 2023-08-30
Reduce preferences

cmkohnen:
Remove default Figure Settings Values from Preferences.
Add Set as Default Button to Figure Settings

### DIFF
--- a/data/se.sjoerd.Graphs.gschema.xml
+++ b/data/se.sjoerd.Graphs.gschema.xml
@@ -45,12 +45,12 @@
     <value nick="Square Root" value="3"/>
   </enum>
 
-  <enum id="se.sjoerd.Graphs.figure.x-positions">
+  <enum id="se.sjoerd.Graphs.x-positions">
     <value nick="bottom" value="0"/>
     <value nick="top" value="1"/>
   </enum>
 
-  <enum id="se.sjoerd.Graphs.figure.y-positions">
+  <enum id="se.sjoerd.Graphs.y-positions">
     <value nick="left" value="0"/>
     <value nick="right" value="1"/>
   </enum>
@@ -80,6 +80,12 @@
     </key>
     <key name="override-item-properties" type="b">
       <default>true</default>
+    </key>
+    <key name="x-position" enum="se.sjoerd.Graphs.x-positions">
+      <default>"bottom"</default>
+    </key>
+    <key name="y-position" enum="se.sjoerd.Graphs.y-positions">
+      <default>"left"</default>
     </key>
   </schema>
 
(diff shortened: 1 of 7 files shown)


Sjoerd1993:
I think this is in general a good change, and also gets rid of the slightly confusing state of having two settings for the Figure properties. (One which sets the defaults, and one that sets the current state).

Only thing I'd like to have added to this is a toast pop up maybe, so that there's some feedback when pressing the "Set as Default" button.

cmkohnen:
> Only thing I'd like to have added to this is a toast pop up maybe, so that there's some feedback when pressing the "Set as Default" button.
> 
> It does overlap somewhat with the button though, but I think it's still good to have some feedback, so that it's clear that the button did something. [Screencast from 2023-08-30 16-27-32.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/98af4950-b437-4b7f-805b-965ddf9dac1f)

I don't think the overlapping is too big of an issue, as you wouldn't want to set the defaults twice without having the time to change a value.

Since you already have an implementation for this, feel free to push to this branch directly

Sjoerd1993 (approved):

Sjoerd1993:
Altenative presentations of the button would be as a pill, or in the top right corner in the header bar:

image


Putting it in the top right corner probably doesn't work to well, as it needs to have a relatively long label to work ("Set as default" is quite perfect, but that takes too much space for having it in the header bar). Having it as a pill also works, see screenshot above. This is what the GNOME HIG say about this:

> The [pill](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/style-classes.html#pill) and [circular](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/style-classes.html#circular) styles change the shape of a button. Pill can be used when a text button is situated in open space, and is a primary action for the view. Here the larger rounder shape helps to accommodate the button to its surroundings, as well as making it a more inviting click target.

But this button may not count as being the primary action for the view (not sure). So I am not 100% convinced that a pill is recommended here. Just wanted to throw it out as an alternative presentation.

cmkohnen:
I've tried adding it to the headerbar but since its an `Adw.PreferencesWindow` that isn't really easily accomplished here. I do find the pill style a bit more appealing than the stretched one we have now. Either way I do not have a strong personal preference here.

Sjoerd1993:
> Either way I do not have a strong personal preference here

Ah right, almost forgot it was an `Adw.PreferencesWindow`, I actually ran into the same issue once. Can't really remember where I tried to add a button. 

Either way, I don't have an extremely strong opinion either. I just suddenly remember @bertob suggested a pill button in our dialogs. Which we've since updated to Action Dialogs. I think my slight preference goes to pill as well, but again I think both work. One thing with the pill, the toast is completely above the button. On the other hand, as you mention you never want/need to press it twice in a row anyway.

PR #442, opened 2023-08-31
Add inverse scaling

Sjoerd1993:
Add inverse scaling to the amount of options. Needed some tweaks to get this to work correctly with panning/zooming, `select_data`, as well as with axis-dependent operations like `shift_vertically` and `optimize_limits`.

Finally, this also adds an `optimize_limits` call to scaling changes, as the scaling is often not good at all directly after changing scale, so running `optimize_limits` is often the first thing you want to do anyway.

### DIFF
--- a/data/ui/preferences.blp
+++ b/data/ui/preferences.blp
@@ -142,4 +142,4 @@ template $PreferencesWindow : Adw.PreferencesWindow {
       }
     }
   }
-})
+}
--- a/data/ui/preferences.blp
+++ b/data/ui/preferences.blp
@@ -64,28 +64,28 @@ template $PreferencesWindow : Adw.PreferencesWindow {
       Adw.ComboRow figure_bottom_scale {
         title: _("Default Bottom Axis Scale");
         model: StringList{
-          strings [_("Linear"), _("Logarithmic"), _("Radians"), _("Square Root")]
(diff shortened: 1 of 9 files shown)


Sjoerd1993:
> Finally, this also adds an optimize_limits call to scaling changes, as the scaling is often not good at all directly after changing scale, so running optimize_limits is often the first thing you want to do anyway.

Given PR #440, I may want to remove that change anyway.

data/ui/preferences.blp
cmkohnen, line 1: I think your're somehow using an old version of this file here

Sjoerd1993: Thanks for noticing, this is correct. I know that I started this as a local branch (before the scaling got removed from preferences), somehow this got out of sync.

cmkohnen:
Correct me if I'm wrong, but now the new scale isn't represented in preferences anymore

Sjoerd1993:
> Correct me if I'm wrong, but now the new scale isn't represented in preferences anymore

Yes, but all scales have been moved away from preferences since PR #435, so that should be the case.

cmkohnen:
Oh sorry, my bad

cmkohnen (approved):

PR #444, opened 2023-08-31
Prevent view menu from stealing focus

Sjoerd1993:
See title. When scaling is changed through the view menu, the focus is changed to the view menu. Making it impossible to use shortcuts. This fixes this.

### DIFF
--- a/data/ui/window.blp
+++ b/data/ui/window.blp
@@ -133,6 +133,7 @@ template $GraphsWindow : Adw.ApplicationWindow {
           tooltip-text: _("View Menu");
           always-show-arrow: true;
           menu-model: view_menu;
+          can-focus: false;
         }
         Separator {styles ["spacer"]}
       }
--- a/data/ui/window.blp
+++ b/data/ui/window.blp
@@ -618,6 +618,7 @@ template $GraphsWindow : Adw.ApplicationWindow {
 
       [content]
       Adw.ToastOverlay toast_overlay {
+        focusable: true;
         height-request: 200;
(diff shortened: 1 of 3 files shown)


Sjoerd1993:
Just realized that this does make it impossible to reach the view menu using keyboard navigation, so maybe we need a slightly more clever solution. Perhaps try and see if we can drop the focus after a scale change (it seems to be the only the radio buttons that steal focus). 

Keyboard navigation is a bit clunky anyway and not really something I'd recommend, but it's one of the GNOME Circle Accesibility requirements. So probably should rethink this: 

>  [Can navigate to all elements via keyboard and use them](https://gitlab.gnome.org/Teams/Releng/AppOrganization/-/blob/main/AppCriteria.md)

Sjoerd1993:
This works, it's somewhat of a workaround, but this forces to set the focus to the toast overlay after changing the axis limits. Which solves the issue adressed in this PR, without losing keyboard navigation.

cmkohnen (approved):
This deviates from the formatting we have going a little bit (which makes it a bit harder to understand which parentheses begins and ends where). But I'm fine with that.

Sjoerd1993:
> This deviates from the formatting we have going a little bit (which makes it a bit harder to understand which parentheses begins and ends where). But I'm fine with that.

Yeah, feel free to change the formatting a bit in a future PR. It was a bit tricky to format it such that the linter approved, especially with the relatively long calls. Was tempted to just define a separate function here instead of using the lambda, but this worked.

PR #445, opened 2023-08-31
Change locator behaviour on custom axes scales

Sjoerd1993:
Changes the behaviour of locators on custom axes scaling, also offers some slight tweaks on the Inverse scaled limits.

The main point is that the locators (ticks) are now distributed evenly for the square root and inverted axis, leading to less awkward scaling in certain cases. For example with tons of ticks extremely close to each other. 

I am not 100% sure if an even distribution is always completely what we want. But in many cases it leads to a more readable graph. Here's the screenshot with and without the even ticks. One option is to have it as a toggle in the style sheet. 
Screenshot from 2023-08-31 20-22-10
 
Screenshot from 2023-08-31 20-23-47

### DIFF
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -16,7 +16,7 @@ from gi.repository import Adw, GObject, Gtk
 from graphs import artist, misc, scales, utilities
 from graphs.figure_settings import FigureSettingsWindow
 
-from matplotlib import backend_tools as tools, ticker, pyplot
+from matplotlib import backend_tools as tools, pyplot, ticker
 from matplotlib.backend_bases import NavigationToolbar2
 from matplotlib.backends.backend_gtk4cairo import FigureCanvas
 from matplotlib.widgets import SpanSelector
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -16,10 +16,12 @@ from gi.repository import Adw, GObject, Gtk
 from graphs import artist, misc, scales, utilities
 from graphs.figure_settings import FigureSettingsWindow
 
-from matplotlib import backend_tools as tools, pyplot
+from matplotlib import backend_tools as tools, ticker, pyplot
 from matplotlib.backend_bases import NavigationToolbar2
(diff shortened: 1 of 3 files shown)


Sjoerd1993:
> I am not 100% sure if an even distribution is always completely what we want. But in many cases it leads to a more readable graph. Here's the screenshot with and without the even ticks. One option is to have it as a toggle in the style sheet.

It doesn't seem possible to add custom parameters to the mplstyle files. I can try to find a way around this, but it would probably be a bit hacky.

cmkohnen (approved):
I feel the whole update_locators function can be done using a custom Locator and simply implementing the `tick_values` function on this to evenly distribute on inverse or square root scale. But this works fine for now.

cmkohnen:
> It doesn't seem possible to add custom parameters to the mplstyle files. I can try to find a way around this, but it would probably be a bit hacky. It will be pretty easy to set such a toggle to the Figure Settings, but it feels more natural to have it in the style sheet together with the other tick settings.

We could extend the format by having lines with our custom settings prefixed with a hashtag, as that would not conflict, if this style were to be loaded outside of Graphs.

Sjoerd1993:
> I feel the whole update_locators function can be done using a custom Locator and simply implementing the tick_values function on this to evenly distribute on inverse or square root scale. But this works fine for now.

Thanks for the suggestion! I actually just went from what is available in matplotlib, and saw that none of the locators really worked well. But based on this comment, I thought I'd just dive into the matplotlib source code and see if I can construct a locator myself.

It's not at all that difficult, in fact it's easier than the implementation in this PR. The basic point is just to implement roughly the same logic I set in update_locators, and couple that to tick_values. The result is much more elegant (no more hacky calls to update_locators when convenient), the code is more flexible and modular and frankly the end result works a lot better as well. 

I've got a PR in the making, I just got the proof of concept working. But I will push a PR with this new implementation later, probably today.

PR #446, opened 2023-09-01
Use custom locator class

Sjoerd1993:
Uses a custom locator class instead of calling a new fixed locator each time the axis is changed. Giving a much more stable and polished experience.

I was a bit unsure whether I should use a seperate class for the inverse and square root case, but it works well enough like this.

### DIFF
--- a/src/scales.py
+++ b/src/scales.py
@@ -10,7 +10,6 @@ custom Scale classes.
         to_int
 """
 from matplotlib import scale, ticker, transforms
-
 import numpy
 
 _SCALES = ["linear", "log", "radians", "squareroot", "inverse"]
@@ -44,8 +43,8 @@ class SquareRootScale(scale.ScaleBase):
     name = "squareroot"
 
     def set_default_locators_and_formatters(self, axis):
-        # Implemented in canvas
-        pass
+        axis.set_major_locator(CustomScaleLocator())
+        axis.set_minor_locator(CustomScaleLocator(scale_type="minor"))
 
     def limit_range_for_scale(self, vmin, vmax, _minpos):
         return max(0, vmin), vmax
@@ -69,7 +68,9 @@ class SquareRootScale(scale.ScaleBase):
         is_separable = True  # Seperable in X and Y dimension
 
         def transform_non_affine(self, a):
-            return numpy.array(a)**2
+            # Don't spam about invalid divide by zero errors
+            with numpy.errstate(divide="ignore", invalid="ignore"):
+                return numpy.array(a)**2
 
         def inverted(self):
             return SquareRootScale.SquareRootTransform()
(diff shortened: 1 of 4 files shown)


src/scales.py
```diff
@@ -107,7 +108,54 @@ def inverted(self):
...
+        return self._numticks
+
+    @numticks.setter
+    def numticks(self, numticks):
+        self._numticks = numticks
+
+    def __call__(self):
+        """Return the locations of the ticks."""
+        vmin, vmax = self.axis.get_view_interval()
+        return self.tick_values(vmin, vmax)
```
cmkohnen, line 143: This is redundant, as the super class has the exact same implementation

src/scales.py
```diff
@@ -107,7 +108,54 @@ def inverted(self):
...
         def transform_non_affine(self, a):
-            return 1 / numpy.array(a)
+            # Don't spam about invalid divide by zero errors
+            with numpy.errstate(divide="ignore", invalid="ignore"):
+                return 1 / numpy.array(a)
+
+
+class CustomScaleLocator(ticker.MaxNLocator):
+    """Dynamically find tick positions on custom scales."""
+    def __init__(self, scale_type="major"):
```
cmkohnen, line 118: scale_type isn't my favourite here. Firstly it describes, what the ticks are so tick_type is a more suitable fit here. Secondly, there are just two types (and likely will be in mpl for the forseeable future) so a boolean like `is_major` would be more suitable imo.

src/scales.py
```diff
@@ -107,7 +108,54 @@ def inverted(self):
...
+            numticks = max(1, self.axis.get_tick_space() - 4)
+            # Amount of ticks is set between 3 and 9
+            self._numticks = numpy.clip(numticks, 3, 9)
+        else:
+            self._numticks = 9
+        if self.scale_type == "minor":
+            # Amount of minor ticks is equal to amount of major ticks
+            # times (N+1) minus N. Where N is the amount of minor ticks
+            # in between the major ticks.
+            self.numticks = len(self.axis.get_majorticklocs()) * 4 - 3
```
cmkohnen, line 133:
```suggestion
            self._numticks = len(self.axis.get_majorticklocs()) * 4 - 3
```

src/scales.py
```diff
@@ -44,8 +43,8 @@ class SquareRootScale(scale.ScaleBase):
     name = "squareroot"
 
     def set_default_locators_and_formatters(self, axis):
-        # Implemented in canvas
-        pass
+        axis.set_major_locator(CustomScaleLocator())
+        axis.set_minor_locator(CustomScaleLocator(scale_type="minor"))
```
cmkohnen, line 47: This is now missing a formatter

cmkohnen (changes requested):

cmkohnen (approved):

PR #447, opened 2023-09-01
Adds more trignometric functions

Sjoerd1993:
After seeing @veggero fumble around in a video trying to insert a cotangent, I thought it would be good to add these extra trigonemetric functions as well. This PR adds support for cotangent, secant and cosecant.

See video [here](https://www.youtube.com/watch?v=nsz-jdP9xvg&t=359s) starting, around 5:14.


veggero:
Thanks!

cmkohnen (approved):

PR #449, opened 2023-09-01
improve optimize limits

cmkohnen:
- hold local figure_settings reference
- reduce number of loops
- reduce duplicated logic
- fix linting

### DIFF
--- a/src/utilities.py
+++ b/src/utilities.py
@@ -48,7 +48,7 @@ def rgba_to_hex(rgba):
 def rgba_to_tuple(rgba, alpha=False):
     if alpha:
         return (rgba.red, rgba.green, rgba.blue, rgba.alpha)
-    return (rgba.red, rgba.green, rgba.blue)
+    return (rgba.red, rgba.sqrt_valuegreen, rgba.blue)
 
 
 def swap(str1):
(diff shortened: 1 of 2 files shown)


src/utilities.py
```diff
@@ -199,49 +197,47 @@ def get_filename(file: Gio.File):
...
-            axes[index][2].append(item)
-
-    for count, (direction, used, items) in enumerate(axes):
+            data = numpy.asarray(item.ydata if index % 2 else item.xdata)
+            nonzero_data = list(filter(lambda x: (x != 0), data))
+            axes[index][2].append(
+                nonzero_data[numpy.isfinite(nonzero_data)].min()
+                if axes[index][4] in (1, 4) and len(nonzero_data) > 0
+                else data[numpy.isfinite(data)].min(),
+            )
```
Sjoerd1993, line 217: I forgot to convert non_zero data to a numpy array, meaning it will error out on line 214 if conditions of line 215 are met.

Probably better to filter out the inf values centrally first on data, instead of on data and nonzero data infinitely. We don't want to use that for finding the limits anyway.

```suggestion
            data = numpy.asarray(item.ydata if index % 2 else item.xdata)
            data = data[numpy.isfinite(data)]
            nonzero_data = numpy.array([value for value in data if value != 0])
            axes[index][2].append(
                min(nonzero_data)
                if axes[index][4] in (1, 4) and len(nonzero_data) > 0
                else min(data)
            )
```

Sjoerd1993 (changes requested):

Sjoerd1993 (approved):

PR #453, opened 2023-09-02
Improve style functions

cmkohnen:
Try to optimize style related functions.
- Don't loop over system styles as often as we know the exact path of files
- Make system style a setting (not to end users)

The last step is the base for an upcomming feature I try to add: Yaru style for snap packaging

### DIFF
--- a/data/se.sjoerd.Graphs.gschema.xml
+++ b/data/se.sjoerd.Graphs.gschema.xml
@@ -1,5 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <schemalist gettext-domain="graphs">
+  <enum id="se.sjoerd.Graphs.system-styles">
+    <value nick="adwaita" value="0"/>
+  </enum>
+
   <enum id="se.sjoerd.Graphs.general.center-values">
     <value nick="Center at maximum Y value" value="0"/>
     <value nick="Center at middle coordinate" value="1"/>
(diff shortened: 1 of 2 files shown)


Sjoerd1993 (approved):
Using a separate Yaru style would work fine. In principle I was thinking of obtaining the color at launch time and set it in the system theme then, this way it would still look good when changing everything manually with Gradiance for example, but it would be a bit more messy. 

It would at least be good to support at least the "official" launches, and solve this color discrepancy in the snap version.

cmkohnen:
> Using a separate Yaru style would work fine. In principle I was thinking of obtaining the color at launch time and set it in the system theme then, this way it would still look good when changing everything manually with Gradiance for example, but it would be a bit more messy.

This was one of the first things I tried a couple of months ago. The problem is, that when the style changes and we connect to the `Adw.StyleManager` the new styles aren't available yet, so we'd always show the wrong style (dark/light)

Sjoerd1993:
> The problem is, that when the style changes and we connect to the Adw.StyleManager the new styles aren't available yet

Yes, I've came upon the exact same problem. My thought was to work around this by getting the correct colors upon launch, and saving this in the system style sheet each time. However there's some problems with that approach (apart from the general hacky nature, and it being unnecessary overhead). The main one being that you can only retrieve the properties of the current active theme, so if a user would start in dark mode, it cannot get the properties for the light mode and vice versa.

As long as changing the system colours is not supported in vanilla GNOME though, I don't think it's a really big issue that we should support. If anyone ever files an issue, we can just point them to the custom style sheets. They can just create a custom style sheet that fits with their colours, and apply that in that case.

PR #456, opened 2023-09-05
Suggestion: Use Vala for more GObject centric classes

cmkohnen:
This is a proof of concept in using the Vala language for classes that make heavy use of GObject based functions.

In this pr I've ported the FigureSettings class.

There are multiple benefits to this:
- Performance gain through compilation to binary
- Compile time errors and warnings just like blueprint, which could prevent regressions
- Type safety even for arrays and other more complex types
- Automatic generation of getter and setter methods
- More readable syntax for Properties and Signals
- Probably some funky stuff, that's complex to implement in python
- (If somehow needed: easier and more reliable access to c code)

This can be seen as just a suggestion and if this would be out of scope, I'm absolutely fine with that.

### DIFF
--- a/Graphs.doap
+++ b/Graphs.doap
@@ -27,6 +27,7 @@
   <bug-database rdf:resource="https://github.com/Sjoerd1993/Graphs/issues" />
 
   <programming-language>Python</programming-language>
+  <programming-language>Vala</programming-language>
 
   <platform>GTK 4</platform>
   <platform>Libadwaita</platform>
--- a/meson.build
+++ b/meson.build
@@ -1,9 +1,17 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
-project('graphs',
+project('graphs', 'c', 'vala',
           version: '1.6.1',
     meson_version: '>= 0.59.0',
   default_options: [ 'warning_level=2', 'werror=false', ],
 )
+
+dependencies = [
+  dependency('glib-2.0'),
+  dependency('gobject-2.0'),
+  dependency('gio-2.0'),
+  dependency('gtk4'),
+  dependency('libadwaita-1', version: '>= 1.2'),
+]
 
 application_id = 'se.sjoerd.Graphs'
 name = 'Graphs'
--- a/src/data.py
+++ b/src/data.py
@@ -194,25 +194,25 @@ class Data(GObject.Object):
                 original_position = new_item.xposition
                 if new_item.xposition == 0:
                     if _is_default("bottom-label"):
-                        figure_settings.props.bottom_label = new_item.xlabel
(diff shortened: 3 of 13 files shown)


Sjoerd1993:
I'm honestly not familiar with Vala (in the sense that I haven't coded in it, I do know what it is), but it is a language that I would like to get familiar with and that I know works incredibly well with GNOME and GTK in particular. I am also well aware of the performance benefits of using a compiled language like Vala or C. 

Furthermore, the GTK Python bindings are not [in the best state](https://www.bassi.io/articles/2022/12/02/on-pygobject/) either. 

The reason for starting all this in Python has always been convenience first, it's the language I'm most familiar with. I'm not really advocating to port everything to Vala directly (or any other language for that matter). But I am not immediately opposed to move certain parts of the code-base to Vala where that makes sense at least for now. So I think this here is change that could make sense. It can make life easier for parts that are more GObject centric, again particularly because PyGObject does have its issues either way.

I'll read through the code itself first though before merging :)

cmkohnen:
> I'm not really advocating to port everything to Vala directly (or any other language for that matter). But I am not immediately opposed to move certain parts of the code-base to Vala where that makes sense at least for now.

I doubt that would be possible in the case of Graphs. As far as I understand, we can generate classes in Vala and let them interact with each other. But the python bridge is one-way, so we can only call classes implemented in Vala from python not vice versa. This means, that the whole matplotlib interface has to stay in python as well as the application init.
Also numpy and scipy are great libraries for mathematical data centric use-cases, so I'd advocate we continue using them.

My point here being, that I have no intentions to convert the whole codebase, just the part that makes sense to do so.

Sjoerd1993:
> Also numpy and scipy are great libraries for mathematical data centric use-cases, so I'd advocate we continue using them.

Absolutely, and also Matplotlib which Graphs heavily leans on, is a fantastic library with active developments and tons of features. Implementing this in another language is a massive task with likely questionable results. I've looked a bit into this at one point, and while GNUPlot for instance is quite good, nothing really compares to Matplotlib. It's also quite easily adaptable I have learned. So the libraries available in Python are really just perfect for Graphs and in many cases the sole reason why certain features can be implemented really quickly. Same would be true for fitting features for example, which can be done with SciPy. Another advantage of Python is that it's one of the most popular languages out there. So both support, but also potential contributors, are probably easier to get than e.g. Vala. 

Just wanted to stipulate that praising compiled languages should not be seen as a call to move everything towards such a language. But again, it does make sense in this case to use Vala, so I think the proposal itself here is good.

Sjoerd1993:
I took the liberty of fixing the window.py calls to Figure settings that were still in the old style. Otherwise this seems to work good. Overall this results in a nicer syntax in general I feel.

Sjoerd1993 (approved):

PR #459, opened 2023-09-07
Use item popover

sstendahl:
Will try and add a popover menu for actions for the items. Will probably include:
Move up
Move down
Edit Item
Remove item
Curve fitting (will be added later in a seperate PR)

Will try to see if I can add GTK callbacks to the menus, or if I need to use actions.

### DIFF
--- a/src/data.py
+++ b/src/data.py
@@ -92,10 +92,6 @@ class Data(GObject.Object):
         """Get all managed items."""
         return list(self._items.values())
 
-    def get_item_by_index(self, index: int):
-        """Get item given an index."""
-        return list(self._items.values())[index]
-
     def set_items(self, items: list):
         """Set all managed items."""
         for item_ in items:
--- a/src/item_box.py
+++ b/src/item_box.py
@@ -89,7 +89,7 @@ class ItemBox(Gtk.Box):
     def move_up(self, _action, _shortcut, item):
         data = self.get_application().get_data()
         before_index = data.index(item.key)
-        top_data = data.get_item_by_index(before_index - 1)
+        top_data = data[before_index - 1]
 
         data.change_position(item.key, top_data.key)
         clipboard = self.get_application().get_clipboard()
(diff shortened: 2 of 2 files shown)


Sjoerd1993:
This should be ready now, see screencast for the behaviour of this PR.
Note, this builds on PR #462 as well to get the behaviour to be correct.

src/data.py
```diff
@@ -92,6 +92,10 @@ def get_items(self) -> list:
         """Get all managed items."""
         return list(self._items.values())
 
+    def get_item_by_index(self, index: int):
+        """Get item given an index."""
+        return list(self._items.values())[index]
+
     def set_items(self, items: list):
```
cmkohnen, line 99: Use regular access instead, see line 138

data/ui/item_box.blp
```diff
@@ -43,13 +32,39 @@ template $ItemBox : Box {
...
     styles ["flat"]
-    clicked => $delete();
-    Image {
-      hexpand: false;
-      icon-name: "edit-delete-symbolic";
-    }
   }
 }
+
+PopoverMenu popover_menu {
```
cmkohnen, line 42: I'm just wondering if it would be possible to use icons here and have move up and down horizontal to each other, wouldn't consider this a blocker however

cmkohnen (changes requested):

src/data.py
Sjoerd1993: You are right, thanks, completely missed this.

data/ui/item_box.blp
Sjoerd1993: I think that would likely be an improvement, perhaps we can do something similar to the icon size in Nautilus:

image

However, I cannot find any reference of how this would be done [in the documentations](https://docs.gtk.org/gtk3/class.PopoverMenu.html). It doesn't seem a GtkBox is supported as  a child to GtkPopoverMenu. Will try to see if I can find a solution later on, but we could do that as a separate PR perhaps.

cmkohnen (approved):

PR #460, opened 2023-09-07
move application properties to vala and update build system

cmkohnen:
It took me literally days to figure out, why the vala compiler wouldn't work with composite templates. The solution was that he was looking for .ui files in data/ui where only .blp files (not yet compiled) exist. The solution was to copy the gresource xml into the build directory, so the vala compiler would have access to the compiled ui files.

This includes exactly that plus moving the applications properties to vala. This also includes the proof of concept on how to handle implementations written in python inside vala code. We simply define interfaces in vala, with functions we like to access, and add them as inheritance to the python class. This way we'd have to change nothing on the python side while still being able to access functions in vala code.
Another solution, like the one being employed here is to simply define the class and all properties and even some logic in vala and simply let the currently existing python class inherit from that.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -72,14 +72,19 @@ blueprints = custom_target('blueprints',
   command: [find_program('blueprint-compiler'), 'batch-compile', '@OUTPUT@', '@CURRENT_SOURCE_DIR@', '@INPUT@'],
 )
 
-resources = gnome.compile_resources(
-  application_id,
-  application_id + '.gresource.xml',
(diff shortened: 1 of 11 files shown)


Sjoerd1993:
Line 107 in `file_io` was still using the old settings instead of the getter, making it impossible to import column files. I fixed that with a direct commit if that's ok. Looks and works good, putting these classes towards Vala will make things a lot easier in the long term.

Sjoerd1993 (approved):

PR #461, opened 2023-09-08
implement PreferencesWindow in Vala

cmkohnen:
Continues https://github.com/Sjoerd1993/Graphs/pull/460

One neat aspect of this, is that errors, that would previously show up on runtime (e. g. callback or widget name mismatch) are now picked up during compile time.
Also Composite widgets can behave like actual properties so we can simply query a property instead of accessing a _propertyhelper attribute which points all methods to an object stored somewhere else like currently in python


cmkohnen:
Welp seems like meson just can't be bothered to execute clearly marked target orders in the right order

Sjoerd1993:
Kinda weird, when I press the build button locally, I get an error (same as in Github actions), but as soon as I just press "Run Project", it works fine. 

Either way, I think line 107 in `file_io.py` should be  `columns_params = self.get_settings().get_child(`. Not sure what's wrong with meson here.

cmkohnen:
> Kinda weird, when I press the build button locally, I get an error (same as in Github actions), but as soon as I just press "Run Project", it works fine.
Apparently both compiling the resources and the actual source happen simultaneously, so when you run the project a second time they are already cached and the vala compiler can actually pick them up

> Either way, I think line 107 in `file_io.py` should be `columns_params = self.get_settings().get_child(`. Not sure what's wrong with meson here.
Haven't merged with main yet.

Sjoerd1993 (approved):

cmkohnen:
> It sure is a bit of a hack, but it does work and I don't really have a cleaner solution (I've been trying, in principle declaring a module as a dependency, should make it wait for that dependency to compile before starting both in parallel, but didn't have much success there.).

Yes, I guess the issue here is, that meson uses the input file of the gresource getting the dependency tree wrong in all sorts of manners. This is adapted from [this issue](https://gitlab.gnome.org/jwestman/blueprint-compiler/-/issues/18)

PR #462, opened 2023-09-08
Fix item change implementation

Sjoerd1993:
The reason for reverting to an item menu reload after changing items, is that the current behaviour is extremely buggy. See screencast:

[Screencast from 2023-09-08 10-12-01.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/7a4d10c1-0e59-4988-8ccc-a44cfe8aa90e)

### DIFF
--- a/src/data.py
+++ b/src/data.py
@@ -151,7 +151,7 @@ class Data(GObject.Object):
         else:
             items[index2:index1 + 1] = \
                 [self._items[key2]] + items[index2:index1]
-        self.set_items(items)
+        self.items = items
 
     def add_items(self, items: list):
         """


src/data.py
```diff
@@ -151,7 +151,7 @@ def change_position(self, key1: str, key2: str):
         else:
             items[index2:index1 + 1] = \
                 [self._items[key2]] + items[index2:index1]
-        self.set_items(items)
+        self.items = items
```
cmkohnen, line 154: Ah, i see the issue, your suggestion boils down to the exact same method again with the overhead of going through a property annotator (which is quite a lot relatively).

The problem with `set_items()` is that no Signal listeners are notified. This can be fixed properly by adding `self.notify("items")` to said method

Sjoerd1993: It shouldn't go through any property annotator here, as `self.items` is not called here, only set. So it line 154 only calls the setter (line 89), as it sets `self.items`. This is also confirmed by some simple print statements.

I think the signal listeners are notified though, as the setter calls set_items, where the signal listener is notified upon the append action.

cmkohnen: Huh, maybe I'm misunderstanding something here, but if it works...

cmkohnen (approved):

PR #465, opened 2023-09-08
Use horizontal buttons for move up/down in popover

Sjoerd1993:
This was slightly easier than I thought, just simply looked how Nautilus did it in their GitLab repo. The trick is to set the buttons in a single section, and simply the inline buttons, this is even stated in their documentation:

`“display-hint”: a string used to determine special formatting for the section. Possible values include “horizontal-buttons”, “circular-buttons” and “inline-buttons”. They all indicate that section should be displayed as a horizontal row of buttons.`


cmkohnen (approved):

Sjoerd1993:
Thanks for the suggestion here, this looks a lot better than having these very similar actions on separate rows.

PR #471, opened 2023-09-09
move window to Vala

cmkohnen:


src/application.py
```diff
@@ -71,7 +70,7 @@ def __init__(self, application_id, **kwargs):
             action.connect("activate", lambda action_, target:
                            (figure_settings.set_property(
                             action_.get_name()[7:], int(target.get_string())),
-                            self.get_window().toast_overlay.grab_focus()))
+                            self.get_window().get_canvas().grab_focus()))
```
Sjoerd1993, line 73: Setting the focus on the canvas makes the shortcuts inaccesable again after changing scale from the view menu.

Sjoerd1993 (approved):
Now this works well, can't see any issues arising with my testing.

Issue #246, opened 2023-04-24
[suggestion] Add linear regression in Mathematical Operations


matbme:
First of all, I'd like to thank you for this amazing application. You have already saved me countless hours by not having to figure out how to properly adjust Matplotlib and Seaborn in code :D

One feature I think would be nice to have is the ability to generate a linear (and maybe even quadratic) regression line for the data in a specific dataset, probably by using the least squares fit from either [Numpy](https://numpy.org/doc/stable/reference/generated/numpy.linalg.lstsq.html) or [Scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.lstsq.html#scipy.linalg.lstsq).

Sjoerd1993:
Hey,

Thanks for the kind words. This is actually a feature that is on my radar and was the next "proper" thing I was thinking of implementing.

I've done this code-wise with scipy in the past and that works quite well. Scipy is already a dependency of Graphs anyway, so it's a perfect fit for this feature.

Right now I'm in the middle of finishing my PhD thesis (printing starts on the 2nd of May), so at the moment my time is rather limited. But I will take a look at this next week.

The biggest open question is more about the UX-side of this. But I've got some ideas there. In principle it should be possible to get a nice window where you can specify a custom function. And then put the coefficients in the equation name when it's added to the data set. 

Either way, it is in my radar. I'll take a look at it next week after I've handed in my thesis :)

cmkohnen:
Hey, if this is still on the radar, I might have come up with some design suggestions.

I'd propose a dialogue window with "live updates". Generally a dialogue could be invoked via the add menu or a new context menu for items.
Introducing a context menu for items could also add features such as move up/down or export just this item.

The dialogue should contain (at the least) information about the actual equation "being used" to generate a regression (e. g. `1.56x + 6.7` for linear regressions) as well as the type (mx + b, ax² + bx + c, etc.). The latter might be changed on the fly by the user.
Also including the sum of R² would definitely be useful here.
It should also be possible to specify the name of that newly generated item (empty by default and using the equation name if it stays empty).

The actual item should only be generated and added, when that dialogue is confirmed.

Sjoerd1993:
This is definitely still on the radar, and I was hoping to implement this with the 1.7 release. 

I was actually planning to properly get started with this, this week or so. Live updates would be very useful, as it would give a nice and intuitive way to get good initial values for the fitting process.

[...]

PR #475, opened 2023-09-11
Curve fitting

Sjoerd1993:
Adds curve fitting functionalities to Graphs, see screencast:
[Screencast from 2023-09-11 17-19-53.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/03feaae8-e0f0-464c-8e23-7adc878876b9)

Note that this is basically just a coarse proof-of-concept. Part of the regular code needs to be adjusted slightly to make this work (although it's all modular enough that this is not that much), and the current "new code" in this PR is in a horrible state, and not at all representative of the final PR. It's just the very first quick proof of concept. At the moment it's also hard-coded to only do a linear regression (in this PR, I intend to already implement arbritrary functions), and the dialog itself is mostly a placeholder. I built from the export figure dialog, hence the widgets that are visible.

Putting the PR up already now is basically just to allow for feedback and suggestions on e.g. the UI and what to expose to the user in the first implementation of this functionality. Suggestions on code itself are welcome, just be aware that the current code itself is known to be awful.

Planned functionalities include at least:
- Specify equation to fit to
- Some statistical values, or at least some FoM measure
- Import fitted curve as item

Setting boundaries and initial guess for the parameters should be added as well, not 100% sure if that will be part of this PR or be added in later on.

Once this is ready and merged (again still early stage, won't be today, likely not tomorrow either), this should close issue #246


Sjoerd1993:
Currently I'm mainly trying to come up with a way to evaluate an arbitrary string output as a callable function. Scipy's curve fitting is relatively simple. It should be quite clear from the example in this PR, but to summarize it basically works like this:

`parameters, coveriance = curve_fit(equation, xdata, ydata).`

Then the parameters simply output the fitted values for the fitting parameters in a list, and covariance is just the statistics.
The equation itself needs to be a callable, so in the current branch here it's called as such:

```
def equation(x, a, b):
    return a*x+b
```
    
So scipy then recognizes the arguments. The first argument in the callable function is the x value, and the other arguments are the fitting parameters. 

But what I want, is for the user to able to just type in an arbritrary function. So they can fit for example a Gaussian, or any random polynomial. This would be done in an AdwEntryRow, so this will be a string. So what I now need is a way to translate a string like "a*x^2 + b*x + c" for example into a function like so:
```
def equation(x, a, b, c):
    return a*x**2+b*x+c
```

Translating the ^ to ** is no issue, we've got the preprocessing utility for that. But the major headache is mostly the translation from the string into a callable function. This is all relatively easy with an eval function, but those I really want to avoid after we've gone through the hassle of removing them. Perhaps I'll try and see if I can "write" a closed/safe kind of eval function for this. But I'm open for smart ideas. I'll probably have time to take another look at this in the afternoon after working hours.

cmkohnen:
What if we don't allow inputting an arbitrary equation but offer a choosable list instead? This is what most tools I know do.

If we do go with an arbitrary string however, perhaps using AST again could solve the issue.

Sjoerd1993:
Eventually, I'd like to have the possibility at least to have arbritrary equations. But I'm also open to the idea of having a choosable list initially, and work out support for arbritrary equations later.

However, I've been stumbling around, and I've got a pretty well-working version ready, but it does require and extra dependency. Parsing an equation with sympy that way, also becomes quite clean. So that works all pretty well, and I think it can be included in here. See working prototype:

[Screencast from 2023-09-12 15-03-14.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/75713d8e-2ef6-428b-af69-663c7ae03eee)

 Perhaps we can provide a list with predefined options, with one of the options being "Custom equation", which triggers the visibility of an entry row. This is similar to what Matlab does iirc.

I did notice during testing that the possibility to set starting values is pretty important. So at least the possibility to set a starting guess should be added as well.

Sjoerd1993:
It now also automatically changes the fit when the equation is changed. As the "Fit" button is no longer needed, it's been replaced by a button to export the fit to the data. It should also expose entry rows when more parameters are added, to set the starting guesses, and eventually also boundaries.

However, I am not 100% sure yet how this automatic behaviour should be coupled with these starting guesses. The issue I'm thinking of is the situation you want to set the starting guess for 3 different parameters. If it's auto-updating, and you want to set the initial value of parameter a, then the values will change for parameter b and c. So if you then want to set the initial guess for parameter b, the value for parameter a will suddenly change etc...

Perhaps it's possible to have it so that the "initial guess" is decoupled from the current active value. So to not make it auto-update when changing the initial guess. But then we'd have to introduce a new "fit" button again. Which is possible, but it requires some thinking UI-wise at least.

Also, again there's still some technical issue with the entire code. Such as the coupling between both canvases, the fact that the clipboard is broken for the fits, and the general messyness of the code. Feel free to add PR's to this branch (pushing directly may lead to conflicts). But my general strategy right now was to first get the functionality in place, and then worry about the state of the code itself.

[Screencast from 2023-09-12 16-26-11.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/c95a2050-dc2b-4a96-9dfa-3abfe165982f)

Sjoerd1993:
Final update for now, I've cleaned up the UI itself a bit. Right now it just adds entry rows according to the fitting variables, and inserts the fitted values in those rows. Changing the values in those entry rows doesn't really do anything though.

One idea is to have each row consist of three rows, where each entries for "value", "lower bound", "upper bound". Performing a fit will change this value, but keep it in between the bounds. 

After that all works, I'll try and think about the implementation with auto-updating. Perhaps it'll be needed to not let it automatically do a fit when changing these entry rows, but again that would mean adding a "Fit" button again. Perhaps below the Equation entry, above these entry rows. And then there's of course the code clean up and such.

[Screencast from 2023-09-12 17-25-46.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/a43bbc5b-e1aa-48c4-aab6-8e87df836367)

Sjoerd1993:
Okay final update for today for real, here's what it works like if a change in the entry row automatically triggers a fit. It works, and in this situation I was able to get a fit to the sine function which was not possible with the default settings. 

But it doesn't feel nice that the value is immediately "set". Especially with the first case where the fit is already in the ideal place with the smallest FoM.

[Screencast from 2023-09-12 19-47-05.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/6fca1bf5-4393-4d56-8c23-0afcf6682a70)

Sjoerd1993:
So now this has got an actually useable interface, still need to fix some issues. Also probably want to add bounds to the fits as well still within this PR maybe, but that can maybe be added later in a seperate PR as we probably want to iterate a bit anyway:

[Screencast from 2023-09-13 10-42-20.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/2e954605-d86f-441c-b38a-69c809d9c8cc)

Sjoerd1993:
So there's some more tweaks I can imagine. For instance, some statistics should be reported in the "Results" section. I also want to give some sort of feedback when an initial value is set that is outside of the span defined by the min and max bounds. But I've landed upon a design that I'm mostly happy with as initial implementation. Said additions should probably be part of another PR, as to limit the scope of this one somewhat.

See screencast, apologies for the slow typing as we just got a new keyboard which I need to get used to.

[Screencast from 2023-09-14 12-17-26.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/eb5440c3-3fad-4b1c-9a67-b3ab7d6fd15c)

data/se.sjoerd.Graphs.gschema.xml
```diff
@@ -155,6 +156,12 @@
     </key>
   </schema>
 
+  <schema id="se.sjoerd.Graphs.curve-fitting">
+    <key name="equation-entry" type="s">
```
cmkohnen, line 160: perhaps just "equation"?

data/ui/curve_fitting.blp
```diff
@@ -0,0 +1,104 @@
...
+        ScrolledWindow {
+          vexpand: false;
+          height-request: 370;
+          hscrollbar-policy: never;
+        Box {
+          vexpand: false;
+          orientation: vertical;
+          margin-start: 12;
+          margin-end: 12;
+        Box fitting_params {
```
cmkohnen, line 61: Starting on this line the indentation is a bit wonky

data/ui/fitting_parameters.blp
```diff
@@ -0,0 +1,73 @@
+using Gtk 4.0;
+using Adw 1;
+
+template $FittingParameterEntry : Box {
+  orientation: vertical;
+  Adw.PreferencesGroup {
```
cmkohnen, line 6: same here

src/curve_fitting.py
```diff
@@ -0,0 +1,272 @@
...
+from graphs.utilities import preprocess, string_to_function
+
+import numpy
+
+from scipy.optimize import curve_fit
+
+
+@Gtk.Template(resource_path="/se/sjoerd/Graphs/ui/curve_fitting.ui")
+class CurveFittingWindow(Adw.Window):
+    __gtype_name__ = "CurveFittingWindow"
```
cmkohnen, line 20:
```suggestion
    __gtype_name__ = "GraphsCurveFittingWindow"
```

data/ui/curve_fitting.blp
```diff
@@ -0,0 +1,104 @@
+using Gtk 4.0;
+using Adw 1;
+
+template $CurveFittingWindow : Adw.Window {
```
cmkohnen, line 4:
```suggestion
template $GraphsCurveFittingWindow : Adw.Window {
```

src/curve_fitting.py
```diff
@@ -0,0 +1,272 @@
...
+    equation_entry = Gtk.Template.Child()
+    fitting_params = Gtk.Template.Child()
+    toast_overlay = Gtk.Template.Child()
+    text_view = Gtk.Template.Child()
+
+    def __init__(self, application, item):
+        super().__init__(
+            application=application, transient_for=application.get_window(),
+
```
cmkohnen, line 29:
```suggestion
```

src/curve_fitting.py
```diff
@@ -0,0 +1,272 @@
...
+        )
+        ui.bind_values_to_settings(
+            self.get_application().get_settings("curve-fitting"), self)
+        canvas = Canvas(application)
+        self.param = []
+
+        # Set data item that contain the curves
+        self.curves = Data(application)
```
cmkohnen, line 37: would a simple list suffice here? Using the whole Data class seems a bit overkill

src/curve_fitting.py
```diff
@@ -0,0 +1,272 @@
...
+        xdata = numpy.linspace(
+            min(self.curves[1].xdata),
+            max(self.curves[1].xdata),
+            5000)
+        ydata = [function(x, *self.param) for x in xdata]
+
+        name = _get_equation_name(
+            str(self.equation_entry.get_text()), self.param)
+        self.curves[0].name = f"Y = {name}"
+        self.curves[0].ydata, self.curves[0].xdata = (ydata, xdata)
```
cmkohnen, line 181: this would just mean adding a canvas.set_items(curves) here

src/utilities.py
```diff
@@ -241,3 +243,17 @@ def optimize_limits(self):
...
+        r"\b(?!x\b|X\b|sin\b|cos\b|tan\b)[a-wy-zA-WY-Z]+\b",
+        equation_name,
+    )
+    sym_vars = sympy.symbols(variables)
+    try:
+        symbolic = sympy.sympify(equation_name,
+                                 locals=dict(zip(variables, sym_vars)))
+        return sympy.lambdify(sym_vars, symbolic)
+    except (sympy.SympifyError, TypeError):
+        return
```
cmkohnen, line 259:
```suggestion
    with contextlib.suppress(sympy.SympifyError, TypeError):
        symbolic = sympy.sympify(
            equation_name, locals=dict(zip(variables, sym_vars)),
        )
        return sympy.lambdify(sym_vars, symbolic)
```

src/curve_fitting.py
```diff
@@ -0,0 +1,272 @@
...
+        self.fit_curve()
+        self.set_entry_rows()
+
+        # Set figure settings
+        figure_settings = \
+            Graphs.FigureSettings.new(
+                application.get_settings().get_child("figure"))
+        for prop in dir(figure_settings.props):
+            if prop not in ["use_custom_style", "custom_style"]:
+                figure_settings.bind_property(prop, canvas, prop, 1 | 2)
```
cmkohnen, line 67: The figure settings aren't used any further any reason to create them?

src/curve_fitting.py
```diff
@@ -0,0 +1,272 @@
...
+        buffer_string = initial_string
+        for index, arg in enumerate(self.get_free_variables()):
+            buffer_string += f"\n {arg}: {self.param[index]}"
+
+        self.text_view.get_buffer().set_text(buffer_string)
+        bold_tag = Gtk.TextTag(weight=700)
+        self.text_view.get_buffer().get_tag_table().add(bold_tag)
+
+        start_iter = self.text_view.get_buffer().get_start_iter()
+        end_iter = self.text_view.get_buffer().get_start_iter()
```
cmkohnen, line 140: this also uses get_start_iter()?

src/curve_fitting.py
```diff
@@ -0,0 +1,272 @@
...
+
+    @Gtk.Template.Callback()
+    def add_fit(self, _widget):
+        """Add fitted data to the items in the main application"""
+        self.curves[0].color = ""  # Reset color, so it is obtained from cycle
+        self.get_application().get_data().add_items([self.curves[0]])
+        self.destroy()
+
+    def set_entry_rows(self):
+
```
cmkohnen, line 193:
```suggestion
```

src/curve_fitting.py
```diff
@@ -0,0 +1,272 @@
...
+    """Class to contain the fitting parameters."""
+    __gtype_name__ = "FittingParameters"
+    __gsignals__ = {}
+
+    def add_items(self, items):
+        for item in items:
+            self._items[item.name] = item
+
+    def remove_unused(self, used_list):
+
```
cmkohnen, line 211:
```suggestion
```

src/curve_fitting.py
```diff
@@ -0,0 +1,272 @@
...
+    def get_bounds(self):
+        lower_bounds = [float(item_.lower_bound) for item_ in self]
+        upper_bounds = [float(item_.upper_bound) for item_ in self]
+        return (lower_bounds, upper_bounds)
+
+
+class FittingParameter(GObject.Object):
+    """Class for the fitting parameters."""
+
+    __gtype_name__ = "FittingParameter"
```
cmkohnen, line 240:
```suggestion
    __gtype_name__ = "GraphsFittingParameter"
```

src/curve_fitting.py
```diff
@@ -0,0 +1,272 @@
...
+        while self.fitting_params.get_last_child() is not None:
+            self.fitting_params.remove(self.fitting_params.get_last_child())
+
+        for arg in self.get_free_variables():
+            self.fitting_params.append(FittingParameterEntry(self, arg))
+
+
+class FittingParameters(Data):
+    """Class to contain the fitting parameters."""
+    __gtype_name__ = "FittingParameters"
```
cmkohnen, line 203:
```suggestion
    __gtype_name__ = "GraphsFittingParameters"
```

data/ui/fitting_parameters.blp
```diff
@@ -0,0 +1,73 @@
+using Gtk 4.0;
+using Adw 1;
+
+template $FittingParameterEntry : Box {
```
cmkohnen, line 4:
```suggestion
template $GraphsFittingParameterEntry : Box {
```

src/curve_fitting.py
```diff
@@ -0,0 +1,272 @@
...
+        """Init the dataclass."""
+        self.name = name
+        self.initial = initial
+        self.lower_bound = lower_bound
+        self.upper_bound = upper_bound
+
+
+@Gtk.Template(resource_path="/se/sjoerd/Graphs/ui/fitting_parameters.ui")
+class FittingParameterEntry(Gtk.Box):
+    __gtype_name__ = "FittingParameterEntry"
```
cmkohnen, line 254:
```suggestion
    __gtype_name__ = "GraphsFittingParameterEntry"
```

src/curve_fitting.py
Sjoerd1993: I'm mainly using the Data class to bind these to the canvas that is used:
`        self.curves.bind_property("items", canvas, "items", 2)`

If you have a suggestion to bind the canvas to the used items without doing this, then that would probably be the preferred solution, since the other properties of this class are not really used.

src/curve_fitting.py
Sjoerd1993: So I've been trying this as a way to update the canvas upon fits, but without any success. There's no refresh being triggered.

src/curve_fitting.py
Sjoerd1993: This was left-over code in an attempt  to couple the figure style to the main canvas without coupling the other settings. The styles are carried over properly when initiating the window, but when changing style from dark to light and vice verse, the plot style is still kept in the original mode. This is something that still needs to be added to this code.

src/curve_fitting.py
Sjoerd1993: Yes, it starts both the start and the end at the first character. Then it moves the end position forwards until the first word is covered. This part is put in bold. It's basically a roundabout way to set the "Results" word in bold. Haven't found a cleaner solution, good chance that one exists but didn't have much success there.

data/ui/curve_fitting.blp
```diff
@@ -0,0 +1,104 @@
...
+        orientation: vertical;
+        vexpand: false;
+        width-request: 300;
+        Adw.EntryRow equation {
+          margin-top: 12;
+          margin-bottom: 12;
+          margin-start: 12;
+          margin-end: 12;
+          title: _("Equation");
+          text: _("a*x+b");
```
cmkohnen, line 46:
```suggestion
```
This should be set by using the preference

src/curve_fitting.vala
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+using Adw;
```
cmkohnen, line 2: were not using Adw here, or am I wrong?

data/ui/curve_fitting.blp
Sjoerd1993: You are right, thanks. This was still left from a previous working build, forgot to remove this.

src/curve_fitting.vala
Sjoerd1993: Correct, not sure why that was there tbh. Probably left-over from using another class as template.

cmkohnen (approved):
This works great from my testing. Visually this works well with the direct visualization of the canvas.
Perhaps we should implement a feature to lock the canvas, so navigating and invoking the figure settings can be blocked. However that can be part of a different pr.

Also maybe having a list of default equations would be nice, but that isn't a necessity either

PR #477, opened 2023-09-12
Complete transition from key to uuid

Sjoerd1993:
I noticed a warning that the uuid cannot be set when going into the item edit window. So I went through the code to check where key was still used instead of the uuid attribute, and found a few instances.

The reason for the warning by the way was because the IGNORELIST when binding the values to the objects in the window, still contained key instead of the uuid. But this PR includes a few other places as well where key was still used.


cmkohnen:
Oh must have missed these, my bad

cmkohnen (approved):

PR #478, opened 2023-09-12
remove get_next_color()

cmkohnen:
Remove get_next_color() from utilities. When adding a number of items, this used to loop quite often. Also during reset. This changes that to use the already existing loop structures

### DIFF
--- a/src/data.py
+++ b/src/data.py
@@ -11,6 +11,8 @@ from gi.repository import GObject, Graphs
 
 from graphs import item, utilities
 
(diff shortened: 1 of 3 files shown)


Sjoerd1993 (approved):
This makes a lot of sense, I was going through this the other day when working on the curve fitting stuff and was already thinking utilities is a bit of an awkard place for this logic.

This method is strongly related to the data set, so I'd agree this is definitely where it belongs.

PR #479, opened 2023-09-12
Split file parsing

cmkohnen:
Split of file parsing to separate file.
Also clean up general import logic:
- get rid of ImportSettings class
- only initialize window if there are modes to be configured


Sjoerd1993 (approved):
Overall an improvment organisationally.

PR #481, opened 2023-09-13
Refactor

cmkohnen:
- Update gtype names and classnames to better fit GObject naming
conventions
- Move Item to vala and use proper getter and setter methods
- Add license headers
- make data.change_position take indeces instead of keys
- require item for data.index()

### DIFF
--- a/src/application.py
+++ b/src/application.py
@@ -26,13 +26,14 @@ _ACTIONS = [
 ]
 
 
-class GraphsApplication(Graphs.Application):
+class PythonApplication(Graphs.Application):
     """
     The main application singleton class.
 
     Functions:
         get_settings
     """
+    __gtype_name__ = "GraphsPythonApplication"
 
     def __init__(self, application_id, **kwargs):
         """Init the application."""
--- a/src/clipboard.py
+++ b/src/clipboard.py
@@ -7,8 +7,8 @@ from graphs import item, ui
 import numpy
 
 
-class BaseClipboard(GObject.Object, Graphs.Clipboard):
-    __gtype_name__ = "BaseClipboard"
+class Clipboard(GObject.Object, Graphs.ClipboardInterface):
+    __gtype_name__ = "GraphsClipboard"
 
     application = GObject.Property(type=Adw.Application)
     clipboard = GObject.Property(type=object)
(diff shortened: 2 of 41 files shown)


src/application.py
```diff
@@ -26,13 +26,14 @@
 ]
 
 
-class GraphsApplication(Graphs.Application):
+class PythonApplication(Graphs.Application):
```
Sjoerd1993, line 29: I am not 100% convinced of renaming this main class, but I have no super strong opinion either. The original class name is what is set by GNOME Builder when creating a new project, so I always assumed that is the most conventional way to do this, and it's close to what most projects I see seem to be doing for the main class name.

But it's not directly a blocker for me.

src/item_box.py
```diff
@@ -57,52 +60,42 @@ def add_actions(self):
...
+            "button { "
+            f"color: {item.get_color()}; "
+            f"opacity: {item.get_alpha()}; "
+            "}", -1,
+        )
 
-    def move_down(self, _action, _shortcut, item):
-        data = self.get_application().get_data()
-        before_index = data.index(item)
+    def move_up(self, _action, _shortcut):
```
Sjoerd1993, line 94: This was a refactor I had in my mind as well, given the strong resemblence between these two functions

src/clipboard.py
```diff
@@ -169,11 +168,10 @@ def perform_redo(self):
                 data.append(item.new_from_dict(change))
                 items_changed = True
             elif change_type == 2:
-                data.pop(change[1]["key"])
+                data.pop(change[1]["uuid"])
```
Sjoerd1993, line 171: I noticed, and fixed these also in the curve fitting PR

Sjoerd1993 (commented):
Mostly changes I agree with. I'm not 100% convinced for PythonApplication for the main application class, as it's super general (there's many different Python applications), and it doesn't seem to be really the general conventions either as far as I can tell. So was just wondering about the motivation for that specific change, before merging this to main.

src/application.py
cmkohnen: The reason here was, that the gtypename should resemble the classname with the namespace prefixed. The classname should'nt include the namespace "Graphs" itself. The problem here was that the typename "GraphsApplication" was taken by our own libgraphs (vala code). Since this was an actual class not an interface like e. g. Data I had to come up with something. I'm open for suggestions here. One idea I had was to revert to main.py as a single entry point and handle all the font registering, config migration etc. outside any class's scope. This way we do not need to inherit from our vala defined main class anymore and this problem would'nt persist.

Sjoerd1993 (approved):

PR #499, opened 2023-09-15
fix hide-unselected

cmkohnen:


Sjoerd1993 (approved):
I did not notice this was broken, thanks this works!

Issue #490, opened 2023-09-15
Add curve fitting confidence band


Sjoerd1993:
It would be nice to be able to show a confidence band in the curve fitting tool.

PR #501, opened 2023-09-15
add curve fitting statistics

### LINKED ISSUE #489: Add curve fitting statistics
The curve fitting tools works quite well, but some statistics would be nice. I'm thinking of adding:
- Confidence interval.
- Figure of merit, probably in the form of sum of R^2.

Sjoerd1993:
Adds some statistics to the curve fitting data, and shows bounds for the confidence interval (based on one standard deviation). Also rounds the given data to three significant digits. 

 For goodness of fit it shows an R2 value. Note that this assumes no intrinsic errors in the y-data that is given in, so it's a measure of how good the model fits to the data, not of the uncertainty of the data itself. Which I believe is why the measure is somewhat controversial, nevertheless it can be useful information imo.

Still need to add options to turn the bounds off, and show different kind of confidence intervals if reasonably possible, but thinking of doing that in a different PR. See picture:

image

Closes #489 and #490, adding more options is kinda separate but would be nice to have. These statistics are kinda the minimal things that I wanted to add at least before starting beta testing. (Again, for the beta period, I suggest to start at least after migrating to GNOME 45, as that will introduce some significant UI changes with the new AdwOverlaySplitView)


cmkohnen (approved):

Sjoerd1993:
There was an issue with the latest PR, mainly that the old bounds were not removed when redrawing new ones. The other issue is that the canvas limits would not update. The limits are bound to the figure settings. Hence those being used now. As the autoscale is basically overwritten in the canvas implementation, a logic similar to optimize_limits is being used here.

In the move to Vala, there were two things that I did not get working:
1 - The binding of the AdwEntryRow to the values in settings. For some reason, the `Tools.bind_settings_to_widgets` would correctly get the key, but did not retrieve any widgets. Hence assigning self.equation so I could use the python variant.
2 - The binding of the button to the Python call to add_fit. Hence the binding in the code here.

The latest commit fixes this. A bit more logic has been moved into Vala as well.

cmkohnen (changes requested):
Again, I feel like Figure Settings aren't needed here, we don't really need labels, scales are set to linear anyways and the limits can be set using just the canvas. I doesn't need to be tied to another class, as this only introduces overhead.

Sjoerd1993:
> Again, I feel like Figure Settings aren't needed here, we don't really need labels, scales are set to linear anyways and the limits can be set using just the canvas. I doesn't need to be tied to another class, as this only introduces overhead.

The problem is that the limits are not properly applied, but I think I found the culprit actually, which seems to lie in the autoscale setting. For some reason, even though the PolyCollection object is removed (the one that contains the confidence bounds), it still keeps the highest limits as reference. Turning off autoscale makes it work quite well without needing figure settings. Am working on a fixing that, will see if I have time for that today. I'll toggle a re-review when I've got a cleaner solution here. As I agree that we should avoid figure settings here, especially if it's just to set one single property (limits). At the time it seemed like it was the only way to set the limits though, as those were stuck (when autoscale was turned on).

cmkohnen (approved):

PR #503, opened 2023-09-16
Improve legend positions

Sjoerd1993:
In the current main branch, the legend is always chosen on the "top" axis, which is top_right. However, this means that it doesn't properly "hide" if a line is in front of the legend. Which is not ideal. This PR fixes this, so that the legend is put on the top axis that happens to have a line on it instead.

### DIFF
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -109,15 +109,14 @@ class Canvas(FigureCanvas, Graphs.CanvasInterface):
         positions = []
         used_axes = []
         for item_ in self.get_application().get_data():
-            positions.append(item_.get_xposition())
-            positions.append(item_.get_yposition())
-        if positions == [0, 0]:
+            positions.append([item_.get_xposition(), item_.get_yposition()])
+        if [0, 0] in positions:
             used_axes.append(self.axis)
-        if positions == [0, 1]:
+        if [0, 1] in positions:
             used_axes.append(self.right_axis)
-        if positions == [1, 0]:
+        if [1, 0] in positions:
             used_axes.append(self.top_left_axis)
-        if positions == [1, 1]:
+        if [1, 1] in positions:
             used_axes.append(self.top_right_axis)
         return used_axes[-1] if len(used_axes) != 0 else self.axis
 
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -105,6 +105,23 @@ class Canvas(FigureCanvas, Graphs.CanvasInterface):
         self._legend_position = misc.LEGEND_POSITIONS[0]
         self._handles = []
 
+    def get_top_axis(self):
+        positions = []
+        used_axes = []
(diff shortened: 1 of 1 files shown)


src/canvas.py
```diff
@@ -105,6 +105,23 @@ def __init__(self, application):
         self._legend_position = misc.LEGEND_POSITIONS[0]
         self._handles = []
 
+    def get_top_axis(self):
```
cmkohnen, line 108: I believe there is a logic problem with this function.

My guess is the intend is to loop over all items, figure out which axes are used and return the axes of the last item.

There are a few issues then:
using the whole application.get_data() ignores hide_uselected
this breaks when plotting more than one item
we are duplicating efforts already made in the item setter method.

If my assumption is true, I'd like to make a separate pr to integrate it into the item setter, as performance is critical here

Sjoerd1993: The intend was not to return the axis of the last item, but to return the top axis.

Basically, the axes have a hierarchy from bottom top top: axis, right_axis, top_left_axis to top_right_axis. The idea was to go through the axes in that order, and then if the axis is used, then it should be added to the list. It then returns the last axis in the list, which will always be the top one.

The problem is in the condition here. I wrote the loop here originally for the old style before we were using enums and then quickly adapted it, where I can see some issues. I'll go and edit it later this evening when I've got time. Basically the positions array should be four elements, (left, right, top, bottom), and then to get the list of used axis I can check if the corresponding directions are used to check if the axes should be considered to place the legend.

Sjoerd1993: This should work now

cmkohnen:
This works, see https://github.com/Sjoerd1993/Graphs/pull/505 for improvement suggestions

Sjoerd1993:
I doubt the original PR has a noticeable impact on the performance, as the biggest hindrance is creating a simple array the size of the amount of items each time the legend is refreshed, which realistically can be counted on one hand.

The reason I like your solution more however, is because it decouples the axis choice from the main application items, and couples it to the items on the canvas instead. This makes the canvas class more versatile. One real example for example is that someone has two items open, one on bottom-left position, and one on bottom-right position. As the top axes that is used, bottom-right will be chosen as axis of choice in both my logic, as well with this new logic.

Now if we open the curve fitting tool on the bottom-left data in my old PR. It checks again which axis is the top axis is used based on the items in the main application. So it decides bottom-right is the top axis. Since this is not actually used in the curve fitting canvas however, the legend will be placed straight on top of the lines. In this new logic however, it actually uses the items that are on the canvas. And thus correctly concludes that the correct canvas for the items is indeed bottom-left. I've done some  testing, and this is indeed a bug that existed in my old PR, and is solved with this latest logic. 

A bit of a long winded explanation of why I prefer this new one, but for my part this is good to go.

cmkohnen (approved):

PR #504, opened 2023-09-17
Refactor

cmkohnen:
- Make figure settings an attribute of data
- Integrate view clipboard and data clipboard into data
- Integrate optimize limits into data
- change project keys

Overall this gives several performance improvements as we already have data close in memory for example. 
Also makes the data object more versatile, as it can now truly be used without calling to graphs specific structures.


src/actions.py
```diff
@@ -76,37 +76,41 @@ def add_equation_action(_action, _target, self):
...
 def optimize_limits_action(_action, _target, self):
-    utilities.optimize_limits(self)
+    self.get_data().optimize_limits(self)
 
 
 def view_back_action(_action, _target, self):
-    if self.get_window().get_view_back_button().get_sensitive():
-        self.get_view_clipboard().undo()
+    data = self.get_data()
+    if data.props.view_back_possible:
```
Sjoerd1993, line 106: Just a suggestion, but perhaps `data.props.can_view_back` would make sense? It's a bit shorter, and has the same meaning in the end.

src/ui.py
```diff
@@ -33,14 +33,15 @@ def on_figure_style_change(_a, _b, self):
...
 def on_items_change(data, _ignored, self):
+    data = self.get_data()
     item_list = self.get_window().get_item_list()
     while item_list.get_last_child() is not None:
         item_list.remove(item_list.get_last_child())
 
-    for item in self.get_data():
-        item_list.append(ItemBox(self, item))
+    for count, item in enumerate(data):
+        item_list.append(ItemBox(self, item, count))
```
Sjoerd1993, line 42: The `count` variable sets the index for the new item to be added, right? Wouldn't it make sense to call it index then?

Sjoerd1993 (changes requested):
I'll try and look into this into a bit more detail tomorrow, when I have time but some quick things I noticed:

- Performing a redo doesn't change the limits back to the position it was at that stage.
- Optimize limits from the action menu is broken: `TypeError: Data.optimize_limits() takes 1 positional argument but 2 were given`

See video, showing some more bugs.

[Screencast from 2023-09-17 21-29-04.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/50d830d6-9611-4de8-84bc-3f42e2b14672)

cmkohnen:
Mentioned bugs should be fixed, some were actually introduced earlier, but just arose now

src/data.py
```diff
@@ -96,16 +101,16 @@ def get_items(self) -> list:
...
 
-    def append(self, item_):
+    def _add_item(self, item_):
         """Append items to self."""
         self._connect_to_item(item_)
         self._items[item_.get_uuid()] = item_
         self.notify("items")
 
-    def pop(self, key):
+    def _delete_item(self, key):
```
Sjoerd1993, line 113: I was actually considering renaming this mehtod either way, as a `pop` action in Python also returns the item that is deleted. This method does not, so it's a bit of misleading name to call this pop.

Sjoerd1993 (approved):
I think the migration from clipboard, figure settings and optimize limits make sense in the context of the new curve fitting tool. While working on that, I noticed there were multiple aspects that had a too strong coupling with the main application. 

This refactor makes it all a lot more versatile. Even though Data is becoming quite big as a class.

Issue #491, opened 2023-09-15
 Scroll- and pinch to zoom gestures in Matplotlib


Sjoerd1993:
See related issue in the GNOME Circle Feedback tracker:

` Canvas navigation feels a bit clunky, is scroll/pinch-to-zoom out of the question with the widgetry you're using for the graph?`

I've tested a bit with the scroll-event on Matplotlib. It doesn't seem to give the accurate event.xdata and event.ydata though, so scrolling into wherever the pointer is hovering probably needs some workaround.

Issue #385, opened 2023-08-07
GNOME Circle feedback tracker


Sjoerd1993:
We've received initial feedback on our [GNOME Circle application](https://gitlab.gnome.org/Teams/Circle/-/issues/185), resulting in a list of suggestions that we will adress. This issue is mean to track the progress on the given feedback so far:

Feedback round 1
---------------------
- [x]  Canvas navigation feels a bit clunky, is scroll/pinch-to-zoom out of the question with the widgetry you're using for the graph?

- [x] I wonder if the expanders in the tools section are needed, since most of the sections aren't that long. Perhaps it could be just three sections with a standard header?

- [x]  I'd move "Translate and Multiply" down to be the last section, since the other two are more similar

- [x]  These entry rows look a bit odd. I'd try regular entries, and using the same width for the buttons to make them all align horizontally.

![image](https://github.com/Sjoerd1993/Graphs/assets/68477016/278d5422-e3f2-4760-ad01-8d2c9a2d20ec)

- [x]  The app has lots of settings, it'd be good to review if all of them are necessary/being used. If you're not sure something's being used it's often better to just choose a reasonable default, because it makes it easier to find the important settings. For example, is it really necessary to have a setting for clipboard length? Could it just be unlimited? If not, are you aware of cases where someone would explicitly seek out the option?

- [x]  Many listboxes (especially in Preferences) have both a title and a subtitle, but the subtitle is just restating the title. In these cases it's better to omit the subtitile.

- [x]  Similarly, many listbox rows have both title and subtitle, with the subtitle being very similar to the title. I'd do the same thing here and drop the subtitle if it doesn't add much.

![image](https://github.com/Sjoerd1993/Graphs/assets/68477016/dad513d4-472e-420e-9cba-52385b15a879)

- [x]  If we keep these options (which I'm not totally sure are necessary) I'd use the same 3 column layout from the "Add Equation" dialog - removed

![image](https://github.com/Sjoerd1993/Graphs/assets/68477016/6e64001d-67d6-4377-9cf4-3e6b319ce288)

Feedback round 2
-----------------------

- [x] This menu is a bit wide, which looks awkward. Perhaps the equation string could be shortened to "Add Equation..."? I'd also drop the divider between the two options, it seems unnecessary with just 2 items.
![image](https://github.com/Sjoerd1993/Graphs/assets/68477016/2441f77b-010f-4312-ab54-db5f8050329a)

[...]

PR #506, opened 2023-09-18
Implement ctrl+scroll and pinch to zoom gestures

Sjoerd1993:
See title. 

Still need to change the zooming logic so that it works on non-linear axes. Also, the way to detect if a ctrl is pressed is not my ideal, but for some reason the key_press event in the canvas is not registered. My guess is that this is because the canvas is not in focus, which is by design as we need to catch the shortkeys in the main application.

[Screencast from 2023-09-18 14-39-55.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/58bf3b36-5f41-4b9f-b512-1474a63d6f54)

Closes issue #491, also closes the last checkbox in #385 but I won't close that one until we've been accepted into GNOME circle, as new feedback will probably arise.

### DIFF
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -21,6 +21,8 @@ from matplotlib.backend_bases import NavigationToolbar2
 from matplotlib.backends.backend_gtk4cairo import FigureCanvas
 from matplotlib.widgets import SpanSelector
 
+import numpy
+
 
 class Canvas(FigureCanvas, Graphs.CanvasInterface):
     """
@@ -104,20 +106,13 @@ class Canvas(FigureCanvas, Graphs.CanvasInterface):
         self._legend = True
         self._legend_position = misc.LEGEND_POSITIONS[0]
         self._handles = []
-        self.mpl_connect("scroll_event", self.on_zoom_event)
+        self.mpl_connect("scroll_event", self.on_scroll_event)
         self.mpl_connect("motion_notify_event", self.get_coord_fraction)
         self.xfrac = 0
         self.yfrac = 0
         zoom_gesture = Gtk.GestureZoom.new()
         zoom_gesture.connect("scale-changed", self.on_zoom_gesture)
         self.add_controller(zoom_gesture)
-
-    def on_zoom_gesture(self, _gesture, scale):
-        scale = 1 + 0.15 * (scale - 1)
-        if scale > 5 or scale < 0.2:
-            # Don't scale if ridiculous values are registered
-            scale = 1
-        self.on_zoom_event(None, scale, touch=True)
 
     def get_application(self):
         """Get application property."""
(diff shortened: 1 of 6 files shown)


Sjoerd1993:
This works now, I'm open to changing ctrl+scroll to just scrolling without having to hold control by the way.

cmkohnen:
As per [GNOME HIG](https://developer.gnome.org/hig/guidelines/pointer-touch.html#scrolling-panning-zooming) Ctrl + Scroll to zoom is intended for views, that can scroll vertically (e. g. viewing a pdf document) and for views that won't scroll vertically (like in our case), just using the scroll wheel should suffice.

src/canvas.py
```diff
@@ -104,11 +106,99 @@ def __init__(self, application):
...
+        scale = 1.6
+        if event.button == "up":
+            scale = 1 / scale
+        if self.get_application().get_ctrl() is True:
+            self.on_zoom_event(scale)
+
+    def on_zoom_event(self, scaling=1.6):
+        for ax in [self.axis]:
+            zoom_factor = scaling
+            scale_mapping = {
```
cmkohnen, line 151: I'd use scales.to_int() here.

src/canvas.py
```diff
@@ -104,11 +106,99 @@ def __init__(self, application):
...
+
+    def on_scroll_event(self, event):
+        scale = 1.6
+        if event.button == "up":
+            scale = 1 / scale
+        if self.get_application().get_ctrl() is True:
+            self.on_zoom_event(scale)
+
+    def on_zoom_event(self, scaling=1.6):
+        for ax in [self.axis]:
```
cmkohnen, line 149: shouldn't this be applied to all axes?

Sjoerd1993: Thanks, yes I put this to only the first axis for testing purposes. Forgot to bring it back

src/canvas.py
Sjoerd1993: Thanks, just didn't think about this function, that's obviously a much better choice

src/canvas.py
```diff
@@ -104,11 +106,99 @@ def __init__(self, application):
...
+        self.add_controller(zoom_gesture)
 
     def get_application(self):
         """Get application property."""
         return self.props.application
 
+    def get_coord_fraction(self, event):
+        xlim = self.top_right_axis.get_xlim()
+        ylim = self.top_right_axis.get_ylim()
+        figure_settings = \
```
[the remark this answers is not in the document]
Sjoerd1993: Also here, yes that's cleaner obviously.

Sjoerd1993:
I also added ctrl+ and ctrl- for zoom in and zoom out respectively. The ctrl+ and ctrl- action zoom into the middle, not where the mouse pointer is pointing. This is also what Inkscape does for instance. But I'm open to let it zoom to where the mouse pointer is pointing, and go for the middle if the mouse pointer is not on the canvas.

The used triggers to match to the different plus/minus keys were taken from [Nautilus](https://github.com/GNOME/nautilus/blob/ca829d48a01df9c9db9e5f68cccb2244874917b5/src/nautilus-files-view.c#L9949), hence the "ZoomIn" and "ZoomOut" accelerators in there as well.

cmkohnen (approved):
Both scolling and Pinch to zoom work in my testing. However two finger scrolling is quite aggressive. Don't know if that can be fixed however.

The keyboard shortcuts behaviour is fine as is, using the mouse position there is unintuitive

Sjoerd1993:
> Both scolling and Pinch to zoom work in my testing. However two finger scrolling is quite aggressive. Don't know if that can be fixed however.

Yeah, it's why I reduced the zoom factor for the scroll event quite a bit compared to the original PR. I tuned them down a bit more now as well.

PR #508, opened 2023-09-18
Migrate to GNOME 45

Sjoerd1993:
There's still quite some work here to be done, and I need to fix the headerbar as that is somewhat broken at the moment. But I thought to put this up, also to show the work so far and to allow for feedback and suggestions. Either way, I won't merge this before the stable Gnome45 runtime is out. The builder also fails (not unexpectedly), because it doesn't have access to the beta repos by default.

See screenshot to see what it looks like with the new SplitOverlayWidget:

image

I probably want to move the redo/undo buttons to the main content part actually, but am not 100% sure yet. Also, the title should probably be in the middle of the content (canvas), and not in the middle of the application itself.

TODO:

- [x] Fix main window with headerbar etc..
- [x] Double check all dependencies
- [x] Update GNOME45beta runtime to stable

Apart from AdwFlap, I don't think there's any widgets that are deprecated. We can discuss whether we want to use SpinRows or the new SpinButtons in places like the amount of rows to skip and the column indices. But as they're not deprecated, I think we can keep that in a separate PR.


cmkohnen:
I believe the styles window uses AdwNavigationView(?) Which is deprecated as well

Sjoerd1993:
> I believe the styles window uses AdwNavigationView(?) Which is deprecated as well

Thank you, you were correct. AdwNavigationView is the widget that replaces the old one (AdwLeaflet). I've replaced that now also in this PR.  I am considering using an AdwNavigationSplitView for the style menu, that would look nice and make it easier to quickly navigate styles and change settings there. But that's outside of the scope of this PR at least, and not even necessarily something that needs to be done before next release, it would probably require a slightly different widget for the style rows as well. The current implementation is just a straight up migration from AdwLeaflet to AdwNavigationView.

I also moved the primary menu to the sidebar, as that's what the GNOME HIG say nowadays. Also, right now I moved the headerbar buttons to the content menu, also in the collapsed state. But that's not a decision that is set in stone, it just felt natural to keep them in the same content section now there's such a clear visual distinction between the sidebar and the content part:
[Screencast from 2023-09-19 10-39-17.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/689ff650-6211-456a-a53f-5028a35251f9)

cmkohnen:
I'd like to use the new ListBox rows introduced in Libadwaita 1.4 here as well (Switch and Spin rows)

Sjoerd1993:
Based on feedback from the GNOME team on the matrix, the title has been moved to the sidebar, and I dropped the card styling in the preference groups widgets in the sidebar:

image

Sjoerd1993:
> We should at least indeed use the Switch rows, they simplify things a bit. The Spin Rows look nice, but it seems like they don't offer the option to enter values from keyboard by default. Which is fine for setting the column indices for instance, but I am less sure about Skip Rows and DPI in the Figure Export settings. Not sure how it looks when we combine the different widgets though.
> 
> But maybe I'm wrong on this, I haven't seen the Spin rows in action tbh

So I was wrong on this, the new SpinRow actually does allow for keyboard entries, making it a viable choice imo for values that one may want to set with the keyboard. I've adopted the new widgets, and changed the breakpoints a bit to allow the item list to shrink a bit on smaller window sizes. This makes it possible to run Graphs in even smaller windows than before.

image

With the new widgets, I've removed the automatic bindings to the GtkSpinButtons that we were using before. I kept the Switch ones, as I can see a potential future use-case for those. But I'm open to keeping the bindings to the SpinButtons in, as well as taking the bindings to the switches out until we actually use them.

Sjoerd1993:
This is now done and seems to be working properly.

data/ui/style_window.blp
```diff
@@ -15,557 +15,522 @@ template $GraphsStylesWindow : Adw.Window {
...
-          icon-name: "list-add-symbolic";
-          tooltip-text: _("Add new style");
-          clicked => $add_style();
-        }
-        [end]
-        Button {
-          icon-name: "history-undo-symbolic";
-          tooltip-text: _("Reset to default styles");
-          clicked => $reset_styles();
+  Adw.NavigationView leaflet {
```
cmkohnen, line 18: Probably better to rename this

src/actions.py
```diff
@@ -43,8 +43,8 @@ def perform_operation(_action, target, self):
 
 
 def toggle_sidebar(_action, _shortcut, self):
-    flap = self.get_window().get_sidebar_flap()
-    flap.set_reveal_flap(not flap.get_reveal_flap())
+    sidebar = self.get_window().get_split_view()
```
cmkohnen, line 46: as this contains both sidebar and content I'd advocate for 'split_view'

cmkohnen (approved):
This looks great, although I kind of liked the cards in the navigation, see some naming suggestions, which aren't really blocking

PR #510, opened 2023-09-20
small improvements

cmkohnen:
- Make data.items notify explicitely. This was already kind of the case before, now it is just a bit more verbose.
- Only update figure_settings on project load, if the value is different
- Use the can_view_{back/forward} property to check if respective actions are possible.
- exit history functions earlier if executing isn't possible


Sjoerd1993 (approved):
The GNOME pages are down, which explains the failing builds, as Blueprint is hosted on the GNOME gitlab.

Issue #496, opened 2023-09-15
Migrate to GNOME 45


Sjoerd1993:
Once GNOME 45 officially releases, it's time to start the migration to GNOME 45. Already now, the runtime is available in the beta channel. For this, we need to update our dependencies to Python 3.11, and upgrade the deprecated widgets to their successors.

PR #512, opened 2023-09-21
Use stack switcher

Sjoerd1993:
Based on #496, didn't want to put it in that PR though as to narrow down the scope of that PR a bit.

I've been thinking of finding a way to organize the action buttons a bit. A long list of different actions is not ideal, especially if some of the tools are not visible by default when opening the application. So I was thinking of using stack pages here.

After asking for some advice in the GNOME design circle, the recommendation is to use inline stack switcher. While this widget is not ready yet in the current libadwaita, it will be added in the future. An implementation is already available from Elastic, which is where I have taken this from. 

See video, this organizes the tools in distinct categories. Exposing fewer choices at once to the user.

[Screencast from 2023-09-21 12-27-31.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/bdf179e1-5fe6-4387-9741-1651266f1678)


Sjoerd1993:
One alternative is to turn off hexpand for the stack switcher, but I think it looks better covering the entire span:
image

A thing I am really unsure about is the slide transition. I'm also happy dropping the transition between the different pages, as it looks a bit more "mature" maybe without the transition.

cmkohnen:
Visually, I like this change. I'd keep both the slide animation as well as the hexpand. This really helps in removing visual clutter on first glance

Sjoerd1993:
Yeah, I'll keep the hexpand on and animations turned on for now. We can always turn them off if feedback suggests that it's better without them.

Sjoerd1993:
The failing build is because the Snap package can't find Libadwaita 1.4 by the way. I've posted a message in the Snap forums, and perhaps @soumyaDghosh knows how to resolve this. Worst case, we'll have to wait for the Snap package to update until the GNOME45 sdk releases. Which is probably not the biggest problem in the world, as it probably won't delay the stable Snap release, just the development builds.

src/inline-stack-switcher.vala
```diff
@@ -0,0 +1,234 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+public class Graphs.InlineStackSwitcher : Gtk.Widget {
```
cmkohnen, line 3: I'm not sure if it is better to define it this way or put it in namespace Graphs here, but I don't have a strong preference

cmkohnen (approved):

src/inline-stack-switcher.vala
Sjoerd1993: I basically just adopted the Widget from [Elastic](https://gitlab.gnome.org/World/elastic/-/blob/main/src/inline-stack-switcher.vala?ref_type=heads), hence the way this was defined. But this is a fair point, I prefer put it in the namespace as that's more consistent with the rest of the code.

soumyaDghosh:
> The failing build is because the Snap package can't find Libadwaita 1.4 by the way, which makes sense because it's a bit behind in the used GNOME sdk. I've posted a message in the Snap forums, and perhaps @soumyaDghosh knows how to resolve this. Worst case, we'll have to wait for the Snap package to update until the GNOME45 sdk releases. Which is probably not the biggest problem in the world, as it probably won't delay the stable Snap release, just the development builds.

Kindly give me th forum post link and I also gonna work to push the gnome-sdk ASAP. Thanks for your kind words, BTW.

Sjoerd1993:
> Kindly give me th forum post link and I also gonna work to push the gnome-sdk ASAP. Thanks for your kind words, BTW.

Thanks for the quick reply, here's the forum link: https://forum.snapcraft.io/t/getting-gnome45-in-a-snap/36950
There's still some polish that needs to be applied for the next release, as well as some testing either way, so in the end this probably won't hinder us too much. We'll probably be able to push the next stable release for Flatpak and Snapcraft simultaneously.

The Snap is at 330 active installs by the way, which shows that the Snap version does reach quite some people :)

PR #515, opened 2023-09-24
add yaru style

cmkohnen:
like the title says. Also add both yaru and seaborn as supported system styles.

### DIFF
--- a/data/se.sjoerd.Graphs.gresource.xml
+++ b/data/se.sjoerd.Graphs.gresource.xml
@@ -28,6 +28,8 @@
     <file compressed="True">styles/solarized-light.mplstyle</file>
     <file compressed="True">styles/tableau-colorblind10.mplstyle</file>
     <file compressed="True">styles/thesis.mplstyle</file>
+    <file compressed="True">styles/yaru.mplstyle</file>
+    <file compressed="True">styles/yaru-dark.mplstyle</file>
     <file preprocess="xml-stripblanks">ui/add_equation_window.ui</file>
     <file preprocess="xml-stripblanks">ui/add_style.ui</file>
     <file preprocess="xml-stripblanks">ui/curve_fitting.ui</file>
--- a/data/se.sjoerd.Graphs.gschema.xml
+++ b/data/se.sjoerd.Graphs.gschema.xml
@@ -2,6 +2,8 @@
 <schemalist gettext-domain="graphs">
   <enum id="se.sjoerd.Graphs.system-styles">
     <value nick="adwaita" value="0"/>
+    <value nick="seaborn" value="1"/>
+    <value nick="yaru" value="2"/>
   </enum>
 
   <enum id="se.sjoerd.Graphs.general.center-values">
--- a/data/styles/yaru-dark.mplstyle
+++ b/data/styles/yaru-dark.mplstyle
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
-# adwaita-dark
+# yaru-dark
 
 # font
 font.family: sans-serif
--- a/data/styles/yaru-dark.mplstyle
+++ b/data/styles/yaru-dark.mplstyle
@@ -1,0 +1,108 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+# adwaita-dark
+
+# font
+font.family: sans-serif
+font.sans-serif: Cantarell
+font.style: normal
+font.weight: normal
+axes.titleweight: normal
+axes.labelweight: normal
+figure.titleweight: normal
+figure.labelweight: normal
+font.size: 12
+axes.labelsize: 12
+xtick.labelsize: 12
+ytick.labelsize: 12
+axes.titlesize: 12
+legend.fontsize: 12
+figure.titlesize: 12
+figure.labelsize: 12
+
+# lines
+lines.linewidth: 3
+lines.linestyle: solid
+lines.marker: none
+lines.markersize: 7
+lines.markerfacecolor: auto
+lines.markeredgecolor: auto
+lines.markeredgewidth: 0
+lines.dash_joinstyle: round
+lines.dash_capstyle: butt
+lines.solid_joinstyle: round
+lines.solid_capstyle: round
+lines.antialiased: True
+lines.dashed_pattern: 6, 6
+lines.dashdot_pattern: 3, 5, 1, 5
+lines.dotted_pattern: 1, 3
+lines.scale_dashes: False
+markers.fillstyle: full
+
+# ticks
+xtick.direction: in
+xtick.minor.visible: True
+xtick.major.width: 1
+xtick.minor.width: 0.5
+xtick.major.size: 4
+xtick.minor.size: 2
+xtick.alignment: center
+
+ytick.direction: in
+ytick.minor.visible: True
+ytick.major.width: 1
+ytick.minor.width: 0.5
+ytick.major.size: 4
+ytick.minor.size: 2
+ytick.alignment: center
+
+xtick.bottom: True
+ytick.left: True
+xtick.top: False
+ytick.right: False
+
+# grid
+axes.grid: True
+axes.grid.which: major
+axes.grid.axis: both
+grid.linewidth: 0.6
+grid.alpha: 1.0
+grid.linestyle: -
+
+# padding
+xtick.major.pad: 7
+xtick.minor.pad: 7
+ytick.major.pad: 7
+ytick.minor.pad: 7
+axes.labelpad: 5
+axes.titlepad: 15
+
+# colors
+text.color: FFFFFF
+axes.labelcolor: FFFFFF
+xtick.labelcolor: FFFFFF
+ytick.labelcolor: FFFFFF
+xtick.color: FFFFFF
+ytick.color: FFFFFF
+axes.edgecolor: FFFFFF
+grid.color: 878787
+axes.facecolor: 242424
+figure.facecolor: 2C2C2C
+figure.edgecolor: 2C2C2C
+
+axes.prop_cycle: cycler('color', ['335280', '0E8420', 'F99B11', 'E95420', 'A91224', '2C001E', '72441D', '878787'])
+patch.facecolor: 1A5FB4
+
+# other
+axes.linewidth: 0.8
+
+mathtext.default: regular
+axes.axisbelow: True
+legend.fancybox: True
+legend.frameon: True
+
+# custom
+patch.linewidth: 1
+figure.figsize: 8.0, 15.5
+image.cmap: Greys
+legend.numpoints: 1
+legend.scatterpoints: 1
--- a/data/styles/yaru.mplstyle
+++ b/data/styles/yaru.mplstyle
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
-# adwaita
+# yaru
 
 # font
 font.family: sans-serif
--- a/data/styles/yaru.mplstyle
+++ b/data/styles/yaru.mplstyle
@@ -1,0 +1,108 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+# adwaita
+
+# font
+font.family: sans-serif
+font.sans-serif: Cantarell
+font.style: normal
+font.weight: normal
+axes.titleweight: normal
+axes.labelweight: normal
+figure.titleweight: normal
+figure.labelweight: normal
+font.size: 12
+axes.labelsize: 12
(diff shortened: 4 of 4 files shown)


Sjoerd1993 (commented):
I have tested the themes (on GNOME Builder, Snap still doesn't support libadwaita 1.4 yet, [it's being worked on](https://github.com/ubuntu/gnome-sdk/issues/170)), and they seem pretty good. Just a few questions:

1 - It seems the default line colours are different from the adwaita theme. Is there a specific reason for this? Where do these specific colours come from? They seem to be slightly different than the Ubuntu accent colours as well.

2 - Any specific reason to add Seaborn to the list of system themes? Do we plan to expose the system theme to the users? If no, in what situation would we ever need seaborn as default? In principle it would probably fit in quite well with KDE, but Graphs in KDE still uses the same colour scheme as in GNOME.

If we do expose the setting for a system theme, then we should think a bit about how we'd do that, as to not have it too messy. My first gut feeling would be to simply be able to switch between Yaru and Adwaita. As the idea is that it should be the style that fits in with the system theme.

In principle I'd say that could be a dropdown from the regular preferences menu. By default, the Snap build should default to Yaru (at least on Ubuntu). The Flatpak behaves nicely, even on Ubuntu, so that can remain Adwaita regardless. This can all be detected well with the `platform` module, which is bundled with basic Python.

3 - Just a general comment, outside of the scope of this PR. I noticed this style did not automatically get added into my list of styles, and I had to trigger a reset to defaults to get it in there. People should be able to remove themes from that list, so it makes sense that these are not automatically loaded. But in principle it would be nice if the Yaru themes get added for people upgrading from 1.6 to 1.7. (In the end it's not vital, as the system themes should be stored independently anyway like Adwaita is)

Sjoerd1993:
On a related note, I found that if you set the "GTK Legacy Theme" to "Adwaita" in GNOME-Tweaks, then the colors are exactly as they should be. Similarly changing the icon style there from Yaru to Adwaita fixes the icons as well. 

The icons is a thing we can discuss. One can argue that it's preferable to follow the Ubuntu theme there to get consistency over the desktop (although I think they look significantly worse than the stock GNOME icons), but perhaps it would be possible to simply override the GTK stylesheet on the Snap package. Just need to figure out how to do that. In that case, we can still add a Yaru theme anyway as custom theme by the way. But then it wouldn't be necessary to have a different look for the Ubuntu platform.

cmkohnen:
Maybe it helps to give my thought process here.
The system-style setting was not intended to be user-facing. It should only be used for packaging. The only requirement was, that a style would have a dark variant else we'd have to change the code here. Adding seaborn was just to have it as an option (for packaging or users with dconf).

Regarding Ubuntu, my prefferred solution would be to enforce the libadwaita theme if possible, as that would also eliminate inconsistency in icons. I'd still want to include yaru in this case however just as an option.

Perhaps for migration purposes, we can split between user styles and prepackeged (system) styles. The user would still be able to select any desired, but can't modify them and has to make a copy first. This way we can modify the stylesheets without deleting user data.

Regarding the line colors, I used the yaru Gradience theme to extract them. But I'm fine changing them.

Sjoerd1993:
> Regarding Ubuntu, my prefferred solution would be to enforce the libadwaita theme if possible, as that would also eliminate inconsistency in icons. I'd still want to include yaru in this case however just as an option.

Yes, I honestly kinda felt the same. It's really an annoyance as well that we spend a lot of time and effort to get the theming right and balanced. And then Canonical just sweeps in and applies a different stylesheet on this. The fact that the icons in Yaru look significantly worse imo is not really helping either. I'm not really someone who cares a lot about marketing and such, but I do feel it kinda hurts the branding of Graphs if a significant part of the userbase (330 active installs and climbing) get a sub-par theme. But again, those icons are consistent with the rest of Ubuntu, so maybe it's not that bad. But my feeling also goes with overridign them. Keeping the yaru theme in as stylesheet for the Graph is probably a good idea either way.

I think I may have found how to set the theme as well, but unfortunately the Snap doesn't build properly locally. And the Snapcraft store is coupled to the main branch, which fails due to libadwaita 1.4 not being avialable in their sdk yet (it seems to have been merged yesterday on Github, so it will probably arrive soon).

But I think these can be set in the slots category like so:

```
slots:
  adwaita:
    interface: content
    content: desktop-theme
    read:
      - /usr/share/themes/Adwaita
  adwaita-icon:
    interface: content
    content: desktop-theme-icon
    read:
      - /usr/share/icons/Adwaita
  graphs:
    interface: dbus
    bus: session
    name: se.sjoerd.Graphs
```

At least, that's what I get from looking at this: https://github.com/ubuntu/snap-theme-examples/blob/master/themes-snap/adwaita-themes/snap/snapcraft.yaml Perhaps it should also be added as part (like in the link), but again not 100% sure. The thing is that building Snaps doesn't work in my VM, and the actual Snapcraft store doesn't build at the moment. So I will try and test a bit later once libadwaita 1.4 is actually in. Then I can just set it in the yaml, and release the version to the edge channel and see if that works. (That is a nice thing actually with Snapcraft, there's many different stages ranging from Stable, to Edge. Where the last one is basically just a daily build directly taken from the Github repo. Unfortunately they're all pinned to the main branch)

Regarding this PR in particular, I'll approve it for merging. Having some options if people want to build it themselves is fine. And if someone wants to package it themselves with a full Ubuntu look that can be done that way, but I am feeling that the "official builds" should resemble the actual intended product (meaning the overriden themes). 

Especially if we manage to override the theming for the Snap package, I think it's fine with the slightly different color scheme. Extracting them from Gradiance's Yaru theme there makes sense.

soumyaDghosh:
@Sjoerd1993 The example you took is not anymore used. All the snaps now gets themes from this snap package. Here all the themes and icons are available. 
https://github.com/ubuntu/gtk-common-themes/blob/main/snap/snapcraft.yaml

and probably, also due to this stylesheet. Though, this stylesheet is mainly for the yaru accent colors.

https://github.com/ubuntu/gnome-sdk/blob/gnome-42-2204-sdk/patches/style-manager-Support-Yaru-accent-colors.patch

Now, regarding the color mismatch, I am not an expert but what you can do is, try to check for `SNAP` in environment variable. Then if it's there, use static colors, other than depending on System. (Not sure if this can be done or not). But, as said

> I'm not really someone who cares a lot about marketing and such, but I do feel it kinda hurts the branding of Graphs if a significant part of the userbase (330 active installs and climbing) get a sub-par theme.

I think it only happens in Ubuntu gnome, not in other flavours of Ubuntu.

cmkohnen:
> Now, regarding the color mismatch, I am not an expert but what you can do is, try to check for `SNAP` in environment variable. Then if it's there, use static colors, other than depending on System. (Not sure if this can be done or not). But, as said

The problem is, we *use* static colors. But the static colors, that match with the actual libadwaita theme we intend to use (by using GTK4 + libadwaita). Using the stylesheet indroduced in this pr on snap would mitigate that style mismatch, but not the icon mismatch or other things we did not notice yet.

> I think it only happens in Ubuntu gnome, not in other flavours of Ubuntu.

Considering Ubuntu GNOME(ish) is the main distribution, used by probably >75% of Ubuntu users, this would still be a breaking issue. Especially, speaking from experience here, if you have access to linux computers from universities, etc. which all probably use Ubuntu GNOME

Sjoerd1993:
Thanks for the info, I saw that audacity seems to set the theme themselves and indeed they use the gtk-common-themes. https://github.com/diddlesnaps/audacity/blob/master/snap/snapcraft.yaml

Indeed it seems that that particular patch is the culprit here, it does exactly what we want to prevent here. Worst case scenario I can probably fork the sdk and remove the patch, but keep it up-to-date with upstream otherwise. But I'll experiment first by trying to apply the themes themselves. (That would be latter half of this week at it's earliest as I'm kinda busy for this week).

The icon mismatch is simply due to Ubuntu setting the icon theme different by default, and should be able to be mitigated by setting it in the Snap ourselves (to be equal to the adwaita ones, instead of following the system), which seems to be supported relatively well in Snapcraft as far as I can see. 

> Considering Ubuntu GNOME(ish) is the main distribution, used by probably >75% of Ubuntu users, this would still be a breaking issue. Especially, speaking from experience here, if you have access to linux computers from universities, etc. which all probably use Ubuntu GNOME

Exactly this. Furthermore, overriding the system theme in a distribution that does respect the Adwaita theming won't really have any effect anyway. So I really don't mind if the same theme is overriden in e.g. the Fedora snap, or in e.g. Kubuntu. (Actually, Fedora KDE actually also had weird icons, but applications looking somewhat weird outside of GNOME is not unexpected)
If Graphs where to be installed on a university environment, you can bet it will be Ubuntu (that's all you see on user-facing university computers that run Linux), and pretty much all of them will use the Snap packages (organisations are usually very skeptical of adding another repository (Flathub) to the base system). Again, which is why I'm happy the Snap exists, it really clearly reaches a significant amount of people, so thanks again for the support here!. 

To give some insight on what distro is being used in the Snap package, at least 94% of all installs are on Ubuntu atm. Which makes sense, since most other distro's prefer Flatpaks (afaik, it's only Canonical that pushes snap on the desktop). Given the lack of Ubuntu spins in the list, I assume that the spins are included in there. But the vast majority will be Ubuntu's version of GNOME. There will likely be some Kubuntu users on there as well, but again, I don't mind setting the theme in those distro's either.

image

soumyaDghosh:
@Sjoerd1993 can you share me with all the changes you want in the yaml file with me? Regarding icons, well, please keep it... I mean that mismatch will really hurt the look of this desktop. And if the patch is the main culprit, then I guess I can do something else. Give me some time. Currently very busy with my exams.

Sjoerd1993:
> can you share me with all the changes you want in the yaml file with me?

I don't have any specific line-by-line suggestions at the moment, as I haven't had the time to experiment. But specifically the stylesheet should be overriden to follow the regular libadwaita styling, which we are using for the plots. 

Regarding the icons, to be clear I personally think it's a distinctly bad move from Canonical to override the icons used in applications. It creates mismatch with screenshots, with available documentation, it can confuse metaphors, and it just creates unpredictable stylings that are impossible to test for. 

Having said that, this is not a packaging issue but rather a distro issue. For instance, the same issue exists in KDE where different icons are being used. In fact, I just tried the Flatpak on Ubuntu, and also the Flatpak version of Graphs uses the Yaru icons in Ubuntu. So I can live with keeping these icons as-is. We're after all not overriding them in the Flatpak version either. See screenshot for the Flatpak version in Ubuntu:

image

Basically, the main point is that the Snap version and the Flatpak version should provide the same user interface and experience. Overriding the icons would not be necessary for that. The thing right now is that the Snap version uses a different stylesheet for the coloring, which is a problem as that's not what we are testing for. So if we could override the stylesheet to be consistent with regular libadwaita (so without the Yaru coloring), that would be great. Then I can personally live with keeping the icons as-is, as I have made my personal feelings about that clear as well by [signing the letter](https://stopthemingmy.app/) on this.

Just to be clear, don't take my ranting here as a personal attack, or an annoyance towards you personally or the snapcraft community at large. I'm really grateful that you are helping out with the snap packaging! That's really appreciated. Like I said, it really helps bringing Graphs to a rather large community, where a significant portion of the user-base will not have Flathub enabled. 

**tl;dr**: The specific thing that needs to be changed is the stylesheet itself. Which seems to be due to said patch. The icons being mismatched from the intended version lies with Ubuntu, not with the Snap version. So I am fine with keeping that, even though I think it's unfortunate.

soumyaDghosh:
@Sjoerd1993 can you hardcode the library path of libadwaita if snap? Then I'll just build it from source for you.

Sjoerd1993:
I haven't looked into too much detail, but in principle it would probably be possible to add a custom build parameter in meson for the snap package, similar to how we now set the debug buildtype for the development versions. 

Then _I think_ we should be able to get libadwaita as subproject using the instructions [here](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/build-howto.html). I am not completely sure if/how this clashes with the libadwaita dependency that is installed from the Snap sdk though. It may complain that the dependency for libadwaita is already defined.

But again, I haven't looked into it in too much detail. But it should definitely be possible to set a slightly different behaviour in the meson build system for the snap package at least.

Sjoerd1993:
For the time being, the easiest thing probably is to just follow the original intention of this PR and set the system style on run-time when detecting Ubuntu and Snap, which we can do very easily using the `platform` module which gets bundled with Python. 

If there's a clean way to just set the default libadwaita theme in the snap, then that would be preferred imo. But setting the Yaru theme as system default in Ubuntu will at least solve the ugly clash we get in dark-mode right now.

PR #519, opened 2023-09-27
various changes for the latest release

soumyaDghosh:
image

The current state of the snap. Didn't get the option to change styles.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -17,16 +17,60 @@ slots:
     name: se.sjoerd.Graphs
 
 parts:
+  numpy:
+    source: https://github.com/numpy/numpy.git
+    source-tag: 'v1.25.2'
+    plugin: python
+    build-environment:
+      - PATH: ${CRAFT_PART_INSTALL}/bin:${PATH}
+      - PYTHONPATH: ""
+    python-packages:
+      - scipy==1.11.2
+      - python-dateutil==2.8.2
+      - pyparsing==3.0.9
+      - Pillow==10.0.0
+      - packaging==23.1
+      - contourpy==1.1.0
+      - kiwisolver==1.4.5
+      - fonttools==4.41.1
+      - cycler==0.11.0
+      - matplotlib==3.8.0
+      - numexpr==2.8.6
+      - mpmath==1.3.0
+      - sympy==1.12
+    prime:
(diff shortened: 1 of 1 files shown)


Sjoerd1993:
Thanks, the reason the top bar has a different canvas is likely because this still uses the Yaru styling. Nevertheless, this does already look a lot better, possibily just already with the new GNOME45 stylings.
As said, it will probably be easiest to set the system theme of Graphs to Yaru for Ubuntu, which we'll probably just implement later in the Python code, so that won't need to be done in the Snap yaml then. 

I hope to have some time this evening to look at this, in more detail, but am not 100% sure. Otherwise @cmkohnen may manage before me.

Sjoerd1993:
As a reference, here's how it looks like in Vanilla GNOME with Fedora:

image

I see the Snap here also uses the Ubuntu fonts instead of Cantarell, but that's likely also distro-specific and therefore still the case for the Flatpak on Ubuntu as well. But at least the color mismatch on the content-pane can be addressed with a system theme setting.

soumyaDghosh:
@Sjoerd1993 as a regular user of Ubuntu, if my opinion matters, I'll say that your app already looks good if not perfect in this desktop. In the end, only work matters. Looks matters, but upto some extent.

Sjoerd1993:
Of course any input is valuable, and you likely have a point as Ubuntu users are used to the fonts and icons anyway, as that's what the rest of the system follows.

Still think we should patch the mismatch in the color of the canvas for Ubuntu, so the header bar integrates nicely like the other GNOME 45 applications. But again, that's a toggle we'll set in the Python code in that case.

soumyaDghosh:
Sure!! Thanks for all the efforts given here. Thanks to @cmkohnen too.

cmkohnen (commented):
Just a question: Is there a reason numpy is compiled from source instead of using pip like all other dependencies do?

soumyaDghosh:
> Just a question: Is there a reason numpy is compiled from source instead of using pip like all other dependencies do?

Well, it's just a hack, to pull those via pip, but when the number gets become, this option seems a more viable and managable. I chose `numpy` randomly. As if I use python plugin I can add any amount of `python-packages` there.

cmkohnen:
Okay, just wondering

cmkohnen (approved):

PR #520, opened 2023-09-29
Set Yaru theme for snap package

Sjoerd1993:
Sets the Yaru system theme for the Snap package. Feel free with suggestions for other implementations. The logic here is that both the Snap and the Flatpak set some fingerprints in the environment variables (most noticable SNAP for the Snap package, the Flatpak can be identified  by the presence of FLATPAK_ID). 

Then for the Snap package, where the system theme is applied instead of libadwaita it checks what the current Gtk theme is. If this is set to anything that starts with Yaru (so also the Yaru variants), then Yaru is set as the preferred system theme. Otherwise it revers to adwaita, our preferred theme we target for. The reasoning for that second line where it reverts to adwaita, is that the settings are saved. So if the yaru theme is set in the Snap, and you then switch themes back to Adwaita, it would still keep the Yaru theming without that else statement.

Open to suggestions for different kind of implementations for this.

### DIFF
--- a/data/se.sjoerd.Graphs.gschema.xml
+++ b/data/se.sjoerd.Graphs.gschema.xml
@@ -1,11 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <schemalist gettext-domain="graphs">
-  <enum id="se.sjoerd.Graphs.system-styles">
-    <value nick="adwaita" value="0"/>
-    <value nick="seaborn" value="1"/>
-    <value nick="yaru" value="2"/>
-  </enum>
-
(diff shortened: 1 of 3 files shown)


cmkohnen (commented):
My initial thought was to add an override like described [here](https://docs.gtk.org/gio/class.Settings.html#vendor-overrides) for the snap package. But since this solution would be just package agnostic not distro agnostic (as your solution is), I'm definetely in favour of this approach. However I feel using a setting then is maybe redundant. I'd just move this check in the styles.update() function and drop the setting.

cmkohnen (approved):

PR #522, opened 2023-09-29
Fix yaru theme

Sjoerd1993:
After trying out the Snap package, I found two errors:
- When the Snap package is ran (which I found is detected correctly, so that's good), and Yaru is not set as theme, `system_stylename` was never declared. This PR fixes this by setting adwaita first as default, and then overriding this for the Snap package in case Yaru is set as the theme
- For some reason, when getting the system theme from the update method in `styles.py`, it always returns `Adwaita-empty`. Not the actual theme. This is fixed by setting the theme upon init as property of the application. Ideally, I'd like it to run a `Gtk.Settings.get_default().get_property("gtk-theme-name")` when getting this attribute. But I haven't succeeded in getting this to work from the Vala class. Either way when I try to import Gtk there, it complains about Settings being ambigious between Glib.settings and Gtk.Settings for some reason. I could just set such function in the main application, but that felt less clean than using the Vala class here, and it would probably return Adwaita-empty anyway. 

Either way, the solution in this PR works. It's just that when having Graphs open, and then setting another theme in GNOME tweaks, that it doesn't immediately adapt until restarting the application. Which is a relatively minor issue. But I'm open for suggestions to run a query for the Gtk theme in the Vala class directly.

### DIFF
--- a/src/application.py
+++ b/src/application.py
@@ -9,7 +9,7 @@ import logging
 
 from gettext import gettext as _
 
-from gi.repository import GLib, Gio, Graphs
+from gi.repository import GLib, Gio, Graphs, Gtk
 
 from graphs import actions, migrate, ui
 from graphs.data import Data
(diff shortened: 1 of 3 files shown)


Sjoerd1993:
The first point already got resolved by PR #521

cmkohnen (approved):

PR #523, opened 2023-09-29
Add StatusPage to empty stylesheets

Sjoerd1993:
Adds a StatusPage to empty stylesheet editor, so it is clear that a user needs to add their own custom styles there and that the window is not just broken.

image

In general, I really like this change of seperating the custom and system styles by the way. As it reduces the risk of ruining existing stylesheets, makes it easier to fall back to these system defaults. And it removes some clutter. We've got the existing stylesheets that are shipped with Graphs, and then the user can create and edit their own here. That's nice. Also fixes the problem with newly added stylesheets (like Yaru) not being imported to existing users.


cmkohnen (commented):
This was one of my next to dos as well, see comment for small improvement

Sjoerd1993:
> This was one of my next to dos as well, see comment for small improvement

Yeah that's clearly better

cmkohnen (approved):

PR #524, opened 2023-09-29
Fix missing style error

Sjoerd1993:
Currently, when a user deletes a style that is set as current custom style (or when the current custom style is deleted for whatever other reason), and then opens figure settings, it will error out when trying to set the list with style options.  In this PR, it sets the style to adwaita instead as a fallback.

Perhaps another option is to just give an error when trying to delete the current used custom style, but this fallback also protects against e.g. removing the flatpak user folder (which one may do as nuclear option to reset everything to default).


src/figure_settings.py
```diff
@@ -54,8 +54,12 @@ def __init__(self, application, highlighted=None):
...
         )
         styles_ = styles.get_available_stylenames()
-        style_index = styles_.index(
-            self.props.figure_settings.get_custom_style())
+        try:
+            style_index = styles_.index(
+                self.props.figure_settings.get_custom_style())
+        except ValueError:
+            style_index = styles_.index("adwaita")
+            self.props.figure_settings.set_custom_style("adwaita")
```
Sjoerd1993, line 62: I am not 100% convinced that it should also change the style directly to adwaita. I just thought it may look weird that the set style does not match with Adwaita. 

On the other hand, I think it may be more preferable for people if it didn't suddenly override the style (and likely all line properties as well). So I'm thinking about removing line 62. Furthermore, in that case we can also just take index 0 (the top choice), but again adwaita seemed like a sensible default.

cmkohnen:
Huh, in theory styles.update() should prevent such issues (by doing essentially the same logic), Are you noticing any issues in testing or is this a fix for a theoretical issue?

Sjoerd1993:
> Huh, in theory styles.update() should prevent such issues (by doing essentially the same logic)

I noticed this during testing. The thing is that the error is not raised in styles.update(), but instead upon the figure settings init:

```
            style_index = styles_.index(
                self.props.figure_settings.get_custom_style())
```

When deleting a custom style, the current custom style in settings is still in-tact. So in this line, it tries to find the index that corresponds to this custom style that was deleted. This index does not exist, so it throws a ValueError.

cmkohnen:
Ah, I see. A better fix then is to check if the style to be deleted is in use and queue a reload in that case, in contrast to only fixing the issue once the figure settings window opens.

Sjoerd1993:
So I tested this, and the case of removing the style manually was a theoretical one. That was indeed already fixed a while ago. So if I remove the existing default style by hand, it just falls back automatically, exaclty a it should.

Will reduce this PR to just this lonely added line.

cmkohnen:
I believe one problem would still persist, that being if the user manually modifies styles inside the directory. Perhaps instead of reloading in all cases, where the style could have changed we implement a watchdog, that reloads the canvas if the currently selected style has changed or was deleted.

Sjoerd1993:
So I was checked out at at main, that was **not** on purpose. Hence the 0/0 changelog. It should just be the update line after removal that is changed though.

Sjoerd1993:
> if the user manually modifies styles inside the directory

As long as they modify the styles correctly (that is, their modifications are valid and don't lead to syntax errors etc..), this still works perfectly fine. I tried this, and it works also when the applciation is still running but obviously only applies upon reloads. So e.g. when setting a different style and back.

In principle we could fix that by catching such errors in parse_style, but that's overall a different issue than what this PR tries/tried to solve. 
```
  File "/home/sjoerd/.local/lib/python3.11/site-packages/matplotlib/__init__.py", line 722, in __setitem__
    raise ValueError(f"Key {key}: {ve}") from None
ValueError: Key axes.grid: Cannot convert 'truasde' to bool
```
Also, in general it's quite clear what happened if one has been messing with a style by hand and it suddenly doesn't start. So it's not the highest priority in the world. But it would not hurt to have some fallback if a style doesn't parse. Perhaps we can just do a similar fallback directly when `parse_style` doesn't succeed.

Also, when this leads to errors it's likely

Sjoerd1993:
Added such a fallback in case the style is broken. It catches syntaxerror, valuerror, attributerror and keyerrors. I think that should reasonably be the ones that can possible turn up when messing up the document.

cmkohnen (approved):

PR #525, opened 2023-09-29
add watchdog for style directory

cmkohnen:
Add a watchdog for the style directory so in case the used style gets deleted by whatever means, a reload is queued

### DIFF
--- a/src/application.py
+++ b/src/application.py
@@ -10,7 +10,7 @@ from gettext import gettext as _
 
 from gi.repository import GLib, Gio, Graphs, Gtk
 
-from graphs import actions, migrate, ui
+from graphs import actions, migrate, styles, ui, utilities
 from graphs.data import Data
 
 from matplotlib import font_manager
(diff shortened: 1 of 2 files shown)


Sjoerd1993 (approved):
Relatively obscure, but indeed I could actually trigger the issue that this PR tried to solve relatively easily by setting a custom style and then deleting it whilst Graphs was running. This PR solved the issue!

This works really well, and is nice and responsive. But we could also reload the canvas on any change maybe? That way, if someone changes the mplstyle by hand, it immediately updates in the actual plot. I doubt many will actually use that so I'm approving this either way, but it works quite well when I tested this (in the shown code in the screencast, it's a bit redundant of course to do all checks to do the same thing anyway).

[Screencast from 2023-09-29 16-17-19.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/be9231af-09de-47ff-a4c1-0d6790d44019)

cmkohnen:
Yes, this would absolutely be possible and trivial to implement, we'd just have to whitelist that event

cmkohnen:
I could also now see this as an advanced feature, where users can drop their own mplstyle files in the directory and have a bit more in-depth customization options in real-time

PR #526, opened 2023-09-29
Fix shift

Sjoerd1993:
Gives more consistently evenly spacing with the shift operation. Specifically when only part of the curve is selected, the shift was increasing between each curve since the ymax-ymin was continuously increasing.

Also fixes a bug in the select data part of the code, where it gave errors when span limits were outside the data limits.

Old behaviour:
image

New behaviour:
image


src/actions.py
```diff
@@ -24,11 +24,17 @@ def perform_operation(_action, target, self):
     if operation in ["center"]:
         args = [self.get_settings("general").get_enum(operation)]
     if operation == "shift":
+        canvas = self.get_window().get_canvas()
+        right_range = (canvas.top_right_axis.get_ylim()[1]
+                       - canvas.top_right_axis.get_ylim()[0])
+        left_range = (canvas.axis.get_ylim()[1]
+                      - canvas.axis.get_ylim()[0])
```
cmkohnen, line 31: I'd rather use figure_settings here, as it is already in scope and should store the exact same values

Sjoerd1993:
There's actually a regression in the log logic here, which is caught by the unit test. Will fix that tomorrow

cmkohnen (approved):

PR #527, opened 2023-09-29
Migrate Styles

cmkohnen:
Migrates user styles. If the style differs from a template or doesn't have one it is kept, else it is just deleted.

### DIFF
--- a/src/migrate.py
+++ b/src/migrate.py
@@ -3,7 +3,9 @@ import contextlib
 import pickle
 import sys
 
-from graphs import file_io, misc, utilities
+from gi.repository import GLib, Gio
+
+from graphs import file_io, misc, styles, utilities
 
 
 CONFIG_MIGRATION_TABLE = {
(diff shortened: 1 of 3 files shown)


src/migrate.py
```diff
@@ -74,10 +82,34 @@ def _migrate_import_params(settings_, import_file):
...
     import_file.delete(None)
 
 
+def _migrate_styles(old_styles_dir, new_config_dir):
+    new_styles_dir = new_config_dir.get_child_for_display_name("styles")
+    if not new_styles_dir.query_exists(None):
+        new_styles_dir.make_directory_with_parents()
+    system_styles = styles.get_system_styles()
+    enumerator = old_styles_dir.enumerate_children("default::*", 0, None)
+    while 1:
```
Sjoerd1993, line 95: Personally I generally prefer while True, but that's not really a blocker to me.

Sjoerd1993 (approved):

PR #530, opened 2023-10-02
Minor ui fixes

Sjoerd1993:
Very minor, but applies some of the quick and easy UI fixes from the latest GNOME Circle feedback.


Sjoerd1993:
So this uses the shorter "Add equation" for the second option in the add menu, removes the separator there, and uses the go-next and go-previous for the view arrows.

image

cmkohnen (approved):

PR #534, opened 2023-10-02
Redesign Style Selection & Editing

cmkohnen:


cmkohnen:
Bildschirmfoto vom 2023-10-03 10-20-46

Current design. Hover and the selected item still need work, but I haven't figured out how to do that in css yet.

cmkohnen:
Bildschirmfoto vom 2023-10-03 15-01-34
My current design iteration, feedback very much apprechiated, as it deviates a bit from the mockup.
Also very much wip, not functional and system is missing

cmkohnen:
Thanks! A few notes:
- Changing the color of the edit button based on the contrast was on my todo-list. A few solutions I found involve opencl or imagemagick. I'm trying to find a solution that works with either cairo or Pillow, as thats already in our stack.
- Regarding fonts, etc. The preview is done by rendering at a given size (currently 6x3in Aspect ratio 2:1). From there the image is scaled to 160x80px during runtime render (The reason I chose to work with svg, because thats low compute without artifacts. Any issues are on the mpl side during preview generation. Changing the height of the preview would work, but we'd have to increase the rendering height as well to preserve the aspect ratio else the view would break.
- I haven't really hooked up the file monitor to the preview generation and update mechanism so thats still todo
- Renaming System styles (at least their display names) was on my list as well. Since the display name is included in the style, that'd even be decoupled from the filename.

Sjoerd1993:
I think the fonts are fine. Choosing svg images make sense, and generally is the best way to do these things imo. 
Regarding the aspect ratio, it's very quick and easy to change these so it's not really the highest priority. I toyed around with it quickly, and a 4:3 aspect ratio is too tall, but I did kinda like 5:3 (at a size of 160x96px): 
image 

But again, it's kinda one of these things we can tweak whenever, not a big design change that will be difficult to alter. So the wider 2:1 aspect ratio is fine for now as well.

cmkohnen:
I do prefer the 5:3 aspect ratio as well (hence why I included it noew). I've updated the code so that the displayed (and used) stylename is the one specified in the file. Also since this new logic doesn't rely on manually reloading the style list and all sorts of race conditions caused by that, Editing, Creating and Deleting files in the style directory works and instantly updates the preview in the figure settings.

cmkohnen:
While this still needs work. This is now functional and visually (almost) where I personally want it to be. SUggestions on design always welcome!

Whats left:
- fix a minor selection bug
- add trash functionality
- change the edit color based on image contrast
- tweak style preview (and bundle sytem ones)

Sjoerd1993:
This indeed mostly works really nice now, really cool work! The UI as now is also mostly how I envisioned things, so that looks good aside from the mentioned issue (edit color having poor contrast on light/dark styles in dark/light mode respectively). 

I'm personally really busy at the moment, we've got a conference in Japan next week that I need to prepare for, and there's other work things as well. So the next two weeks I'll be a bit slow from the development side of things, but I will probably have time to review PR's and other such things.

cmkohnen:
Bildschirmfoto vom 2023-10-06 12-43-37
Automatic color change working now too. Threshold and colors may be tweaked.

Sjoerd1993:
Apologies for the slow review in my part, maybe I have some time tonight or tomorrow to look at it a bit more, but if I recall correctly the following issues still exist:

- the system style does not have a preview
- when editing the style that is currently active and the backing out, it changes the active style to the system style, which should not happen. 

But I'll take a look at it as a soon as I've got some free time with my computer. As far as I recall it otherwise works well, and this is one of my favorite changes in a long time. Really great work on this!

cmkohnen:
> Apologies for the slow review in my part

No worries, I have a lot on my plate currently as well

> * the system style does not have a preview

That is intentional (for now) as are the manually regenerated previews for system styles. The idea for them is when we are happy with the preview generation progress to bundle the image inside the gresource (and having the split version for system as discussed) to save time on each startup. However The style preview does require a bit more finetuning which I'd like to have in a separate pr, as this should only address the logic part

> * when editing the style that is currently active and the backing out, it changes the active style to the system style, which should not happen.


That is an issue I haven't come across, will look into it

bertob:
Just stumbled across this, looks great!

Just one small nitpick, I'd move the checkmark to the bottom right, and move it a little bit closer to the edge, like how Text Editor does it:

image

cmkohnen:
The reason I initially put it up there was, that it get's too close to the edit buttons imo.
Bildschirmfoto vom 2023-10-13 12-53-03

cmkohnen:
> * when editing the style that is currently active and the backing out, it changes the active style to the system style, which should not happen.

I haven't found a proper solution due to all sorts of race conditions. The workaround I've implemented is, when a selected style is edited, the system style is selected automatically. This solves both mentioned bug as well as live updates not being "missing"

Sjoerd1993:
> I haven't found a proper solution due to all sorts of race conditions. The workaround I've implemented is, when a selected style is edited, the system style is selected automatically. This solves both mentioned bug as well as live updates not being "missing"

I noticed this when testing this. It works at least a lot nicer, as the previous bug lead to serious performance issues when it happened. For now I think the workaround works well enough to merge this. I did notice one more thing, which is that the line colors (the colour dots themselves) do not update when changing colors. 

It works well using e.g. the following logic in the line color box:

```
@Gtk.Template(resource_path="/se/sjoerd/Graphs/ui/style_color_box.ui")
class StyleColorBox(Gtk.Box):
    __gtype_name__ = "GraphsStyleColorBox"
    label = Gtk.Template.Child()
    color_button = Gtk.Template.Child()

    parent = GObject.Property(type=StyleEditor)
    index = GObject.Property(type=int, default=0)

    def __init__(self, parent, index):
        super().__init__(parent=parent, index=index)
        self.label.set_label(_("Color {}").format(index + 1))
        self.provider = Gtk.CssProvider()
        self.color_button.get_style_context().add_provider(
            self.provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION,
        )
        self.color_button.connect("notify", self._reload_color)
        self._reload_color()

    def _reload_color(self, *args):
        color = self.props.parent.line_colors[self.props.index]
        self.provider.load_from_data(
            f"button {{ color: {color}; }}", -1,
        )
```
Not sure if the unused `args` will set off the linter.

cmkohnen:
> I did notice one more thing, which is that the line colors (the colour dots themselves) do not update when changing colors.


I forgot to add one line in the color changing logic. Works now

bertob:
> The reason I initially put it up there was, that it get's too close to the edit buttons imo.

Ah, I guess those show on hover/long press?

Could probably just move one of those buttons to another corner, e.g. top left? Also, they probably need some kind of background, probably .osd would work?

cmkohnen:
> Ah, I guess those show on hover/long press?

These show up for styles, the user has added. And for those they are always visible.

> Could probably just move one of those buttons to another corner, e.g. top left? Also, they probably need some kind of background, probably .osd would work?

They are somewhat related to each other as one enters the edit mode and the other deletes the style respectively.

One idea how this could work is the following:
Bildschirmfoto vom 2023-10-13 16-11-05

cmkohnen:
Although one Idea (that was initially in my mockup) was removing the delete button from the overview and moving it inside the editor

Sjoerd1993 (approved):
For my part this works quite well with the remove option in the style editor itself.

As this is kinda a blocker for some other PRs, I'd say this is good for merging now. Further tweaks to e.g. UI can be done with separate Pr's.

PR #542, opened 2023-10-04
Improve spacing in figure setting limits category

Sjoerd1993:
The current main branch has a weird spacing below the axes limits in Figure Settings. This is due to the entries themselves being hidden, but the box where they are in is not hidden. An empty box still takes some spacing, hence the spacing in the axes limits being wrong.

This PR fixes this, by hiding the boxes themselves instead of the entries. So the spacing is correct. 
Furhermore, the entire group with the limits is put into it's own box. This is so that a single `spacing` attribute can be set for all these widgets. This is necessary, as the current spacing set individually in each box means that there's a bottom-spacing on the lowest entry. (Unless that entry happens to be the limits for the top axis, but in 95% of the cases one works with left and bottom). Grouping these in a box, and setting a spacing attribute, means that there won't be any new artificial spacing between the Limits group and the Scaling group.

### DIFF
--- a/data/ui/figure_settings.blp
+++ b/data/ui/figure_settings.blp
@@ -41,86 +41,81 @@ template $GraphsFigureSettingsWindow : Adw.PreferencesWindow {
 
     Adw.PreferencesGroup {
       title: _("Axis Limits");
-      Adw.ActionRow no_data_message {
-        title: _("No Data");
-        subtitle: _("Add data to the figure to set axis limits");
+    	Adw.PreferencesGroup no_data_message {
+        Adw.ActionRow {
+          title: _("No Data");
+          subtitle: _("Add data to the figure to set axis limits");
+        }
       }
       Box {
-      	orientation: horizontal;
-      	margin-bottom: 12;
-      	Adw.PreferencesGroup {
-      	  margin-end: 3;
-      	  Adw.EntryRow min_left {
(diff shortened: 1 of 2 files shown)


data/ui/figure_settings.blp
```diff
@@ -41,86 +41,81 @@ template $GraphsFigureSettingsWindow : Adw.PreferencesWindow {
...
-      	  margin-end: 3;
-      	  Adw.EntryRow min_left {
-	          title: _("Minimum Left-Hand Y-Axis");
-	          visible: false;
-	        }
+        orientation: vertical;
+        spacing: 12;
+        visible: bind no_data_message.visible inverted;
+        Box left_limits {
+        	orientation: horizontal;
```
cmkohnen, line 55: The indentation is all over the place here, perhaps your builder config is set to tabs instead of spaces?

Sjoerd1993: That's so weird, it looked all fine in Builder. But double checking it, this commit has a random mix of spaces and tabs. Not sure why that happened, as I double checked and it is set to spaces. Should be OK though.

cmkohnen (approved):
This has been bothering me for a while now, thanks for fixing it.

Here goes me hoping this doesn't introduce merge conflicts with my branch

PR #544, opened 2023-10-10
Move default position to figure settings

cmkohnen:
Based on the latest Style redesign.

Makes Default X and Y positions part of the figure settings. This way it is project specific and more logically placed (imo).


Sjoerd1993:
I think it's fine if this setting is project based, but I'm not 100% sure it logically should be in figure settings, as this is not setting for the active figure but rather a more general setting. 

As @bertob mentioned, there's not a lot of use legitimate cases I can think of where this setting should be anything other than bottom/left _by default_ for newly generated data. So I'm personally fine with removing the option as well.

Sjoerd1993:
Either way, I think in a general workflow it's probably most intuitive to set this in the figure settings instead of the general settings anyway. So it's definitely an improvement in my books.

Having it coupled to the project also makes sense for a more niche setting like this.

I'm just wondering if it maybe is worth it to remove this default option altogether.

cmkohnen:
I'd be in favour of removing these as well. Expecially with the logic we introduced a while ago that separates items based on their labels. I can see no reason to plot everything to the top and right axis by default, as these axes are most commonly used for comparing data on different scales.
Will make another pr then.

PR #545, opened 2023-10-11
Split of Titlesize

cmkohnen:
Based on https://github.com/Sjoerd1993/Graphs/pull/534.

Makes the Titlesize a separately adjustable value in the style editor. Instead of absolute values however, it is put in relation to the main fontsize and limited to 100% - 200%


Sjoerd1993:
Yeah I quite like the idea. On a similar note, the label values and label text should be split as well. But this is a good change. Will try and give it a rest asap.

cmkohnen:
> On a similar note, the label values and label text should be split as well.

I also though about this. My suggestion for that would be keep the labelsize like the main size but put ticklabels between 50% and 100% of that.

Sjoerd1993 (approved):
Reason for not approving this earlier was because I wanted #534 in earlier.

PR #546, opened 2023-10-12
Remove default position setting & optimize xry import

cmkohnen:
As discussed removes the preferences for the default x and y position.
This also means items no longer need to get a reference to the application on init.

Also optimize xry import slightly.


Sjoerd1993 (approved):
Creating new items does become a lot cleaner without the reference to the application. The preferences is really a lot more sparse compared to 1.6.1, but overall without losing any useful functionality so that's likely an improvement.

PR #548, opened 2023-10-15
General cleanup

cmkohnen:
Based on: #547 

General code cleanup throughout the project


cmkohnen:
I've done some reading and as it turns out, this implementation for checking if limits are close is both more accurate and more performant than numpys implementation.

PR #550, opened 2023-10-16
General improvements and fixes

cmkohnen:
- Disallow the usage of savefig parameters in styles
- split of file parsing and writing
- misc fixes
- Use __iter__ functions for iterating over directories and models

### DIFF
--- a/data/styles/classic.mplstyle
+++ b/data/styles/classic.mplstyle
@@ -246,15 +246,6 @@ path.simplify_threshold: 0.1111111111111111
 path.snap: True
 path.sketch: None
 
-savefig.dpi: 100
-savefig.facecolor: w
-savefig.edgecolor: w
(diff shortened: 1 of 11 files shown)


data/styles/classic.mplstyle
```diff
@@ -246,15 +246,6 @@ path.simplify_threshold: 0.1111111111111111
 path.snap: True
 path.sketch: None
 
-savefig.dpi: 100
-savefig.facecolor: w
-savefig.edgecolor: w
-savefig.format: png
-savefig.bbox: standard
```
Sjoerd1993, line 253: Never really noticed these, but yeah those shouldn't be in the stylesheet. Savefig options should be of course defined in the export figure dialog

Sjoerd1993 (approved):

PR #553, opened 2023-10-18
Update metadata

Sjoerd1993:
Not really a consequential change, but updates the metadata. The .doap file was outdated, I fixed some typos in the README, and updated the changelist for the next 1.7 release. I also noticed that a bug fix is tagged as a 1.6.2 release, which I'm pretty sure was actually put into the 1.6.1 version last-minute.

I also bumped the version to `1.7.0-beta`, while I'm not planning to announce a public beta just yet (not before we've worked through the current open feedback from GNOME Circle), I thought it makes more sense than using this 1.6.1 tag which is a bit outdated at this point for this build. 

Regarding the timing for the actual release. I think it's probably the most responsible thing to maintain a "it's finished when it's finished" attitude on this. At the least, the current open issues on GNOME Circle should be addressed, the bugs in the release milestone should be fixed, and it would probably be nice to have a period of just testing where we have the beta open.


README.md
```diff
@@ -75,9 +75,9 @@ If you want to build without Flatpak anyway these instructions might help:
 
 build-time dependencies: `meson, blueprint-compiler, gettext`
 
-runtime dependencies: `matplotlib, python3-matplotlib-gtk4, scipy, numpy`
+runtime dependencies: `matplotlib, python3-matplotlib-gtk4, scipy, numpy, numexpr, sympy`
 
-The actual package names might vary depending on your distribution.
+This actual package names might vary depending on your distribution, and depending on your distribution additional packages may be required.
```
cmkohnen, line 80: Why the change to "This" here?

cmkohnen:
> Regarding the timing for the actual release. I think it's probably the most responsible thing to maintain a "it's finished when it's finished" attitude on this. At the least, the current open issues on GNOME Circle should be addressed, the bugs in the release milestone should be fixed, and it would probably be nice to have a period of just testing where we have the beta open.

I agree. This is also a very big release (probably the biggest jet), so it should recieve a lot of polish and testing.

README.md
Sjoerd1993: Yeah that's a mistake somehow, not sure why I changed that.

 I remember that I ran into issues with GTK libraries that also had to be installed on Silverblue (which don't come in the Fedora Toolbox), so I thought it makes sense to add the latter part that some additional packages may be required as well. (gtk and libadwaita dev packages mainly). 
 
 Probably wanted to rewrite the sentence a bit differently and started on that, but then decided on appending this new sentence at the end. Either way, the "This" change should be gone, thanks for spotting that.

cmkohnen (approved):

PR #556, opened 2023-10-25
implement FillItem

cmkohnen:
Adds a FillItem.
This is both the foundation for the current fill inside curve_fitting and future fills under select curves.

Also use tuples in cases, where arraylikes shouldn't be mutable

### DIFF
--- a/src/actions.py
+++ b/src/actions.py
@@ -20,7 +20,7 @@ def perform_operation(_action, target, self):
     elif operation == "cut" and self.get_mode() != 2:
         return
     args = []
-    if operation in ["center"]:
+    if operation in ("center"):
         args = [self.get_settings("general").get_enum(operation)]
     if operation == "shift":
         figure_settings = self.get_data().get_figure_settings()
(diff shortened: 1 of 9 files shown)


src/actions.py
```diff
@@ -20,7 +20,7 @@ def perform_operation(_action, target, self):
     elif operation == "cut" and self.get_mode() != 2:
         return
     args = []
-    if operation in ["center"]:
+    if operation in ("center"):
```
Sjoerd1993, line 23: Not something related to this PR directly, but now I'm seeing this I wonder if there's any reason why we're not just using `if operation == "center"`

Sjoerd1993 (approved):
Using tuples is a minor thing, but it makes sense. Regarding the fillitem, the existing `fillbetween`  function in Matplotlib is perfectly fit for the purpose in curve-fitting. But I agree that it's nice to have the foundation here so we can support fills under other curves as well. And then it's probably better to use the same logic in both cases indeed.

src/actions.py
cmkohnen: The reason for this was to keep it modular. In case other operations will get a dedicated setting, they should just be added to this list. This worked using list brackets but tuple brackets are ignored if a trailing comma isn't present. Fixed that now.

Sjoerd1993: Ah right, yeah that makes sense.

PR #557, opened 2023-10-25
Use rc context

cmkohnen:
Based on #556 

Use instanced version of rcParams instead of global one.
This should no longer break custom installs, where Graphs was running and somebody ran a simple mpl script (although that was never supported anyway)
This also allows to have the curve fitting canvas always follow the system:
Bildschirmfoto vom 2023-10-25 18-28-49


Sjoerd1993 (approved):
Probably good to merge #556 first, but this looks good to me. Makes things a bit easier when a different canvas is used, such as with the curve fitting.

PR #559, opened 2023-10-30
Update style upon init

Sjoerd1993:
Calls update_style upon init of the style manager, as `self._system_style_params` needs to be initialised. Otherwise the current main branch does not launch.

### DIFF
--- a/src/styles.py
+++ b/src/styles.py
@@ -280,7 +280,7 @@ class StyleManager(GObject.Object, Graphs.StyleManagerInterface):
         style_io.write(destination, source)
 
     def complete_style(self, params):
-        for key, value in self.get_system_style_params().items():
+        for key, value in self._system_style_params.items():
             if key not in params:
                 params[key] = value
         return params
--- a/src/styles.py
+++ b/src/styles.py
@@ -53,6 +53,7 @@ class StyleManager(GObject.Object, Graphs.StyleManagerInterface):
             application=application, gtk_theme=gtk_theme.lower(),
             style_model=Gio.ListStore.new(Style),
         )
+        self._update_styles()
         self._stylenames = []
         self._cache_dir = utilities.get_cache_directory()
         if not self._cache_dir.query_exists(None):
--- a/src/styles.py
+++ b/src/styles.py
@@ -280,7 +280,7 @@ class StyleManager(GObject.Object, Graphs.StyleManagerInterface):
         style_io.write(destination, source)
 
     def complete_style(self, params):
-        for key, value in self._system_style_params.items():
+        for key, value in self.get_system_style_params().items():
             if key not in params:
                 params[key] = value
         return params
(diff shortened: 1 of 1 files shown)


src/styles.py
```diff
@@ -279,7 +280,7 @@ def copy_style(self, template: str, new_name: str):
         style_io.write(destination, source)
 
     def complete_style(self, params):
-        for key, value in self._system_style_params.items():
+        for key, value in self.get_system_style_params().items():
```
cmkohnen, line 283: Any reason to change it to the getter?

cmkohnen (changes requested):
I see my mistake here. I haven't tested with custom styles installed. However, this adds another problem: If the settings for default style are set to some custom one, this doesn't work either. A solution is to split the _update_styles() function into _update_system_style() and _update_selected_style() respectively.

src/styles.py
Sjoerd1993: Not really, except for that I generally try to avoid accessing underscored attributes directly, especially if a getter exists. But in practice it will end up the same.

Sjoerd1993:
> If the settings for default style are set to some custom one, this doesn't work either

I didn't really found any issues when I tested just that. I'm guessing the hypothetical issue would come from the above part of the code:
```
                        style_params = style_io.parse(style.file)
                        if not style.mutable:
                            style_params = self.complete_style(style_params)
                        self._selected_style_params = style_params
```
Where the `style_params = self.complete_style(style_params)` line get skipped for custom styles. But `style_params` is already neatly defined in the line before, which worked for me when I tested this just now.

src/styles.py
cmkohnen: The reason here being, that a getter will introduce (a very small) overhead. Since we are in the class, a private access would be fine imo. From outside I'm very much with you.

cmkohnen:
Oh, this seems to be another mistake. The code should be invoked *if* the style is mutable

cmkohnen (approved):

Sjoerd1993:
> The code should be invoked if the style is mutable

I see, looking at my previous comment, I didn't even notice myself either. As I said this line gets skipped for custom styles. Correcting this does not change too much with the issue, `style_params` is still assigned both for custom styles and system styles.

Just to double check, the reason for not calling `self.complete_style(style_params)` on custom styles, is just overhead right? The parameters are all written to the file when creating the style. So it would be redundant.

cmkohnen:
> Just to double check, the reason for not calling `self.complete_style(style_params)` on custom styles, is just overhead right? The parameters are all written to the file when creating the style. So it would be redundant.

Kind of, yeah. Mutable means, the style is inside the user folder. Not mutable ones are included in the gresource. We assume that styles in the gresource are valid and fulfill all requirements we set. This cannot be said for the custom ones inside the style folder hence this function

Issue #563, opened 2023-11-08
Consider removing item override on style preference option.


cmkohnen:
Currently we check based on a preference if all items should be set to the new styles default options.

However I propose a different solution:
On a per-item per-property basis, check if that property is the default for the previously selected style (easy with latest refactors). If it is, that property is updated, if not, it is left the way it is.

Another variation could be, if only one property is not the default, leave the whole item as is.

My preference would be on the first suggestion. Feedback highly apprechiated.

Sjoerd1993:
I think this is a reasonable thing to do, to further streamline the settings. 

To me, the two most sensible options are to either override everything, or go for the first mentioned option of overriding the defaults on a per-property basis (the first option). I feel like overriding on a per-item per-property basis would probably be the option that would lead to the most sensible workflow. That would allow to apply a style sheet on the open project, without losing any of the things you've changed. So I think that would lead to the least friction in most workflows.

PR #565, opened 2023-11-10
rework resetting item properties on style change

cmkohnen:
As discussed in #563, remove the preference option in regards to item override.

For visual reasons this also removes the grouping in the preferences, as there are only three items now.
Creating items has also been made more modular using some of the logic necessary to handle reset() properly.
Also revert usage of `__gtype_name__` to check for instances of Item. This allows to have future items that inherit from e. g. DataItem.


Sjoerd1993 (approved):
This works well for me, also with the intended purpose of not overwriting custom parameters. It's on this build where I found bug #567, which is not related to this particular PR.

PR #566, opened 2023-11-13
tweak style parameters

cmkohnen:
Slightly tweak the colors of the adwaita stylesheet for the Graphs. See below for previews:

Bildschirmfoto vom 2023-11-13 10-34-50
Bildschirmfoto vom 2023-11-13 10-34-40

### DIFF
--- a/data/styles/adwaita-dark.mplstyle
+++ b/data/styles/adwaita-dark.mplstyle
@@ -82,10 +82,10 @@ text.color: FFFFFF
 axes.labelcolor: FFFFFF
 xtick.labelcolor: FFFFFF
 ytick.labelcolor: FFFFFF
-xtick.color: FFFFFF
-ytick.color: FFFFFF
-axes.edgecolor: FFFFFF
-grid.color: A7A7A7
+xtick.color: C0BFBC
+ytick.color: C0BFBC
+axes.edgecolor: C0BFBC
+grid.color: 77767B
 axes.facecolor: 242424
 figure.facecolor: 242424
 figure.edgecolor: 242424
--- a/data/styles/adwaita.mplstyle
+++ b/data/styles/adwaita.mplstyle
@@ -82,10 +82,10 @@ text.color: 000000
 axes.labelcolor: 000000
 xtick.labelcolor: 000000
 ytick.labelcolor: 000000
-xtick.color: 000000
-ytick.color: 000000
-axes.edgecolor: 000000
-grid.color: 646464
+xtick.color: 5E5C64
+ytick.color: 5E5C64
+axes.edgecolor: 5E5C64
+grid.color: 9A9996
 axes.facecolor: FFFFFF
 figure.facecolor: FAFAFA
 figure.edgecolor: FAFAFA


cmkohnen:
Also, @bertob: since this is trying to adress parts of the GNOME Circle review, suggestions for further changes are highly appreciated.

Sjoerd1993:
The fainter llines do look easier on the eyes. It may be the way contrast works, but the graph border in dark mode _feels_ thicker to my eyes than in light mode. Even though I know they have the same thickness. Maybe it's not important, but I wonder if we should either make the border slightly thinner in dark mode to compensate for that.

bertob:
Toning down the grid looks better, but I'd also tone down the axes at the top and right when they're not being used as axes. The border on all sides looks a bit like a frame, rather than axes.

Generally it's true that white-on-dark lines of the same thickness look thicker than black-on-white because of how light bleeds outwards, but not sure that's a big problem here.

Sjoerd1993:
> The border on all sides looks a bit like a frame, rather than axes.

It's interesting that you take this up, as by default this is intended to be a frame that boxes the entire graph. So this does make sense. However, it's perfectly possible to only draw such a frame around active axes, see the examples from the [matplotlib docs](https://matplotlib.org/stable/gallery/spines/spines.html#sphx-glr-gallery-spines-spines-py):

image

Were we currently default to the top option, and you're basically suggesting the bottom option. Perhaps this should be something that is part of the style sheet itself. In principle, with powerful plotting tools it is important to be flexible and have a broad set of options. I'm not opposed to adding the option to set a frame around the entire plot. That could just be a single switch. "Draw frame around the graph", maybe slightly different wording. If that is turned off, go for the second option in the graph and only draw the axes that are used. Alternatively we can do this granularly and set an on/off switch for each axis.

 Another option is to couple the presence of such a border with the presence of ticks, but that might be more confusing than helpful.

Sjoerd1993 (approved):
Whether we'll rework the splines such that we can "turn off" the frame around all borders is a thing for a different PR imo. This PR is good to go for me.

Sjoerd1993:
Probably under the hood, we can do this also by just setting the linewidth on a per-axis basis. This fits quite well with the current code also, thanks to the simple use of a dictionary style object for `ax.spines`:

```
import matplotlib.pyplot as plt

# Create some example data
x = [1, 2, 3, 4, 5]
y = [2, 4, 6, 8, 10]

# Create a figure and axis
fig, ax = plt.subplots()

# Plot the data
ax.plot(x, y)

# Customize the appearance of the spines
ax.spines['left'].set_linewidth(1.25)    # Set left y-axis spine to be thick
ax.spines['right'].set_linewidth(0.5)  # Set right y-axis spine to be thin
ax.spines['bottom'].set_linewidth(1.25) # Set bottom x-axis spine to be thin
ax.spines['top'].set_linewidth(0.5)    # Set top x-axis spine to be thin

# Show the plot
plt.show()
```

This looks like this:
image
Similarly, the color of the spines can also be set. So we can just make the non-active spines grey as well. But I think this particular setting should probably be done over the thickness.

I'm not fully convinced of just adding four more sliders for each axis where we can set the thickness of each axis. I think it's important for Graphs to be versatile, as my ambition is that it's a viable tool to make professional scientific plots. For that purpose, you want to have the flexibility to truly adapt it to your use-case. But at the same time, that shouldn't be at the cost of usability. I think a good example of striking a balance is the recent addition that allows to set the title size in relation to the text size, instead of having an individual font setting for that.

Perhaps in this case, a simple "Use a border for unused axes" switch would suffice. The question then kinda becomes if we'd still include a subtle border if that switch is turned off. My gut feeling would be "no".

Sjoerd1993:
Just to add, it may be obvious, but setting the linewidth of a spine to zero makes it dissapear:

image

cmkohnen:
I'd argue in the sake of simplicity here: I'd add a switch like you suggested to turn (on/off) the frame in the figure settings. I'm not sure which should be the default currently.
But that can be part of another pr as this adds functionality (and would break projects again).

Issue #567, opened 2023-11-13
Ticks don't work on linear scale


Sjoerd1993:
I noticed that ticks do not work on a linear scale. The weird thing is that they do work perfectly fine in the curve fitting tool, as well as in the generated previews. It's only in the actual main canvas that these do not work. I've taken some time to investigate, and the problem does not lie with the style or the logic that turns of the minor or major ticks.

Instead, the problem seems to lie with the AutoLocator, which for some reason does not work. I've got it fixed in a local build by simply setting the locator "by hand" for the linear scale, but I kinda want to check if there's a prettier solution as this may just be adressing the symptom rather than the cause:

This is the bit that I added to `scales.py` which solves the issue
```
class LinearScale(scale.LinearScale):
    """Linear Scale."""

    name = "linear"

    def set_default_locators_and_formatters(self, axis):
        super().set_default_locators_and_formatters(axis)
        axis.set_minor_locator(ticker.AutoMinorLocator())

```
After which the LinearScale needs to be registered of course, overwriting the bugged version: `scale.register_scale(LinearScale)`

PR #569, opened 2023-11-13
Do not revert rcparams

Sjoerd1993:
I found the root issue behind bug #567. The main issue is in line 122 in `canvas.py`:
```
        dict.update(pyplot.rcParams, orig_params)  # revert rcParams
```

The `orig_params` dictionary always holds `xtick.visible` and `ytick.visible` equal to `False`, as this is just simply the rcparams in the matplotlib defaults. For some reason, when setting `rcparams` back to the defaults, it copies the minor tick settings back to these defaults as well. I am a bit stumped about why this does not happen for e.g. `ytick.direction` though. If I set `        pyplot.rcParams["ytick.minor.visible"] = True` at the end of the init function, then the minor ticks on the y axis are turned on. Even though only rcParams have been updated, while these are not even used directly. From what I can tell, the `ax.tick_params` somehow gets trumped to whatever `pyplot.rcParams` has set. 

As far as I know, we do not need `rcParams` to be reset after init. But I am not 100% sure here, perhaps it's necessary to avoid left-overs from previous styles. I've noticed for instance that font settings do not always properly reset to the default value if a style with custom font was set, and then a style is set that does not have a font set. For this reason, this now updates the style to the default `rcParams` before applying a new style. So this issue with the fonts not setting seems fixed here at least.

### DIFF
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -87,7 +87,6 @@ class Canvas(FigureCanvas, Graphs.CanvasInterface):
         style context. Bind `items` to `data.items` and all figure settings
         attributes to their respective values.
         """
-        orig_params = dict(pyplot.rcParams.copy())
         self._style_params = style_params
         pyplot.rcParams.update(self._style_params)  # apply style_params
         GObject.Object.__init__(self, application=application, can_focus=False)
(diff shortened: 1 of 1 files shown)


cmkohnen:
Since we have a safeguard mechanism when loading custom styles anyway this redundancy is not really needed.

cmkohnen (approved):

Issue #570, opened 2023-11-14
Whether or not to frame the Graph


cmkohnen:
As discussed in #566 already a bit, there should be a way to have the Graph not boxed.
We can bind this to styles or to figure settings.

I'm not sure which would make more sense. Technically both is possible and would need custom logic as just using rcParams "axes.spines.*direction*" wouldn't work. see the following:
![Bildschirmfoto vom 2023-11-14 11-52-45](https://github.com/Sjoerd1993/Graphs/assets/59118042/fc832ca8-8fb8-4772-aa48-4c48ab2092ce)

Another thing would be to argue how fine grained the user should specify these values.
I'm on the more simple side here:
A simple switch to specify whether or not the Graph should be boxed would suffice. In logic we then hide all spines, which aren't used.

Ideas welcome

bertob:
This definitely looks like a better default, but also, the backgrounds on the graph and the right side of the window should be the same. Probably making the latter full-on white is better than having the graph on a gray background?

Sjoerd1993:
With regard to the screenshot, I guess the data is plotted on the right-hand and bottom axis, right? I take the custom logic is in turning the correct splines in or off.

This could be done with the splines line width as well, just setting them to zero. But if we'd go with a box/no box approach, then this would require custom logic regardless.

Either way, I've been thinking about the same and I also landed on the conclusion of allowing a user to use a full frame, or only draw the line for the used axes. As I struggle to find a single use case for turning off active axes, or for drawing only one of the unused axes. So I too think that single option would be the way to go.

I was also thinking about exactly the same choice about binding this to figure settings or to the style. My gut feeling says binding to the style would make most sense, as it is part of the general styling imo.

bertob:
What are the use cases for having a frame to begin with? Is this needed at all?

Sjoerd1993:
>  Probably making the latter full-on white is better than having the graph on a gray background?

This is of course a matter of whatever we decide to be the "System Style". The gray background was to match with the Adwaita theming. Whilst a white background on a graph just makes sense, it maximizes contrasts and it is typically the default as well. If someone saves/exports a graph, they probably want to share it in a different format and there white typically works best. So either way, I think a white background sounds most sensible to me. Although, one can of course just turn on a different style if they want to export a graph, or alternatively export with transparent background.

[...]

PR #571, opened 2023-11-14
Do not draw frame

Sjoerd1993:
Adapts some of the principles discussed in issue #570.
In this PR, only a line is drawn for the axes that are actually used, the ticks and line is hidden for unused axes. On top of that, the canvas background in light mode is set to match the rest of the application. See screenshot for the new default styling.

image

I'd also like to add the functionality to turn back on such a frame, I think it makes sense to set this on a per-style basis. Also, in that case it should be turned on in most of these style sheets, as those are all based on existing matplotlib styles that do have a frame around the entire plot. But again, I'm still open for discussion on this, also to the idea of integrating that in "Figure Settings" instead.

But adding the option to turn that on and skip this new logic could be done in a separate PR.

### DIFF
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -261,8 +261,8 @@ class Canvas(FigureCanvas, Graphs.CanvasInterface):
             # Set tick where requested, as long as that axis is not occupied
             # and visible
             axis.tick_params(which=ticks, **{
-                direction: draw_frame and not visible_axes[i]
-                or direction in directions
+                direction: (draw_frame and not visible_axes[i]
+                            or direction in directions)
                 and params[f"{'x' if i < 2 else 'y'}tick.{direction}"]
                 for i, direction in enumerate(possible_directions)
             })
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -246,21 +246,24 @@ class Canvas(FigureCanvas, Graphs.CanvasInterface):
             ["top", "right"],     # top_right_axis
         ]
         if not any(visible_axes):
-            visible_axes = [True, False, True, False]
+            visible_axes = [True, False, True, False] # Left and bottom
+            used_axes = [True, False, False, False] # self.axis visible
             self._legend_axis = self._axis
 
         params = self._style_params
         ticks = "both" if params["xtick.minor.visible"] else "major"
+        used_directions = []
         for directions, axis, used \
                 in zip(axes_directions, self.axes, used_axes):
             axis.get_xaxis().set_visible(False)
             axis.get_yaxis().set_visible(False)
             # Set tick where requested, as long as that axis is not occupied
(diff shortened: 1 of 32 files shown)


src/canvas.py
```diff
@@ -268,6 +270,17 @@ def items(self, items: list):
                 axis_legend.remove()
             if used:
                 self._legend_axis = axis
+                used_directions += directions
+
+        used_directions = list(set(used_directions))  # Remove duplicates
+        for direction in ["bottom", "left", "top", "right"]:
```
Sjoerd1993, line 276: Basically this entire loop could be skipped if the frame is turned on. Should be easy to adapt to that.

cmkohnen:
The logic has potential for optimization imo. This would be introducing a new loop structure where one is already present. Also in some cases ticks are set to visible only to be set to invisible again which can just be skipped.

See #573 for changes in this regard and the implementation of coupling this setting to the style

Sjoerd1993:
That was a nice change with #573, I did not really have the time to work on the implementation of adding the setting myself, but this works really well.

Sjoerd1993:
I just now noticed that when "Draw frame" is on, ticks are drawn regardless on unused axes, even if those ticks are turned off in the settings. Note that it does respect the settings if the axes are used.

I may have some time tomorrow morning to see if that can be fixed.

cmkohnen:
Huh, probaly an execution order thing as the `and params[...]` should turn them off if they aren't enabled in the style

cmkohnen:
One thing I would couple to this is drawing arrows like described [here](https://stackoverflow.com/a/58410781) or in the [mpl examples](https://matplotlib.org/stable/gallery/spines/centered_spines_with_arrows.html):
example

Sjoerd1993:
> Huh, probaly an execution order thing as the `and params[...]` should turn them off if they aren't enabled in the style

Yep, that diagnosis was correct. A couple of parentheses fixed it.

 ```
          axis.tick_params(which=ticks, **{
                direction: (draw_frame and not visible_axes[i]
                or direction in directions)
                and (params[f"{'x' if i < 2 else 'y'}tick.{direction}"])
                for i, direction in enumerate(possible_directions)
            })
```

The parentheses after `and (params...` are redundant strictly speaking, it's the one before the `and (params...` that fixed it. But I added them there anyway for clarity's sake.

Sjoerd1993:
I think such arrows would be a nice addition. If we were to make these optional however, we may need to make some non-trivial changes to the way we handle styles, as this would require custom parameters as far as I can tell.

I guess we can load and save the style as dictionary, and update these with rcparams on demand. But a custom class may be cleaner.

Either way, I'd suggest to have that in a separate PR in that case.

cmkohnen (approved):

PR #572, opened 2023-11-14
Redesign curve fitting dialog

Sjoerd1993:
Redesigns the curve fitting dialog, addresses [the points in the GNOME Circle feedback](https://gitlab.gnome.org/Teams/Circle/-/issues/185), and makes the dialog more responsive for different screen sizes.

Before:
image

After:
image

### DIFF
--- a/data/ui/curve_fitting.blp
+++ b/data/ui/curve_fitting.blp
@@ -8,7 +8,7 @@ template $GraphsCurveFittingTool : Adw.Window {
   default-width: 1000;
   default-height: 650;
   width-request: 305;
-  height-request: 430;
+  height-request: 410;
 
   ShortcutController {
     Shortcut {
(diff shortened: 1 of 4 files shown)


cmkohnen:
The "Graphs (Development)" Title on this feels off.
Also since this is kind of a modal dialog, I don't know if this style of sidebar would fit like it does in the main application

Sjoerd1993:
The "Graphs" title on the left probably feels better for a stable build, so without the "(Development)" part. But I think I agree with you that the Splitview look is probably a bit weird with a modal window. It kinda feels like a second application that's stuck to the main window, instead of it being a modal dialog.

The other issue, that did annoy me somewhat, is that the Action buttons simply clash with this design. If you resize the window the cancel button will be in the sidebar. So you will first have to open the sidebar before getting this button. Adding it to the content view in compact mode was also a bit weird. So if we'd go with a Splitview, we'd probably have to change the controlls a bit.

I updated it to be more in line with regular modal windows, so closer to the orginal. See screenshot:

image

Sjoerd1993:
Made some more tweaks, this now also disables interaction with the curve fitting window.

image

cmkohnen (approved):

Sjoerd1993:
For reference, here's with the updated figure style:

image

image

PR #576, opened 2023-11-15
Update dependencies

cmkohnen:
Routine update of dependencies as well as renaming for consistency and adding x-checker-data

### DIFF
--- a/python-dependencies.json
+++ b/python-dependencies.json
@@ -17,14 +17,14 @@
     "sources": [
         {
             "type": "file",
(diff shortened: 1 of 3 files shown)


Sjoerd1993 (approved):
Looks good, I already noticed that Matplotlib was a point-version behind. So this was on my list as well. I was not aware of the x-checker-data. [Seems very useful for this stuff](https://github.com/flathub/flatpak-external-data-checker).

PR #582, opened 2023-11-16
small improvements

cmkohnen:
- Allow editing bottom and left axis if no data is present
- Use generator expressions if better suited
- fix lining
- cleanup color deduplication

### DIFF
--- a/data/ui/figure_settings.blp
+++ b/data/ui/figure_settings.blp
@@ -59,14 +59,9 @@ template $GraphsFigureSettingsWindow : Adw.Window {
 
                 Adw.PreferencesGroup {
                   title: _("Axis Limits");
-                  Adw.ActionRow no_data_message {
-                    title: _("No Data");
-                    subtitle: _("Add data to the figure to set axis limits");
-                  }
                   Box {
                     orientation: vertical;
                     spacing: 12;
-                    visible: bind no_data_message.visible inverted;
                     Box left_limits {
                       orientation: horizontal;
                       spacing: 6;
--- a/src/data.py
+++ b/src/data.py
@@ -182,23 +182,12 @@ class Data(GObject.Object, Graphs.DataInterface):
         style_manager = self.get_application().get_figure_style_manager()
         selected_style = style_manager.get_selected_style_params()
         color_cycle = selected_style["axes.prop_cycle"].by_key()["color"]
-        used_colors = []
-
-        def _append_used_color(color):
-            used_colors.append(color)
-            # If we've got all colors once, remove those from used_colors so we
-            # can loop around
-            if set(used_colors) == set(color_cycle):
-                for color in color_cycle:
-                    used_colors.remove(color)
-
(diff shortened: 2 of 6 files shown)


data/ui/figure_settings.blp
```diff
@@ -59,14 +59,9 @@ template $GraphsFigureSettingsWindow : Adw.Window {
 
                 Adw.PreferencesGroup {
                   title: _("Axis Limits");
```
Sjoerd1993, line 61: Yeah this makes sense of course, the bottom and left axes are still there when no data is loaded. So it's a bit weird to not be able to edit them then.

src/data.py
```diff
@@ -182,23 +182,12 @@ def add_items(self, items: misc.ItemList) -> None:
         style_manager = self.get_application().get_figure_style_manager()
         selected_style = style_manager.get_selected_style_params()
         color_cycle = selected_style["axes.prop_cycle"].by_key()["color"]
-        used_colors = []
-
-        def _append_used_color(color):
-            used_colors.append(color)
-            # If we've got all colors once, remove those from used_colors so we
-            # can loop around
```
Sjoerd1993, line 190: Removing this breaks the current behaviour that it loops around. So if there's three colours in the settings: Red, Green, Blue. Then in the current main branch it goes like this when adding six items: Red, Green, Blue, Red, Green, Blue
But by removing this logic, it now goes like this: Red, Green, Blue, Red, Red, Red. 

That's a regression imo. Especially if you're adding a lot of data at once, this can become a bit weird with one color suddenly dominating.

Sjoerd1993 (changes requested):
See comment about colour looping.

src/data.py
cmkohnen: This is replaced by the logic on line 236, which is responsible for detecting that

cmkohnen: Nevermind, found the cause of the issue you were describing. Also fixed an issue, that was in main in regards to custom colors (when having a custom color present, the looping wouldn't work either)

Sjoerd1993:
The latest commit solves most issues, it works mostly correctly when adding indivual colours. But if all colours have been used once, and then some colours are removed, it doesn't always pick a colour that has only been used once. 

Bit of a weird description, see I made the following screencast, it should not have three lines with a blue color.
[Screencast from 2023-11-16 15-24-47.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/03ce7a98-5081-4770-8b3c-3c8e9b77c385)

See also [this branch](https://github.com/Sjoerd1993/Graphs/tree/Fix-single-color-additions), which is based on the first commit, which does solve these issues (but you beat me to it with last commit).

cmkohnen:
Fixed it, also fixed another issue related to cycling with twice the color_cycle present

Sjoerd1993 (approved):
Great, this seems to work flawless in my testing!

PR #585, opened 2023-11-16
drop handle-duplicates preference

cmkohnen:
also fixed an accidental regression

### DIFF
--- a/data/se.sjoerd.Graphs.gschema.xml
+++ b/data/se.sjoerd.Graphs.gschema.xml
@@ -5,13 +5,6 @@
     <value nick="Center at middle X value" value="1"/>
   </enum>
 
-  <enum id="se.sjoerd.Graphs.general.handle-duplicates-values">
-    <value nick="Auto-rename duplicates" value="0"/>
-    <value nick="Ignore duplicates" value="1"/>
-    <value nick="Add duplicates" value="2"/>
-    <value nick="Override existing items" value="3"/>
-  </enum>
-
   <enum id="se.sjoerd.Graphs.export-figure.file-formats">
     <value nick="Encapsulated Postscript" value="0"/>
     <value nick="Joint Photographic Experts Group" value="1"/>
@@ -63,9 +56,6 @@
   <schema id="se.sjoerd.Graphs.general">
     <key name="center" enum="se.sjoerd.Graphs.general.center-values">
       <default>"Center at middle X value"</default>
-    </key>
-    <key name="handle-duplicates" enum="se.sjoerd.Graphs.general.handle-duplicates-values">
-      <default>"Auto-rename duplicates"</default>
     </key>
     <key name="hide-unselected" type="b">
       <default>false</default>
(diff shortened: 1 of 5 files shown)


Sjoerd1993 (approved):
This works. This makes me wonder if we'd move/change the other preferences as well. In principle the "Hide selected items" could fit under the Appearance section of Figure Settings. 

The "Center action button" behaviour is a bit more difficult to move. It's not that pretty to have an entire preference dialog for a single option. I was thinking of perhaps adding some more action settings there. Specifically which smoothening algorithm to use (it would be nice maybe to support a [Savitzky-Golay](https://en.wikipedia.org/wiki/Savitzky%E2%80%93Golay_filter) filter for instance, which is readily available in SciPy), and for certain algorithms how coarse it should filter.

cmkohnen:
Ive had some ideas as well. I agree that hide unselected might be a better fit in figure_settings, as that also makes it project dependant.

For the action behaviour, I was thinking of using a splitbutton for the ui with the actual button performing the operation and the dropdown menu offering the choide for the wanted behaviour. This places it more contextually and is better for discoverability. Also avoids potential confusion even more.

Sjoerd1993:
> For the action behaviour, I was thinking of using a splitbutton for the ui with the actual button performing the operation and the dropdown menu offering the choide for the wanted behaviour

That's definitely a good thing to experiment with. The buttons are wide enough to support a dropdown arrow for the ones that warrant a setting. It would be overall a major UI improvement, as the presence of two distinct settings dialogs are honestly a bit confusing. (It has gotten better over time, but especially in my first design this was pretty bad, with "Default settings" in preferences and "Current settings" in Figure settings) 

I've toyed a bit with a Savgol filter, and it's incredibly easy to implement and leads to nicer results than the currently moving averaging that it does. The basic principle is that a savgol filter draws a polynomial over a certain window of the data, leading to smoother curves.

The entire smoothening then becomes like this:

```
def smoothen(_item, xdata, ydata):
    """Smoothen y-data."""
    new_ydata = scipy.signal.savgol_filter(ydata, min(len(xdata), 8), 3)
    return xdata, new_ydata, False, False
```

The `min(len(xdata), 8)` argument is the window of data points it takes to get the polynomial. This can never be smaller than the total amount of data points, hence this min construction. Then the `3` argument just means that it uses a third-order polynomial. I'm not sure if a dropdown allows for a clean way to set these arguments, but I think in the first implementation at least for the next release it would be fine with some sane defaults. 

One option at a later point is to have an additional "Advanced settings" button in the dropdown for smoothening. Where a small dialog pops up where such details can be set. But again, that might be something to think about later. I think in the short-term it would be nice if we'd move the "Center" setting to such a dropdown button.

Sjoerd1993:
Just played around with it quickly, putting this into the action buttons will work beautifully! Great idea.

image

The actual drop-down is of course a place-holder. But if I have some free time tonight, I can implement it this evening. Otherwise I can do it tomorrow if you haven't already started on this now:

image

cmkohnen:
Well just to add: I've already got it working:
Bildschirmfoto vom 2023-11-16 17-28-20

PR #586, opened 2023-11-16
remove preferences window

cmkohnen:
- Move hide_unselected to figure_settings (and the canvas)
- Make Center behaviour part of the main UI as discussed


Sjoerd1993 (approved):
Great!

PR #590, opened 2023-11-17
Implement smoothen options

Sjoerd1993:
This implements extra options for smoothening. This includes a new filter technique, the Savitzky–Golay filter which allows to do some pretty cool filtering. See example in the screencast, where I'm filtering a noisy sine wave. The original sine wave is has an amplitude of one, with some `np.random` noise added to it. As you can see, with a moving average you won't really get a smooth sine back, but the Savitzky–Golay draws a polynomial using defined window ranges and comes with this much nicer graph.

This all works now, but it probably needs some cleaning and it will not pass linting likely. I'm already running late though, so I wanted to put this up before I start on work.

[Screencast from 2023-11-17 08-35-11.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/f82b8fbc-8491-452a-8d7b-64db6561d1f0)

### DIFF
--- a/data/se.sjoerd.Graphs.gschema.xml
+++ b/data/se.sjoerd.Graphs.gschema.xml
@@ -60,6 +60,7 @@
   </schema>
 
   <schema id="se.sjoerd.Graphs.actions">
+    <child name="smoothen" schema="se.sjoerd.Graphs.actions.smoothen"/>
     <key name="center" enum="se.sjoerd.Graphs.actions.center-values">
       <default>"middle-x"</default>
     </key>
@@ -68,7 +69,7 @@
     </key>
   </schema>
 
-  <schema id="se.sjoerd.Graphs.smoothen">
+  <schema id="se.sjoerd.Graphs.actions.smoothen">
     <key name="savgol-window" type="i">
       <default>10</default>
     </key>
--- a/data/se.sjoerd.Graphs.gschema.xml
+++ b/data/se.sjoerd.Graphs.gschema.xml
@@ -51,7 +51,6 @@
 
   <schema id="se.sjoerd.Graphs" path="/se/sjoerd/Graphs/">
     <child name="actions" schema="se.sjoerd.Graphs.actions"/>
-    <child name="smoothen" schema="se.sjoerd.Graphs.smoothen"/>
     <child name="add-equation" schema="se.sjoerd.Graphs.add-equation"/>
     <child name="curve-fitting" schema="se.sjoerd.Graphs.curve-fitting"/>
     <child name="export-figure" schema="se.sjoerd.Graphs.export-figure"/>
--- a/data/se.sjoerd.Graphs.gschema.xml
+++ b/data/se.sjoerd.Graphs.gschema.xml
@@ -5,6 +5,11 @@
     <value nick="middle-x" value="1"/>
   </enum>
 
+  <enum id="se.sjoerd.Graphs.actions.smoothen-types">
+    <value nick="savgol" value="0"/>
+    <value nick="moving-average" value="1"/>
+  </enum>
+
   <enum id="se.sjoerd.Graphs.export-figure.file-formats">
     <value nick="Encapsulated Postscript" value="0"/>
     <value nick="Joint Photographic Experts Group" value="1"/>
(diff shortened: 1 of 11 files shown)


data/se.sjoerd.Graphs.gschema.xml
```diff
@@ -46,6 +51,7 @@
 
   <schema id="se.sjoerd.Graphs" path="/se/sjoerd/Graphs/">
     <child name="actions" schema="se.sjoerd.Graphs.actions"/>
+    <child name="smoothen" schema="se.sjoerd.Graphs.smoothen"/>
```
cmkohnen, line 54: This is redundant with it also being present in actions

src/actions.py
```diff
@@ -153,3 +165,38 @@ def delete_selected_action(_action, _target, self):
     names = ", ".join(item.get_name() for item in items)
     self.get_data().delete_items(items)
     self.get_window().add_toast_string(_("Deleted {}").format(names))
+
+
+@Gtk.Template(resource_path="/se/sjoerd/Graphs/ui/smoothen_settings.ui")
```
cmkohnen, line 170: This can probably moved to vala in a separate pr

src/actions.py
```diff
@@ -18,8 +20,14 @@ def perform_operation(_action, target, self):
...
-    if operation in ("center", ):
+    if operation in ("center", "smoothen"):
         args = [self.get_settings("actions").get_enum(operation)]
+    if operation == "smoothen":
+        params = {}
+        settings = self.get_settings("actions").get_child("smoothen")
+        for setting in settings:
+            params[setting] = int(settings.get_int(setting))
+        args += [params]
     if operation == "shift":
```
cmkohnen, line 31:
```suggestion
    elif operation == "shift":
```

src/application.py
```diff
@@ -102,6 +103,10 @@ def __init__(self, application_id, **kwargs):
         center_action = settings.get_child("actions").create_action("center")
```
cmkohnen, line 103: As more actions are probably implemented using this way, a loop structure would be useful here. Can be part of a separate pr however

cmkohnen (changes requested):

src/actions.py
Sjoerd1993: I moved most of it to Vala, but couldn't get the reset button to work. So that's still in the python code for now.

src/application.py
```diff
@@ -99,8 +100,9 @@ def __init__(self, application_id, **kwargs):
         operation_action.connect("activate", actions.perform_operation, self)
         self.add_action(operation_action)
 
-        center_action = settings.get_child("actions").create_action("center")
-        self.add_action(center_action)
+        for action_key in ["center", "smoothen"]:
+            action = settings.get_child("actions").create_action(action_key)
```
cmkohnen, line 104: I'd outline the get_child("actions") to be outside the loop

src/smoothen_settings.vala
```diff
@@ -0,0 +1,31 @@
...
+        [GtkChild]
+        public unowned Button reset_button { get; }
+
+        public SmoothenWindow (Application application) {
+            Object (
+                application: application,
+                transient_for: application.window
+            );
+            Tools.bind_settings_to_widgets (
+                application.settings.get_child("actions").get_child("smoothen"), this
```
cmkohnen, line 26:
```suggestion
                application.settings.get_child ("actions").get_child ("smoothen"), this
```

src/actions.py
```diff
@@ -123,6 +131,16 @@ def save_project_action(_action, _target, self):
     ui.save_project_dialog(self)
 
 
+def smoothen_settings_action(_action, _target, self):
+    def _on_reset(_button, self):
+        params = self.get_settings("actions").get_child("smoothen")
+        for key in params.list_keys():
+            params.reset(key)
```
cmkohnen, line 138: SInce this shares functionality with the import settings, perhaps there could be a way to move it to utilities? or even to vala, that way we can make direct use of builder callbacks

cmkohnen (changes requested):
some minor stuff still, otherwise this looks good

cmkohnen (approved):

PR #603, opened 2023-11-20
linguistic changes

sstendahl:
Very small PR

- Use title case in menu entries
- Use "curve" rather than item, which is probably more intuitive for most people

### DIFF
--- a/data/ui/figure_settings.blp
+++ b/data/ui/figure_settings.blp
@@ -175,7 +175,7 @@ template $GraphsFigureSettingsWindow : Adw.Window {
                   }
 
                   Adw.SwitchRow hide_unselected {
-                    title: _("Hide Unselected Curves");
+                    title: _("Hide Unselected Items");
                   }
 
                   Adw.ActionRow {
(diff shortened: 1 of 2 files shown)


cmkohnen:
The thing is, all items Curve, Text, Fill have to be considered in this wording. Curve kind of excludes them there

Sjoerd1993:
> The thing is, all items Curve, Text, Fill have to be considered in this wording. Curve kind of excludes them there

At the moment, the only user-facing thing that can be added to the canvas is the curve itself (I'd count the legend item as part of the curve here). But for future-facing additions it may be more general to talk about Items here. The change to curves was based on a comment from @bertob in the GNOME Circle feedback about the unclearity of the word "Items".

I think in the item submenu it's perfectly fine either way, as it's very clear that the options correspond to the item that had been clicked on. In the Figure Settings menu it's a bit more difficult. If it's still unclear with "Items" we can use a small subtitle of course, but I don't have any specific suggestion out of the top of my head without making it more confusing. Although I could argue that the move from Preferences to settings of this option does make it more clear anyway, as it's now at least somewhat obvious that this button corresponds to the canvas.

cmkohnen (approved):

PR #605, opened 2023-11-21
calculate luminance on preview generation

cmkohnen:
Currently the brightness of a style is calculated when the Figure settings are opened. This requires parsing the file and introduces percieved delay. This changes that functionality so that the luminance is calculated when the preview is generated.

### DIFF
--- a/src/styles.py
+++ b/src/styles.py
@@ -31,7 +31,7 @@ def _generate_filename(name: str) -> str:
 
 
 def _is_style_bright(params: RcParams):
-    return utilities.get_luminance(params["figure.facecolor"]) < 150
+    return utilities.get_luminance(params["axes.facecolor"]) < 150
 
 
 class StyleManager(GObject.Object, Graphs.StyleManagerInterface):
--- a/src/styles.py
+++ b/src/styles.py
@@ -30,6 +30,10 @@ def _generate_filename(name: str) -> str:
     return f"{name.lower().replace(' ', '-')}.mplstyle"
 
 
+def _is_style_bright(params: RcParams):
+    return utilities.get_luminance(params["figure.facecolor"]) < 150
+
+
 class StyleManager(GObject.Object, Graphs.StyleManagerInterface):
     __gtype_name__ = "GraphsStyleManager"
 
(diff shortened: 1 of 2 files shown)


Sjoerd1993 (approved):

Sjoerd1993:
Perhaps around line 213, it would be nice to use the same is_style_bright method there as well.

Sjoerd1993 (approved):
looks good to me

src/styles.py
```diff
@@ -199,12 +208,11 @@ def _on_style_change(self, override: bool = False) -> None:
         headerbar.provider = Gtk.CssProvider()
 
         # Set headerbar color and contrast
-        bg_color = self._selected_style_params["figure.facecolor"]
-        contrast = utilities.get_luminance(bg_color)
-        color = "@dark_5" if contrast > 150 else "@light_1"
+        color = "@light_1" if _is_style_bright(self._selected_style_params) \
```
Sjoerd1993, line 211:
```suggestion
        bg_color = self._selected_style_params["figure.facecolor"]
        color = "@light_1" if _is_style_bright(self._selected_style_params) \
```

Sloppy review from my part, the `bg_color` declaration was missing, that should be there

Sjoerd1993:
Noticed another issue. For the "Edit" pencil, we'd need to check against the brightness of the `axes.facecolor` instead of `figure.edgecolor`. So using the exact same function won't do.

cmkohnen:
reverted the change with using the function for the headerbar

src/styles.py
```diff
@@ -30,6 +30,10 @@ def _generate_filename(name: str) -> str:
     return f"{name.lower().replace(' ', '-')}.mplstyle"
 
 
+def _is_style_bright(params: RcParams):
+    return utilities.get_luminance(params["figure.facecolor"]) < 150
```
Sjoerd1993, line 34:
```suggestion
    return utilities.get_luminance(params["axes.facecolor"]) < 150
```

Should check for the axes background, not the figure background, as that is where the edit pencil is located.

Sjoerd1993 (changes requested):

Sjoerd1993 (approved):

PR #606, opened 2023-11-21
Improve drag and drop

Sjoerd1993:
Improves drag and drop visuals. Unfortunately I'm a bit stuck trying to get the blue frame around the widget during the entire dragging phase.

[Skärminspelning från 2023-11-21 21-24-54.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/7c88dc44-e904-43ab-94e2-6ad25badb77d)

### DIFF
--- a/src/item_box.py
+++ b/src/item_box.py
@@ -43,9 +43,6 @@ class ItemBox(Gtk.Box):
 
         self.item.connect("notify::color", self.on_color_change)
         self.on_color_change(self.props.item, None)
-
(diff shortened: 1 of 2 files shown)


Sjoerd1993:
The working examples I have on this implement the drag and drop behaviour on the Gtk.ListBox item instead of the individual boxes. Perhaps that should be the way to go to get the outline during the drag action working, even though it requires some rewriting..

PR #607, opened 2023-11-22
Shorten label using ellipsize

Sjoerd1993:
Uses ellipsize to shorten labels. This ensures a consistent label length ("A" and "1" don't have the same character length so capping the characters leads to different max lengths), and also makes sure that all available space is actually used.


cmkohnen:
see #608

cmkohnen (approved):

PR #608, opened 2023-11-22
remove proxy

cmkohnen:
The reason the ItemBox got the name property was to be able to shorten it. Using ellipsize instead makes this redundant and unnecessary.


Sjoerd1993:
Agreed, I thought about this but thought the property maybe was used elsewhere but didn't check.

PR #615, opened 2023-11-28
Open project files

sstendahl:
Allows to open project files directly from the file manager. 
Currently runs into an issue with the sandbox though, as the opened file cannot be read. (file.read(None) gives an error, granting access to the home directory solves but should be avoided). So need to work that out first.

Also need to try and get the icons to work. It registers the filetype correctly, but still uses a blank icon.

TODO:
- [x] Fix the portal so files can properly be read without breaking the sandbox
- [x] Add other filetypes in the mime than just the project
- [ ] Fix the icon for associated filetypes. (Perhaps this is actually OK as-is, not sure if GNOME actually does use the mime icon at all)

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -57,20 +57,12 @@ parts:
       - -usr/lib/python3/dist-packages/pip*
       - -usr/lib/python3/dist-packages/setuptools*
       - -usr/lib/python3/dist-packages/pkg_resources*
-      - -usr/lib/python3/dist-packages/setuptools*
     organize: 
       bin: usr/bin
       lib/python3.10/site-packages: usr/lib/python3/dist-packages
-
-  blueprint-compiler:
-    source: https://gitlab.gnome.org/jwestman/blueprint-compiler.git
-    source-tag: 'v0.10.0'
-    plugin: meson
-    meson-parameters:
-      - --prefix=/usr
-    override-prime: ''
+      
   graphs:
-    after: [ blueprint-compiler, numpy ]
+    after: [ numpy ]
     plugin: meson
     source: .
     meson-parameters:
@@ -79,8 +71,8 @@ parts:
     build-environment:
       - PYTHONPATH: $CRAFT_STAGE/usr/lib/python3/dist-packages:$PYTHONPATH
       - GI_TYPELIB_PATH: /snap/gnome-42-2204-sdk/current/usr/lib/$CRAFT_ARCH_TRIPLET/girepository-1.0:/snap/gnome-42-2204-sdk/current/usr/lib/girepository-1.0
-    override-pull: |
-      craftctl default
+    build-snaps:
+      - blueprint-compiler/latest/stable
     override-build: |
       craftctl default
       sed -e '1c#!/usr/bin/env python3' -i ${CRAFT_PART_INSTALL}/snap/graphs/current/usr/bin/graphs
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,15 +44,53 @@ jobs:
         architecture:
         - amd64
         - arm64
+      fail-fast: false
     steps:
-    - uses: actions/checkout@v3
-    - uses: docker/setup-qemu-action@v2
-      with:
-        image: tonistiigi/binfmt@sha256:df15403e06a03c2f461c1f7938b171fda34a5849eb63a70e2a2109ed5a778bde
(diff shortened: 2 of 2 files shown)


Sjoerd1993:
Two remaining issues:
- [ ] This breaks the Snap packages, I'm not sure why but I guess it has to do with the new MIME type. @soumyaDghosh do you ahve any idea how to implement this in Snap, or at least get it to build properly?
- [ ] I haven't figured out how to get a custom icon to work for the project types, but for now I think we can leave that as-is.

src/parse_file.py
```diff
@@ -27,9 +27,9 @@ def import_from_xrdml(self, file):
     scan_type = content.getElementsByTagName("scan")
     scan_axis = scan_type[0].attributes["scanAxis"].value
     if scan_axis.startswith("2Theta"):
-        scan_axis = _("2Theta")
+        scan_axis = "2Theta"
```
Sjoerd1993, line 30: These shouldn't be translated, as they are used later on in the code to parse the xml. Perhaps will add a translatable version specifically for the label later in a seperate PR.

src/data.py
```diff
@@ -424,7 +424,10 @@ def optimize_limits(self) -> None:
                 xydata = numpy.asarray(
                     item_.ydata if index % 2 else item_.xdata,
                 )
-                xydata = xydata[numpy.isfinite(xydata)]
+                try:
+                    xydata = xydata[numpy.isfinite(xydata)]
```
Sjoerd1993, line 428: This does not always work with certain exceptions, better to just skip optimizing limits in those cases.

soumyaDghosh:
> * This breaks the Snap packages, I'm not sure why but I guess it has to do with the new MIME type. @soumyaDghosh do you ahve any idea how to implement this in Snap, or at least get it to build properly?

Exactly, the patch is broken. I think, we can drop the patch now

Sjoerd1993:
Thanks for the quick reply. Just one question about the patch, it seems to override the icon definition, is that necessary? Would that break if we remove the patch?

soumyaDghosh:
> it seems to override the icon definition

As I said before, it's to make the symbolic icon of the gnome top bar properly work with snap. But, now gnome has dropped it too. 



> Would that break if we remove the patch?

Dropping that will break nothing.

src/application.py
```diff
@@ -106,6 +106,17 @@ def __init__(self, application_id, **kwargs):
             "items-ignored", ui.on_items_ignored, self,
         )
 
+    def do_open(self, files, nfiles, _hint):
+        self.do_activate()
+        if nfiles == 1 and files[0].get_uri().endswith(".graphs"):
```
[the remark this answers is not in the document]
cmkohnen: sorry let me rephrase. I think we shouldn't load a project dict, if the application is already open and contains data, as that overrides the data present. My suggestion is to add `and data.is_empty()` to the checks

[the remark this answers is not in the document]
Sjoerd1993: Ah right then I indeed misunderstood, I was mainly considering the case when the application wasn't open yet. 

But that case should definitely be handled. Perhaps it makes sense to ask with a pop-up in that case if there's data present, like we do already when opening a new project from the UI.

snap/snapcraft.yaml
```diff
@@ -81,7 +81,6 @@ parts:
       - GI_TYPELIB_PATH: /snap/gnome-42-2204-sdk/current/usr/lib/$CRAFT_ARCH_TRIPLET/girepository-1.0:/snap/gnome-42-2204-sdk/current/usr/lib/girepository-1.0
     override-pull: |
```
cmkohnen, line 82: Also, I believe this is now redundant

Sjoerd1993: I think you're correct.

data/se.sjoerd.Graphs.mime.xml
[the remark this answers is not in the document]
Sjoerd1993: As far as I understood, this file is mainly to register new types that haven't been registered before. `xrdml` is already recognized as XML document. The current registration in the .desktop file under `MimeType` makes Graphs a recommended program to open `xrdml` files. 

Perhaps the MimeType should be `application/xrdml` for `xrdml` by the way. As the recommended type of regular XML files is apparently `application/xml`. But changing that doesn't really seem to change anything.

cmkohnen (approved):

cmkohnen (approved):

Issue #486, opened 2023-09-15
Radian scaling crashes with large span


Sjoerd1993:
Graphs crashes on a radian scale if the axis spans a large amount of data. In its current state, it puts a tick at every integer in the data set. If data goes from 0 to 1e6, then it tries to put a million ticks in the graph and crashes. While this is not really intended usage, having a button that crashes the program when the wrong data is loaded is not ideal obviously.

Ideally the locator should be changed bit. I was thinking of changing it that if there will be fewer than 4 ticks on the graph, it halves the spacing between the grids until there's at least four ticks again. Similarly, if there's more than let's say 12 ticks, it doubles the spacing between the ticks until there's more than four ticks again. 

This would require a slightly customized locator for the radian axis, but isn't that difficult to implement.

PR #617, opened 2023-11-30
Fix radian scales

Sjoerd1993:
Closes issue #486 

Logic is a bit ugly still, but it works well and relatively efficient. Will clean this up before marking this as ready for review.
If there's fewer than four ticks, it halves the spacing between the ticks until there's at least four ticks again.

If there's more than 9 ticks. It will try with a spacing of 2 pi. If that's not enough, it will try with a spacing of 5 pi. Otherwise, it will increase that spacing with 5 pi until there's fewer than 9 ticks again.

The work to do is in making this a bit more clean. Also in the case of halving the amount of ticks, I'll do this with a calculation like for the case with more than 9 ticks:

```
        if num_ticks > 20:
            distance += (5*numpy.pi) * (int(num_ticks/50)) - numpy.pi
            num_ticks = int((vmax-vmin) / distance)
```

The above needs commenting, but works and is will tested. Similar logic can be applied for the halving logic.

### DIFF
--- a/src/scales.py
+++ b/src/scales.py
@@ -161,7 +161,14 @@ class CustomScaleLocator(ticker.MaxNLocator):
         return tick_pos
 
 class RadianLocator(ticker.MultipleLocator):
-    """Dynamically place tick positions on radian scale."""
+    """
+    Dynamically place tick positions on radian scale.
+    Places ticks at a distance of pi if there's between 4 and 12 ticks
+    Otherwise it places ticks at a distance of 2pi if reasonable, or with
+    a multiple of 5 pi such that a number between 3 and 9 ticks are placed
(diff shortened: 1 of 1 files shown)


Sjoerd1993:
That was basically a complete rewrite in the logic compared to the initial draft, but it's at the same time more readable and much more efficient. 

The while loops became computationally expensive if it needs to multiple the distance a lot of times to get to an appropriate amount of ticks (e.g. when having 1e7 on the left-hand scale). So calculating this directly using the ratio between the amount of ticks when using a distance of pi, and the desired amount of ticks, basically reduces that loop into a single calculation.

cmkohnen (approved):

PR #624, opened 2023-12-04
Add project titles

Sjoerd1993:
Adds a title to the content window. By default it uses "Unsaved Project" as title. Changes to title to the project name when saving a project, or when opening a project. Furthermore, asks for confirmation when closing the application, giving the alternative as well to save the project before closing the window.

There's some room for polish, but this is functional right now. Future additions may be to add an indicator if unsaved changes are present (so the dialog won't be necessary if everything is saved), and to allow the user to save the current project without needing a dialog. 

[Skärminspelning från 2023-12-04 16-44-49.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/44719ed2-bdac-4f83-9d22-11268fbe769f)


Sjoerd1993:
Converting this into a draft, as I'm implementing a tracker to for unsaved files as well. Current implementation is a bit clunky though and will probably be better off without a notifier.

Sjoerd1993:
It currently works quite well with the bullet point as an indicator if there's any unsaved changes, but this is still quite a bit from being ready. I'll probably add the following options to the menu:
- New Project (will just clean the slate and open a status that's equivalent ot a fresh start)
- Save Project... (will save the project, skipping the dialog if there's already a path assigned to the project)
- Save Project as...(will save the project, always giving a dialog)

I may add a subtitle as well giving the path of the project, like Gnome Text Editor has, but I'm very open to discussion about that.

Sjoerd1993:
Current working state. Most functionality works, but the code still needs a lot of work, will not finish this today but should be ready this week somewhere. Tried to mirror the behaviour from GNOME Text Editor.

Forgot to add in the video, but closing the window when there's no unsaved changes will not trigger a dialog, it will be triggered if there are any unsaved changes. Will also probably use the same dialog with a "Unsaved changes" title for all actions that will discard the open project. (Closing the Application, and "Open Project"). 

[Skärminspelning från 2023-12-05 11-21-34.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/c246878d-43eb-4bfe-887d-80c59d1b8568)

Sjoerd1993:
Also added a "New Project" option. Which now functionaly mostly works, noticed slight bugs. Most of the things on the TODO-list will be more related to actually writing proper code to implement this all. But again, likely not today.

src/actions.py
```diff
@@ -137,8 +137,62 @@ def export_figure_action(_action, _target, self):
     ExportFigureWindow(self)
 
 
+def new_project_action(_action, _target, self):
+    # Load default figure settings, close all data, reset clipboard
+    # Basical
+    def reset_project(self):
```
cmkohnen, line 143: this calls get_data() very often, I'd assign it to a local variable

src/data.py
```diff
@@ -86,6 +95,46 @@ def empty(self) -> bool:
...
+        return self._unsaved
+
+    @unsaved.setter
+    def unsaved(self, unsaved):
+        self._unsaved = unsaved
+        title = \
+            self.get_application().get_window().get_content_title().get_title()
+        if title.startswith("• "):
+            title = title.split("• ")[-1]
+
```
cmkohnen, line 110: same here, I'd assign content_title to a local variable

src/data.py
```diff
@@ -86,6 +95,46 @@ def empty(self) -> bool:
...
+                title)
+
+    @GObject.Property(type=str, default="")
+    def project_uri(self) -> str:
+        """Whether or not the class is empty."""
+        return self._project_uri
+
+    @project_uri.setter
+    def project_uri(self, project_uri):
+        self._project_uri = project_uri
```
cmkohnen, line 136: This should be replaced using a simple definition:
`project_uri = GObject.Property(type=str, default="")`

cmkohnen (commented):
Though still WIP, some comments from my part

Sjoerd1993:
This is mostly ready, but I need to do some testing and have a second look at the code. 
I'm not 100% sure if it's worth it to remove the subtitle on a blank state, only to add "Draft" immediately with the very first change in the project. But I've basically just been mimicking that behaviour from GNOME Text Editor.

src/operations.py
```diff
@@ -239,10 +239,14 @@ def shift(item, xdata, ydata, left_scale, right_scale, items, ranges):
         previous_item = data_list[index - 1]
 
         # Only use selected span when obtaining values to determine shift value
-        new_xdata, new_ydata = filter_data(
-            previous_item.xdata, previous_item.ydata, ">=", min(xdata))
-        new_xdata, new_ydata = filter_data(
-            new_xdata, new_ydata, "<=", max(xdata))
+        new_xdata, new_ydata = xdata, ydata
```
Sjoerd1993, line 242: This should perhaps be a seperate PR, but this put is just a bug fix. Basically currently the shift doesn't work properly if a selection is used, and one of the curves is falls outside of the selected range while the other does not. This is due to this filter. So this filter is skipped now if that is the case. That fixes that issue

src/file_io.py
```diff
@@ -107,6 +109,18 @@ def save_item(file, item_):
             wrapper.write(fmt % values + "\n")
 
 
+def save_project(self, require_dialog=False, close=False):
+    if self.get_data().project_uri != "" and not require_dialog:
+        file_uri = self.get_data().project_uri
+        file = Gio.File.new_for_uri(file_uri)
+        write_json(file, self.get_data().to_project_dict(), False)
+        self.get_data().props.unsaved = False
+        if close:
```
Sjoerd1993, line 118: This is not super pretty. The issue is that Python doesn't wait until a response before executing the next line. So in the following dialog:

```
        if self.get_data().props.unsaved:
            def on_response(_dialog, response):
                if response == "discard_close":
                    self.quit()
                if response == "save_close":
                    file_io.save_project(self, close=True)

            dialog = ui.build_dialog("save_changes")
            dialog.set_transient_for(self.get_window())
            dialog.connect("response", on_response)
            dialog.present()
```
It would probably be nicest to handle the "close"  signal directly after the `if response == "save_close":` line. But if that is done, then the save dialog is shown, and immediately the application will be closed before the user even has the option to choose something. So either way, one would need to wait for a response.

I've tried to look into some solutions. I looked a bit after trying to emit a specific signal, but that didn't become much prettier. Ideally, we would maybe just halt everything when a GtkFiledialog is shown, and wait until that is resolved. But I am not 100% if that's possible, or compatible with modern GObject logic.

Either way, the way I do it here by handling this on the response signal works. The idea is that this may be cleaned in a future PR though. As I'd probably prefer to handle the close signal at the close dialog.

data/ui/window.blp
```diff
@@ -619,8 +628,12 @@ template $GraphsWindow: Adw.ApplicationWindow {
 }
 
 menu primary_menu {
+  section{
+    item (_("New Project"), "app.new_project")
+  }
   section{
     item (_("Save Project…"), "app.save_project")
+    item (_("Save Project as…"), "app.save_project_as")
     item (_("Open Project…"), "app.open_project")
```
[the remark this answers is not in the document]
cmkohnen: I'd its fine as is

cmkohnen (changes requested):

src/data.py
```diff
@@ -502,25 +564,42 @@ def to_project_dict(self) -> dict[str, Any]:
             "history-position": self._history_pos,
             "view-history-states": self._view_history_states,
             "view-history-position": self._view_history_pos,
+            "project-uri": self.get_project_uri(),
```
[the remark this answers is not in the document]
Sjoerd1993: Agreed, this is also completely redundant as the uri is never retrieved from the project file. Furthermore, `get_project_uri()` won't work anyway. It's the wrong syntax.

src/data.py
```diff
@@ -55,6 +59,7 @@ class Data(GObject.Object, Graphs.DataInterface):
     can_redo = GObject.Property(type=bool, default=False)
     can_view_back = GObject.Property(type=bool, default=False)
     can_view_forward = GObject.Property(type=bool, default=False)
+    project_uri = GObject.Property(type=str, default="")
```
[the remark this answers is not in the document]
Sjoerd1993: The `GFile.get_project_uri()` method returns a regular `str` type. But there's no need to save this as GObject Property actually in this case...

src/data.py
```diff
@@ -64,13 +69,24 @@ def __init__(self, application, settings):
...
+        self.initialize()
+
+    def initialize(self):
+        """Initializes the data class and sets the default values"""
+        figure_settings = self.get_figure_settings()
         limits = figure_settings.get_limits()
+        self.props.can_undo = False
+        self.props.can_redo = False
+        self.props.can_view_back = False
+        self.props.can_view_forward = False
```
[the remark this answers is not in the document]
cmkohnen: oh, thats fine then

Sjoerd1993:
The issue I was facing is that the response signal after finishing a save dialog is defined on `save_project()`, and then the next line of code is executed. Thus placing a `self.quit()` or a load project method after saving doesn't work, unless you use solutions like conditional statements in that response signal.

I've gotten around this by emitting a new `GSignal` on saving a project instead, and connecting a handler to that signal in the cases when action is required *after* saving the project (mostly save and close, and save and open). There may be some refactors imaginable in the future to make that prettier, but it works well now at least without any new bugs I could find.

PR #637, opened 2023-12-08
fix focus ring

cmkohnen:
Alternative to #635 
Have been working on this for a few days now but never got it to work correctly until now.

Makes the focus-ring appear on the picture only

Bildschirmfoto vom 2023-12-08 15-20-29


Sjoerd1993 (approved):
Significantly nicer than #635,

PR #638, opened 2023-12-08
build system changes

cmkohnen:
I haven't found a way to generate previews reliably during build, but this is the first step towards that

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -147,9 +147,7 @@ gresource_conf.set('STYLES', style_conf)
 gresource_conf.set('PREVIEWS', preview_conf)
 ui_conf = ''
 foreach b : blueprints_ui
-  if b != 'ui/help_overlay.ui'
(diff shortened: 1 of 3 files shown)


Sjoerd1993:
```
FAILED: data/se.sjoerd.Graphs.gresource 
/usr/bin/glib-compile-resources data/se.sjoerd.Graphs.gresource.xml --sourcedir /run/build/graphs/_flatpak_build/data --sourcedir data --sourcedir ../data --c-name graphs --internal --generate --target data/se.sjoerd.Graphs.gresource --dependency-file data/se.sjoerd.Graphs.gresource.d
data/se.sjoerd.Graphs.gresource.xml: Failed to locate “ui/help_overlay.ui” in any source directory.
```
In general it wouldn't be too terrible by the way if the previews are generated upon boot-time by the way, as long as it's during the initial boot and they're cached afterwards.

PR #640, opened 2023-12-09
Add matplotlib style

Sjoerd1993:
A relatively simple change, but I just noticed we are shipping pretty much every style that comes bundled with standard matplotlib, except for the default matplotlib style. This adds the default matplotlib style.

image


data/styles/matplotlib.mplstyle
cmkohnen, line 1: Might it be possible to sort this like the other shipped styles?

cmkohnen (approved):

PR #642, opened 2023-12-11
Curve fitting redesign

Sjoerd1993:
Redesigns Curve Fitting dialog according to [GNOME Circle Review](https://gitlab.gnome.org/Teams/Circle/-/issues/185):

[Skärminspelning från 2023-12-11 14-40-07.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/05a5d860-f3da-47b5-a96f-de51a1ebf9b5)

- [x]  The overall layout is awkward, I'd try just doing a split view, i.e. the Cancel button would go away (in favor of a normal close button) and the Apply button would probably be a pill button centered at the bottom of the right panel
- [x]  When switching between dark and light while the curve fitting dialog is open, the preview doesn't recolor until you close/reopen the dialog
- [x]  The empty box at the bottom of the sidebar is still weird, I'd hide it when it's empty
- [x]  The sidebar layout with fixed top and bottom elements and a scrollview in the middle is awkward. Is there a reason why we can't just put everything in a single scrollview?
- [x]  A few of the labels ("Fitting parameters", "Initial guess", etc.) don't use Title Case
- [x]  The entries have a darker hover state than normal rows/entries

The only point where I deviated is the ` The empty box at the bottom of the sidebar is still weird, I'd hide it when it's empty` comment. The results card there can only be empty when one of the values was wrong during initialization (e.g. when starting the curve fit while an invalid boundary or equation was set). Instead of hiding it, I put an error message in there instead.

### DIFF
--- a/data/ui/curve_fitting.blp
+++ b/data/ui/curve_fitting.blp
@@ -7,6 +7,9 @@ template $GraphsCurveFittingTool : Adw.Window {
   focus-widget: confirm_button;
   default-width: 1000;
   default-height: 650;
+  width-request: 305;
+  height-request: 410;
+
   ShortcutController {
     Shortcut {
       trigger: "Escape";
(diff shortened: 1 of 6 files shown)


cmkohnen:
Should the add button receive the suggestive action class?

Sjoerd1993:
> Should the add button receive the suggestive action class?

I'm honestly unsure, but it does make some sense to me to do so.

cmkohnen (approved):

PR #643, opened 2023-12-12
Style fixes

Sjoerd1993:
Fixes some issues with the styles:

- Stops connecting style preview buttons to `window.edit_style` after the initial time. This prevents multiple calls to the same functions which lead to performance issues, and "critical" errors in the terminal output about the view already being present in the stack.

- Reorders a bit with regard to the behaviour on style changes/edits. Allows to edit the current style, and then keep this style selecting (and applying changes) when backing out.

- New styles now use the currently active style as template by default, instead of the top style.

I've also looked into why the scroll window has such an incredibly long list of whitespace. The cause of this is that it reserves whitespace for when the window is resized to fewer columns. So the reserved amount of space is equal to the amount of space that is needed for the case when there's only one column present per row. Not completely sure if it's fixable.

### DIFF
--- a/src/styles.py
+++ b/src/styles.py
@@ -33,6 +33,9 @@ def _is_style_bright(params: RcParams):
 
 class StyleManager(GObject.Object, Graphs.StyleManagerInterface):
     __gtype_name__ = "GraphsStyleManager"
+    __gsignals__ = {
+        "add-style": (GObject.SIGNAL_RUN_FIRST, None, (str,)),
+    }
 
     application = GObject.Property(type=Graphs.Application)
     use_custom_style = GObject.Property(type=bool, default=False)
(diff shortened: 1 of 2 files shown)


src/styles.py
```diff
@@ -605,6 +614,19 @@ def add_color(self, _button):
...
+        for index, style in enumerate(style_model):
+            if style is not None:
+                file2 = style.get_file()
+                if file2 is not None and file.equal(file2):
+                    stylename = style.get_name()
+                    style_model.remove(index)
+                    style_manager._stylenames.remove(stylename)
+                    style_manager.props.use_custom_style = False
+                    style_manager.props.custom_style = \
+                        style_manager._system_style_name
```
cmkohnen, line 629: This logic is making a few calls, that should rather be in the stylemanager itself imo.

cmkohnen (approved):

PR #646, opened 2023-12-13
Hide unused axes settings

Sjoerd1993:
Hide all settings for each axis if they're currently not used.

image

image

Second screenshot is slightly extended compared to default height to show all relevant fields.

### DIFF
--- a/src/canvas.py
+++ b/src/canvas.py
@@ -110,6 +110,7 @@ class Canvas(FigureCanvas, Graphs.CanvasInterface):
 
         self.highlight = _Highlight(self)
         self._legend = True
+        self._visible_axes = (False, False, False, False)
         self._legend_position = misc.LEGEND_POSITIONS[0]
         self._handles = []
         if interactive:
@@ -274,10 +275,18 @@ class Canvas(FigureCanvas, Graphs.CanvasInterface):
             self._renderer.dpi = self.figure.dpi
             self.figure.draw(self._renderer)
 
+    @property
+    def visible_axes(self):
+        return self._visible_axes
+
+    @visible_axes.setter
+    def visible_axes(self, visible_axes):
+        self._visible_axes = visible_axes
+
     def _redraw(self, *_args):
         # bottom, top, left, right
-        visible_axes = [False, False, False, False]
         used_axes = [False, False, False, False]
+        visible_axes = [False, False, False, False]
         if self.props.hide_unselected:
             drawable_items = []
             for item in self.props.items:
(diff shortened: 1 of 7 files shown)


src/canvas.py
```diff
@@ -294,14 +303,16 @@ def _redraw(self, *_args):
             visible_axes[xposition] = True
             visible_axes[2 + yposition] = True
             used_axes[xposition + 2 * yposition] = True
+        self.visible_axes = visible_axes
```
cmkohnen, line 306: id move this assignment after the not any() check

src/canvas.py
```diff
@@ -274,10 +275,18 @@ def on_draw_event(self, _widget, ctx):
...
             self.figure.draw(self._renderer)
 
+    @property
+    def visible_axes(self):
+        return self._visible_axes
+
+    @visible_axes.setter
+    def visible_axes(self, visible_axes):
+        self._visible_axes = visible_axes
+
```
cmkohnen, line 285: This is redundant. I'd make this either a GProperty or refer to the next comment.

src/figure_settings.py
```diff
@@ -130,13 +129,9 @@ def on_select(self, model, _pos, _n_items):
...
     def set_axes_entries(self):
-        used_axes = [False, False, False, False]
-        for item in self.get_application().get_data():
-            for i in item.get_xposition() * 2, 1 + item.get_yposition() * 2:
-                used_axes[i] = True
-        if not any(used_axes):
-            used_axes = [True, True, False, False]
-        for (direction, visible) in zip(_DIRECTIONS, used_axes):
+        canvas = self.get_application().get_window().get_canvas()
+        visible_axes = canvas.visible_axes
```
cmkohnen, line 133: The issue I have with this it that it ties the figure settings to the canvas, when the application is structured the other way around. I'd keep the implementation here

src/canvas.py
Sjoerd1993: Ueaj, a GProperty would be better here.

src/figure_settings.py
Sjoerd1993: The reason for doing it this way was mainly because these theoretically should give the same results, and the old implementation thus kinda is a repetition of the logic in the canvas.

But I am fine with keeping the earlier implementation still. Or perhaps reimplement the logic as a property of the Data would make more sense. That would make it a bit more agnostic to the specific canvas or the application itself.

cmkohnen (approved):

PR #647, opened 2023-12-13
Remove zoom mode

cmkohnen:
Since we have finer control over zooming now, this mode can be dropped imo.


cmkohnen:
Well personally I never really used this zoom mode. But if it is still useful, as you describe, I'm not at all against keeping it.

PR #650, opened 2023-12-14
Fix style overview columns to three

Sjoerd1993:
It's a bit of a bandaid solution, as it'd still be preferable to allow resizing to fewer columns. But GtkGridView reserves the vertical space it needs for the case when there's only one column available. This leads to a large amount of whitespace in the scrollbar in the style overview.

We can maybe take a closer look at this later, I couldn't find a direct solution for now. But I think for now at least for the coming release it's sensible to just fix the amount of columns to three, which solves this issue:

image

### DIFF
--- a/data/ui/figure_settings.blp
+++ b/data/ui/figure_settings.blp
@@ -230,6 +230,7 @@ template $GraphsFigureSettingsWindow : Adw.Window {
               GridView grid_view {
                 name: "style-grid";
                 max-columns: 3;
+                min-columns: 3;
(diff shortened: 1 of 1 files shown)


cmkohnen (approved):

cmkohnen:
Maybe we can use a Breakpoint to set the limit to 2 or 1 if the size is too small. However this circumvents an upstream Bug in my opinion. Maybe we should consider filing an upstream issue on this.

Issue #494, opened 2023-09-15
Add curve-fitting predefined equations


Sjoerd1993:
It would be nice to have some predefined equations in the curve fitting tool. Main question would be how  to implement this nicely in the UI. Having a dropdown above the current entry seems logical, but it would take some space which already is somewhat sparce.

This could be moved into a menu as well in the top-right (similar to the main application), where the equation type can be toggled (similar to the axis scaling in the view menu). The equation entry can be set to inactive (so one cannot edit the entry) when a predefined equation is chosen, and the text then can be set to the equation name and type. (e.g. "Linear: Y = a*x+b" or "Quadratic, Y = a*x² + b*x +c").

But again, I'm not yet 100% settled to where to show this. But a menu would be probably be added anyway likely, to set other settings as well.

Options to include at least:

- Linear
- Quadratic
- Cubic
- Exponential
- Gaussian
- Lorentzian (At least if we are adding Gaussian anyway, it makes sense to also have lorentzian, but I am ok with skipping this one to avoid having to many choices)
- Custom equation

PR #656, opened 2023-12-19
Add curve fitting options

Sjoerd1993:
Adds the following options to the Curve Fitting dialog:

- **Equation**: Sets the equation that is currently used. I've set the default value to "Custom" for discoverability's sake, the idea there was that out of the box all entries are working and editable. Then when one discovers the setting, it is more obvious why this box is grayed out. The reason for not removing this box, is so that it's clear what the equation itself is that is being fitted to.
- **Optimization**: Sets the optimization algorithm that is being used for the fit. Typical Levenberg-Marquardt is the standard for most tools I've come across. But there may be reasons for prefering others.
- **Confidence bounds**: This sets the confidence bounds. Allowing this to turn them off, or to just set it to a multiple of the standard deviation. Which works well enough as a set of options for now at least.

Compared to the video below, I've removed the "Toggle Sidebar" option, as it seemed a bit unintuitive to actually use that in the dialog.

[Skärminspelning från 2023-12-19 09-58-44.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/db8bd2ef-6d9c-477d-9c08-098298db3a14)

Closes issue #494

### DIFF
--- a/data/se.sjoerd.Graphs.gschema.xml
+++ b/data/se.sjoerd.Graphs.gschema.xml
@@ -107,7 +107,7 @@
       <default>"linear"</default>
     </key>
     <key name="optimization" enum="se.sjoerd.Graphs.curve-fitting.optimization">
-      <default>"lm"</default>
+      <default>"trf"</default>
     </key>
     <key name="confidence" enum="se.sjoerd.Graphs.curve-fitting.confidence">
       <default>"1std"</default>
--- a/data/se.sjoerd.Graphs.gschema.xml
+++ b/data/se.sjoerd.Graphs.gschema.xml
@@ -10,6 +10,30 @@
     <value nick="moving-average" value="1"/>
   </enum>
 
+  <enum id="se.sjoerd.Graphs.curve-fitting.equation">
+    <value nick="linear" value="0"/>
+    <value nick="quadratic" value="1"/>
+    <value nick="exponential" value="2"/>
+    <value nick="power" value="3"/>
+    <value nick="log" value="4"/>
+    <value nick="sigmoid" value="5"/>
+    <value nick="gaussian" value="6"/>
+    <value nick="custom" value="7"/>
+  </enum>
+
+  <enum id="se.sjoerd.Graphs.curve-fitting.confidence">
+    <value nick="none" value="0"/>
+    <value nick="1std" value="1"/>
+    <value nick="2std" value="2"/>
+    <value nick="3std" value="3"/>
+  </enum>
+
+  <enum id="se.sjoerd.Graphs.curve-fitting.optimization">
+    <value nick="lm" value="0"/>
+    <value nick="trf" value="1"/>
+    <value nick="dogbox" value="2"/>
+  </enum>
+
   <enum id="se.sjoerd.Graphs.export-figure.file-formats">
     <value nick="Encapsulated Postscript" value="0"/>
     <value nick="Joint Photographic Experts Group" value="1"/>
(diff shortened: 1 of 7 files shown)


Sjoerd1993:
I think you're right. It's also more discoverable, and it gives an immediate look at the current setting. I was just slightly unsure as it would make the sidebar a bit taller by default, but that may be worth it.

Sjoerd1993:
Have to think about how to do it nicely UI-wise. Just using an entry row will look like this. Perhaps it may be good to have it in a seperate group so there's some space in between, although it does make sense to couple them:

image

Sjoerd1993:
So this now uses a comborow, where the subtitle is used to show the equation when a preset value is chosen.

[Skärminspelning från 2023-12-19 20-14-14.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/8f15a21d-8941-42b3-acc6-fb078a3967ca)

Sjoerd1993:
So right now if I select an equation from the Comborow, that all works fine. But if I then close the dialog, and select an equation again, I get a crash. It seems to trigger the `on_select` signal twice for some reason, and never runs `set_equation`. I don't know why at this moment.

Sjoerd1993:
Okay, the fix to the above was simple. The reason I was connecting this to an action using Vala in the first place was because the `@Gtk.Callback()` were not recognized, it seems to only look in the Vala scope, likely because that's where the Gtk template is set from the ui file.  However, instead of using callbacks, simply connecting them is a simple workaround, and solves the above issue with the crash that I described earlier. So now it all works as expected.

Sjoerd1993:
See attached video. I've also added an "Y=" label in front of the equations. While a lowercase letter looks somewhat nicer in this particular situation imo, I used an uppercase for consistency's sake with the rest of the program.

It would perhaps be cleaner to use `f(x)=`, especially now there's more variables. But again I kept it as `Y=` for consistency's sake. We could change the Add Equation dialog to `f(x)` as well. But I was a bit unsure if that needed changing at all.

Sjoerd1993:
My final change to this PR, I've also changed the wording for the standard deviations, so that it now also provides the confidence interval according to the [68-95-99.7 rule](https://en.wikipedia.org/wiki/68%E2%80%9395%E2%80%9399.7_rule). I think the certainty is a measure that many are interested in, and if I recall correctly is the measure that Matlab also shows rather than the standard deviation. I did have to shorten "Standard Deviation"  to SD though. Which is common, but maybe slightly too jargon for some...? Although people who won't recognize that won't really recognize "Standard deviation" either, and are more helped with the certainty wording was my reasoning there.

image

Sjoerd1993:
On the latest changes, here's SciPy about the optimization methods:

```
method{‘trf’, ‘dogbox’, ‘lm’}, optional

    Algorithm to perform minimization.

            ‘trf’ : Trust Region Reflective algorithm, particularly suitable for large sparse problems with bounds. Generally robust method.

            ‘dogbox’ : dogleg algorithm with rectangular trust regions, typical use case is small problems with bounds. Not recommended for problems with rank-deficient Jacobian.

            ‘lm’ : Levenberg-Marquardt algorithm as implemented in MINPACK. Doesn’t handle bounds and sparse Jacobians. Usually the most efficient method for small unconstrained problems.

```

So Levenberg-Marquardt doesn't handle bounds, hence hiding them when selected. And in general `trf` seems by far the most obvious choice as default option.

cmkohnen (approved):

PR #659, opened 2023-12-21
hide dpi with vector based images

Sjoerd1993:
Hides the dpi when exporting images in vector-based formats. It's meaningless to talk about pixel density in vector images, and this setting doesn't even have an impact on the exported figure (not even on the rendered size), so it's best to hide them when not applicable.

Also removes unsupported filetypes, those were broken from the beginning.

### DIFF
--- a/src/export_figure.py
+++ b/src/export_figure.py
@@ -31,8 +31,7 @@ class ExportFigureWindow(Adw.Window):
             Gtk.StringList.new(list(self.file_formats.keys())))
         ui.bind_values_to_settings(
             self.get_application().get_settings_child("export-figure"), self)
-        self.dpi.set_visible(self.file_format.get_selected()
-                             not in [0, 2, 4, 5])
+        self.on_file_format(None, None)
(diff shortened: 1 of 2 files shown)


src/export_figure.py
```diff
@@ -19,16 +19,27 @@ class ExportFigureWindow(Adw.Window):
...
             application=application, transient_for=application.get_window(),
-            file_formats=self._canvas.get_supported_filetypes_grouped(),
+            file_formats=valid_formats,
         )
         self.file_format.set_model(
             Gtk.StringList.new(list(self.file_formats.keys())))
         ui.bind_values_to_settings(
             self.get_application().get_settings_child("export-figure"), self)
+        self.dpi.set_visible(self.file_format.get_selected()
+                             not in [0, 2, 4, 5])
```
cmkohnen, line 35:
```suggestion
        self.on_file_format(None, None)
```

cmkohnen (approved):

PR #661, opened 2023-12-21
Minor string changes

Sjoerd1993:
- Removes the "hand" suffix for the direction in the axes limits so it's consistent with the rest of the application. (Could use it everywhere instead as well, but I guess removing it is somewhat cleaner)
- Fixes a capitalization in the subtitle of the text and label size, in the style editor.
- Removes the X and Y prefix when talking about e.g. "Bottom X axis", as the X/Y part is already implicitly specified when giving the direction.
- Change "Best" in the legend to "Auto", as the former is a bit confusing imo.


cmkohnen (approved):
Regarding my comment on the last issue, I haven't read this pr yet.

PR #662, opened 2023-12-21
Generate system style previews

### LINKED ISSUE #632: Add system theme preview
The system theme preview is not implemented yet, this should be considered a clear blocker for the next release:

image

Just as a side-note, there's not really any underlying motive to putting this up as a separate issue. I was just gathering open release blockers in the issue trackers. Once this issue and #494 are closed, I think we can put up an announcement for the beta itself on This Week in GNOME perhaps.

Sjoerd1993:
Adds a system style preview. This is currently done on boot-time, boot I am not 100% sure if it's trivial to move this to build time. Not 100% sure how that will be handled with updates if/when we generate these on build-time. Will an update trigger a new build? (Not 100% sure). An alternative could then be to generate these as a GitHub action, and then bundle the previews in the `data`  folder of the repository somewhere.

But for now I think it's okay as-is. I've timed it with the Python time module, and generating the regular previews takes a total of 1.6 seconds, while generating the system preview takes a total of 1.8 seconds. So it's not negligible, but it's OK for now. Note that this will also correctly show a Yaru style preview for the Ubuntu snap build.

Also uses a png for the system theme (because you can't really cut vector images, you can just mask them), which makes things a bit easier. Can't tell I see any affects of it being a raster image though.

image

Closes #632

### DIFF
--- a/src/style_io.py
+++ b/src/style_io.py
@@ -12,7 +12,6 @@ from gi.repository import Gio
 from graphs import file_io, utilities
 
 from matplotlib import RcParams, cbook, rc_context
-from matplotlib.backends.backend_gtk4cairo import FigureCanvas
 from matplotlib.figure import Figure
 from matplotlib.font_manager import font_scalings, weight_dict
 from matplotlib.style.core import STYLE_BLACKLIST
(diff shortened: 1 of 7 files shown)


src/style_io.py
```diff
@@ -133,3 +137,69 @@ def generate_preview(style: RcParams) -> Gio.File:
...
+        figure = Figure(figsize=(5, 3))
+        axis = figure.add_subplot()
+        axis.spines.bottom.set_visible(True)
+        axis.spines.left.set_visible(True)
+        if not light_style["axes.spines.top"]:
+            axis.tick_params(which="both", top=False, right=False)
+        axis.plot(_PREVIEW_XDATA, _PREVIEW_YDATA1)
+        axis.plot(_PREVIEW_XDATA, _PREVIEW_YDATA2)
+        axis.set_xlabel(_("X Label"))
+        axis.set_ylabel(_("Y Label"))
```
cmkohnen, line 155: Given how much duplication is here, I'd outline the actual generation to an 'internal' function, which takes both the style as rcParams, a file-like object to write to, and the actual save mode.

src/style_io.py
```diff
@@ -133,3 +137,69 @@ def generate_preview(style: RcParams) -> Gio.File:
...
+        canvas = FigureCanvas(figure)
```
cmkohnen, line 156: Is creating a canvas really needed here? shouldn't `figure.savefig()` work as well? (afaik the canvas implicitly calls this either way)

src/styles.py
```diff
@@ -65,9 +65,20 @@ def __init__(self, application: Graphs.Application):
                 _compare_styles,
             )
         enumerator.close(None)
-        # TODO: add System style preview
-        self._system_style = Graphs.Style.new(_("System"), None, None, False)
         self._update_system_style()
+        system_style = self._system_style_name
```
cmkohnen, line 69: Given this is accessed twice assigning it locally is a bit redundant here. I'd just use `self._system_style_name`

src/styles.py
```diff
@@ -65,9 +65,20 @@ def __init__(self, application: Graphs.Application):
...
+            "resource:///se/sjoerd/Graphs/styles/" + light_file,
+        ))[0]
+        dark_file = _generate_filename(system_style + " Dark")
+        dark_style = style_io.parse(Gio.File.new_for_uri(
+            "resource:///se/sjoerd/Graphs/styles/" + dark_file,
+        ))[0]
+
+        preview = style_io.generate_system_preview(light_style, dark_style)
+        self._system_style = \
+            Graphs.Style.new(_("System"), None, preview, False)
```
cmkohnen, line 81:
```suggestion
        system_preview = style_io.generate_system_preview(light_style, dark_style)
        self._system_style = \
            Graphs.Style.new(_("System"), None, system_preview, False)
```

src/style_io.py
Sjoerd1993: I've did some deduplication here. It can probably be deduplicated a bit more, but that needs a bit more restructuring, which I probably won't have time for until at least after christmas.

Sjoerd1993:
Have been looking a bit into de-duplicating this a bit more. But merging the system style preview with the regular theme previews will be clunky enough that I'm personally happy with the current state here now.

cmkohnen (approved):

PR #669, opened 2023-12-30
Use dynamic axes strings

Sjoerd1993:
Dynamically changes the strings for the axes. So by default the X and Y axes are called "X Axis" and "Y Axis" respecitvely. So the scale for instance is called "X Axis Scale" and "Y Axis Scale".

However, if two opposite axes are used simultaniously, then they get a directional prefix to distinguish them. So "Left Y Axis Scale"  and "Right Y Axis Scale".

I realize this is not super concise everywhere. Amount of lines could be greatly reduced by creating the labels in a loop, however this will lead to unexpected problems with localization, since the order of words is not the same in every language. E.g. "Left Y Axis Scale" in French would be equivalent to "Scale of the Y Axis Left" if I remember my French lessons correctly. It's difficult to predict how things will conjugate gramatically, so I thought it's important to keep the entire string intact. Hence using this thick dictionary in `figure_settings`.

Secondly, when the scale is now changed for an axis, it also changes the scale accordingly for the opposite axis if it's not in use. This because it can be unexpected behaviour that your scale is different if you just change the axis direction. See screencast:
[Skärminspelning från 2023-12-30 07-51-38.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/07068e27-3977-4b46-83d4-4b6936e07353)


src/figure_settings.py
```diff
@@ -128,9 +142,58 @@ def on_select(self, model, _pos, _n_items):
...
+                "min": _("Right Y Axis Minimum") if visible_axes[2]
+                and visible_axes[3] else _("Y Axis Minimum"),
+                "max": _("Right Y Axis Maximum") if visible_axes[2]
+                and visible_axes[3] else _("Y Axis Maximum"),
+                "scale": _("Right Y Axis Scale") if visible_axes[2]
+                and visible_axes[3] else _("Y Axis Scale"),
+                "label": _("Right Y Axis Label") if visible_axes[2]
+                and visible_axes[3] else _("Y Axis Label"),
+            },
+        }
```
cmkohnen, line 195: There most definitely has to be a way to make this work as a set comprehension, as this is quite a bit of duplicated logic

Sjoerd1993: Yeah the thing is that the following strings should all be defined as a whole and not be broken up to avoid issues with translations. Not all languages has the adjectives in front of the word for example (see French):

Top X Axis Minimum
Top X Axis Maximum
Top X Axis Scale
Top X Axis Label
Bottom X Axis Minimum
Bottom X Axis Maximum
Bottom X Axis Scale
Bottom X Axis Label
X Axis Minimum
X Axis Maximum
X Axis Scale
X Axis Label
Left Y Axis Minimum
Left Y Axis Maximum
Left Y Axis Scale
Left Y Axis Label
Right Y Axis Minimum
Right Y Axis Maximum
Right Y Axis Scale
Right Y Axis Label
Y Axis Minimum
Y Axis Maximum
Y Axis Scale
Y Axis Label

That's a lot of strings that you can't just define upon runtime using a comprehension. So I haven't found a much neater way to do this myself so far.

cmkohnen (approved):

PR #705, opened 2024-01-03
improve translate/multiply handling

cmkohnen:
Atop a general refactor this fixes a few issues:
Sensitivity of the buttons is now handled correctly.
Fixed condition regarding an empty string.
Buttons can no longer be sensitive if no data is selected

### DIFF
--- a/src/application.py
+++ b/src/application.py
@@ -222,14 +222,19 @@ class PythonApplication(Graphs.Application):
             stack_switcher.set_hexpand("true")
             window.get_stack_switcher_box().prepend(stack_switcher)
             window.set_title(self.props.name)
-            actions = ["multiply_x", "multiply_y", "translate_x",
-                       "translate_y"]
-            entries = [getattr(window, f"get_{action}_entry")()
-                       for action in actions]
-            buttons = [getattr(window, f"get_{action}_button")()
-                       for action in actions]
-            for entry, button in zip(entries, buttons):
-                entry.connect("notify::text", self.set_entry_css, button)
+            actions = ("multiply_x", "multiply_y", "translate_x",
+                       "translate_y")
+            for action in actions:
+                entry = window.get_property(action + "_entry")
+                button = window.get_property(action + "_button")
+                entry.connect(
+                    "notify::text", self.set_entry_css, entry, button,
+                )
+                data.connect(
+                    "notify::items-selected",
+                    self.set_entry_css, entry, button,
+                )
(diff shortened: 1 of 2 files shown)


Sjoerd1993:
>  Buttons can no longer be sensitive if no data is selected

Totally missed that this was a consequence of last change there, thanks for noticing. That the buttons became sensitive if the string was empty was on purpose, as otherwise the default state looks broken. That is mitigated in this PR by simply setting a default value of 10 on each field. Which I think is reasonable.

In principle we could go for `1` on multiply, and `0` on translate, so that the default value is chosen such that its inconsequential, but personally I'm fine with `10` as well.

Sjoerd1993 (approved):

src/application.py
```diff
@@ -244,15 +249,14 @@ def do_activate(self):
...
             window.present()
 
-    def set_entry_css(self, entry, _string, button):
+    def set_entry_css(self, _object, _param, entry, button):
         try:
             value = utilities.string_to_float(entry.get_text())
-            if value is None and entry.get_text() != "":
-                entry.add_css_class("error")
-                button.set_sensitive(False)
+            if value is not None and entry.get_text() != "":
```
Sjoerd1993, line 255:
```suggestion
            if value is not None:
```
If the entry is a blank string, the value will be zero. Not completely sure why I had I had that as condition, but probably has to do with the order I was testing exceptions. (First handled empty strings, then noticed that other strings could also deliver `None` without throwing an error so added `None` as condition.)

Sjoerd1993 (changes requested):
See above comment, it we merge this then I'll update the beta branch again.

src/application.py
cmkohnen: I'm not sure that would be the better solution. An empty string passing as 0 wouldn't be intuitive. So the idea here being still, that you cant use the button when the input is empty

Sjoerd1993: If `entry.get_text() != ""`, then value will always return `None`. Thus the second condition seems kind of redundant, as if the second condition holds true (`entry.get_text() != ""`), then the first one (`value is not None`) will be false. So it would never hit that case, and removing the second condition would functionally be the exact same thing.

Unless I'm missing something. My point was not to change anything with the functional logic itself, that is fine. It works exactly as it should work now imo, just meant there's a logical redundancy.

Sjoerd1993 (approved):

Issue #717, opened 2024-01-11
Legend position menu capitalization


bertob:
The menu items in the legend position menu are not title case:

![image](https://github.com/Sjoerd1993/Graphs/assets/1908896/2a8866eb-ebe7-45af-84d4-7bb5f75fe89e)

Sjoerd1993:
Woops, thanks for noticing. Have put a PR that fixes this. Can probably push this out in a quick patch so that it's fixed before the next TWIG issue. 

I always notice that this brings a measurable amount of installs, so would be nice to have that updated then (I also see in the Github analytics that most unique visitors come from either TWIG or Planet GNOME).

PR #718, opened 2024-01-11
Update Title Case

Sjoerd1993:
Fixes issue #717 

I was thinking of maybe pushing the 1.7.1 update already after merging this, but in principle I can wait until tomorrow so we could update the last localization for the languages that are at 99% now. 

It's okay to do it for the 1.7.2 patch, but if you have time maybe @sabriunal could update the last five strings or so in Turkish? Changed some of the strings close to release which didn't get updated. 

Same with German @cmkohnen.  Also, just a hand full of strings, would be nice to have those in the next update as well. My idea was to release the patch before the next TWIG release, as I notice in the statistics that that always brings a measurable wave in installs.

### DIFF
--- a/data/se.sjoerd.Graphs.appdata.xml.in.in
+++ b/data/se.sjoerd.Graphs.appdata.xml.in.in
@@ -51,6 +51,11 @@
     </screenshot>
   </screenshots>
   <releases>
+    <release version="1.7.1" date="2024-01-11">
+      <description translatable="no">
+        <p>Minor patch that sets the legend position strings to title case, and updates the locals to the latest version on Weblate.</p>
+      </description>
+    </release>
     <release version="1.7.0" date="2024-01-10">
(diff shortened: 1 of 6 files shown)


Sjoerd1993:
> I would wait a little before releasing. So that tge translations can indeed be updated.

Yeah, I think we should at least get the german and turkish updated first. I can wait for the spanish one for now, as that one has more work left and there's not someone specifically I can ask to fix that directly.

sabriunal:
Turkish translation updated.

Sjoerd1993:
> Turkish translation updated.

Thank you! It will probably mark a few strings as "Needs Editing" after this PR gets merged, as this PR changes the capitalization of said strings. But the translation itself is complete now. Thanks a lot!

cmkohnen (approved):

PR #734, opened 2024-01-17
Do not optimise limits for hidden items

Sjoerd1993:
Does not optimise limits for items that are hidden. I found it unintuitive if items are hidden, that these are still taken into account when doing optimise limits. Skipping those items make the following possible:

[Skärminspelning från 2024-01-17 08-17-17.webm](https://github.com/Sjoerd1993/Graphs/assets/68477016/bf0557ee-461e-4c41-b6cb-062bad58579e)

Which is thus useful if you have a bunch of datasets in the same project with different limits, and just want to toggle between them. When items are not hidden, but non-active, then they are still used when optimising the limits.

### DIFF
--- a/src/data.py
+++ b/src/data.py
@@ -537,7 +537,7 @@ class Data(GObject.Object, Graphs.DataInterface):
         for item_ in self:
             if not isinstance(item_, item.DataItem) or (
                     not item_.get_selected()
-                    and figure_settings.get_property("hide-unselected")):
+                    and figure_settings.get_hide_unselected()):
                 continue
             for index in \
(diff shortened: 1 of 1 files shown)


src/data.py
```diff
@@ -535,7 +535,9 @@ def optimize_limits(self) -> None:
             for direction in ("bottom", "left", "top", "right")
         ]
         for item_ in self:
-            if not isinstance(item_, item.DataItem):
+            if not isinstance(item_, item.DataItem) or (
+                    not item_.get_selected()
+                    and figure_settings.get_property("hide-unselected")):
```
cmkohnen, line 540: This could use the direct call of get_hide_unselected()

